PR #5, opened 2019-04-18
Fixed plotSamples clear. Was only clearing half the buffer.

ashley-b:
Minor fix. I noticed only half plotSamples was being cleared to to memset works interms of bytes, not shorts


drowe67:
@ashley-b Thanks :+1:

PR #6, opened 2019-04-20
Fixed dynamic array mismatch new[] & delete -> delete[]

ashley-b:
Dynamic arrays are not freed with matching delete e.g
https://github.com/drowe67/freedv-gui/blob/ad08dae7b2efa836d901ef2db2d5f2235f5fab71/src/dlg_filter.cpp#L133
https://github.com/drowe67/freedv-gui/blob/ad08dae7b2efa836d901ef2db2d5f2235f5fab71/src/dlg_filter.cpp#L214

Should be eg
```
  SomeType xxxx = new SomeType[];
  delete[] xxxx;
```
But I would suggest using std::vector to take care of memory allocation and cleanup

### DIFF
--- a/src/dlg_filter.cpp
+++ b/src/dlg_filter.cpp
@@ -211,8 +211,8 @@ FilterDlg::FilterDlg(wxWindow* parent, bool running, bool *newMicInFilter, bool 
 //-------------------------------------------------------------------------
 FilterDlg::~FilterDlg()
 {
-    delete m_MicInMagdB;
-    delete m_SpkOutMagdB;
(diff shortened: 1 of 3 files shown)


drowe67:
Thanks @ashley-b 

I'e just merged some Travis integration with master.  Could you pls test your PR against that?

ashley-b:
I have move the bugfix branch up to your current master and test it works for me. Hopeful you see the branch at its new location .

drowe67:
Thanks @ashley-b

PR #9, opened 2019-05-01
Fixups for cross-compiling for Windows from Fedora Linux

hobbes1069:


drowe67:
@hobbes1069 would separating libcodec2 into libcodec2 and libfreedv help the cmake loop problem you are having?

hobbes1069:
Nope, I've got that all squared away, you can use build directories or internal builds and it will work. The problem is picking up all the libraries that need to be bundled with the windows installer for it to work. It's missing the libjpeg-62.dll dependency for some reason. My theory is that I'm only checking for dependencies of top level dependencies (2nd level deps) of freedv.exe and I'm not picking it up so I'm working on a recursive strategy but I need a way to know when to stop looking.

hobbes1069:
We'll let the build try but I'm not confident it will build. Now that it's merged let's work on any fixups in brad-2020.

hobbes1069:
Codec2 needs to be the brad-2020 branch.

drowe67:
I'm close with Travis build, just that version.h thingy....

hobbes1069:
> I'm close with Travis build, just that version.h thingy....

I have the fix in the codec2 library export in codec2/brad-2020, not master.

build_ubuntu.sh
```diff
@@ -0,0 +1,41 @@
...
+cd src
+export LD_LIBRARY_PATH=$LPCNETDIR/build_linux/src
+./freedv_tx 2020 $LPCNETDIR/wav/wia.wav - | ./freedv_rx 2020 - /dev/null
+
+# Finally, build freedv-gui
+cd $FREEDVGUIDIR
+mkdir  -p build_linux && cd build_linux
+cmake -DCMAKE_BUILD_TYPE=Debug -DCODEC2_BUILD_DIR=$CODEC2DIR/build_linux -DLPCNET_BUILD_DIR=$LPCNETDIR/build_linux ..
+make
+
```
drowe67: @hobbes1069 Here is a version of the script @ac0zj sent us, that pulls the codec2 and lpcnet repos under freedv-gui.  I recall earlier in the week you mentioned something similar as part of your cmake work?  Apologies if there is some overlap here, and pls feel free to suggest the best way fwd. 

Aims here are to make building FreeDV easy, document it (in the README.md), and test it (with Travis).

drowe67:
@hobbes1069 are the "Building and installing on Windows" & "Bootstrapping wxWidgets build" sections of README.md useful any more?

drowe67:
@ac0zj when I press "Start" in 2020 mode I get a segfault in the Speex init code, looking at it with gdb:
```
1  0x00007ffff5ae11a2 in speex_preprocess_state_init () at /usr/lib/x86_64-linux-gnu/libspeexdsp.so.1
#2  0x00005555555e6711 in MainFrame::OnTogBtnOnOff(wxCommandEvent&) (this=0x55555598e590, event=...)
```
I have Filter-Speex Preprocess unchecked.  Will poke about a bit.

drowe67:
Ah, think it was a shared lib search path thingy.  Have freedv-gui 2020 deciding now from a wave file :-)  Nice work guys :+1:

drowe67:
@ac0zj at the bottom of User Guide text file is ```Full Duplex Testing with loopback```.  This worked well - you can basically run full duplex and talk to yourself.  By feeding in audio files from Tool-Mic In you can play audio through the system.  It's useful as you only need one PC.

In this configuration:
1/ I measured CPU load (full duplex) at 57% in my Lenovo X230 laptop
2/ Enabled some channel noise in Tools-Options and it performs as expected by simulations - bit errors start at 3dB, and increase rapidly beneath 2dB SNR.  Cool!

hobbes1069:
@drowe67 , It could probably be reformatted but the "Building and Installing on Windows" is just a disclaimer that it should be possible, but only cross-compiling (on Fedora for now, hopefully others later) is formally supported.

The bootstrapping could still be useful but I can't think of a modern Linux distro that doesn't already have wxWidgets of some form available. I did make sure it still works and updated it to the latest stable release.

drowe67:
@hobbes1069 Yes I agree on both points. I'll rm those sections, simpler is better.  If someone in future gets serious about native windows builds ... they can maintain/document it.

drowe67:
Oops I ididn't mean to close this, must have pressed the wrong button.

drowe67:
@hobbes1069 pls fele free to re-open if you need to do more work on the build system .... the re-open branch is greyed out for me ...

hobbes1069:
No worries! I was going to close it now that it's merged.

PR #28, opened 2019-08-21
cmake: fix codec2_include_dirs variable

ra1nb0w:
Just one question: lpcnetfreedv is strictly required or the code can be built without?
thanks


drowe67:
codec2 can be built without it, but you don't get 2020 mode.  Not sure if freedv-gui/master will still build without it,

ra1nb0w:
I know that. Thanks

Anyway, the PR is not related to the question. Only fix a variable.

hobbes1069:
LGTM.

PR #32, opened 2019-11-09
OSX Catalina build updates

tmiw:
Some updates to the build script to ensure that binaries built on OS X Catalina (10.15) can still be run on older versions:

1. Add building of dylibbundler to build_osx.sh script (newer version has updates to set rpath correctly) and update call to it in CMakeLists.txt.
2. Build DMG using older HFS+ file system (instead of APFS).


drowe67:
Thanks @tmiw :+1:

Is README.osx up to date?

tmiw:
It is now. :) Thanks for reminding me!

PR #34, opened 2019-11-18
Issues causing FreeDV 1.4 release candidate to crash

drowe67:
Some issues causing freedv-gui to crash, that were found by the UK team.  Anything that causes crashes should be fixed for the FreeDV 1.4 release:

1. assert() that is firing on Windows machines, was pretty benign, just testing an assumption about a FIFO having space that can tracked down later.
1. When USB sound devices are changes, and Start is pressed without re running Tools/Audio Config, freedv-gui crashes.  OK have reproduced this one on Linux, but no quick fix, so have posted as Issue https://github.com/drowe67/freedv-gui/issues/35

### DIFF
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -4094,8 +4094,6 @@ void txRxProcessing()
                 fprintf(stderr, "  nout: %d\n", nout);
             }
             ret = codec2_fifo_write(cbData->outfifo1, outsound_card, nout);
-            // should never fire as we check there is enough room before entering while loop
-            assert(ret != -1);
         }
     }
    
--- a/docker/README_docker.md
+++ b/docker/README_docker.md
@@ -19,17 +19,23 @@ Building is only required once
 ```
 docker-compose -f docker-compose-win.yml build
 ```
+
 ## Running the image(s)
 ```
 docker-compose -f docker-compose-win.yml up 
 ```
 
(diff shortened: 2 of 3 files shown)


src/fdmdv2_main.cpp
```diff
@@ -4094,8 +4094,6 @@ void txRxProcessing()
                 fprintf(stderr, "  nout: %d\n", nout);
             }
             ret = codec2_fifo_write(cbData->outfifo1, outsound_card, nout);
-            // should never fire as we check there is enough room before entering while loop
-            assert(ret != -1);
```
tmiw: I actually just noticed this on OSX as well (happens the instant I push PTT) and it appears it's related to sample rate mismatches. For instance, my Griffin iMic USB sound card seems to default to 48000Hz while my MacBook Pro's internal sound card defaults to 44100. Setting both to 44100 in the app causes the crash not to occur.

Anyway, I haven't dug too much into the Codec2 end of things but does removing the assertion result in other side effects?

drowe67: No it's just checking an assumption about how many samples are in a FIFO.  The current Windows Beta has been built off this branch, and have been working OK for a several weeks.

tmiw: Cool. I'm fine with this fix, then. I'll see about building a new OSX image with it as well.

PR #36, opened 2019-11-24
Prompt the user to grant permission to FreeDV for microphone access

tmiw:
Adds OSX-specific code and build logic to enable the Apple-mandated permissions prompt for microphone access. Otherwise, PortAudio can't open the microphone and thus FreeDV will be unable to function.


drowe67:
@tmiw is this one OK to merge too?

tmiw:
Go for it. I don't think there have been any issues reported related to these changes.

PR #46, opened 2020-02-08
Horus Binary Updates

darksidelemm:
Updates the call to horus_rx to match the changes made in https://github.com/drowe67/codec2/pull/96

Also updates the user manual to refer out to the Horus Binary setup guide.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -401,25 +401,13 @@ FreeDV 2020 Tips:
 
 ### Horus Binary Mode
 
-Horus Binary mode (HorusB) High Altitude Balloon (HAB) telemetry using
-the same FSK modem as 2400A/B and 800XA.
+The FreeDV GUI also supports the Horus Binary (HorusB) modulation, which is used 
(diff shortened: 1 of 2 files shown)


darksidelemm:
Note that I fully expected this CI build to fail, as it depends on changes to codec2. Not sure how you deal with that in CI!

drowe67:
> Note that I fully expected this CI build to fail, as it depends on changes to codec2. Not sure how you deal with that in CI!

You can change the branch ```build.sh``` uses https://github.com/drowe67/freedv-gui/blob/master/build_linux.sh#L11

However need to set that back to master before a merge.

drowe67:
@darksidelemm silly question but have you tested the changes with freedv-gui by running freedv-gui with Horus signals?  I don't think these changes can break anything else in freedv-gui (e.g. the FreeDV modes).

Once we get your codec 2 changes merges this PR should start building properly, and we can merge.

darksidelemm:
Yes, I've done a complete build of freedv_gui against my modified codec2_branch and run it against off-air signals - no problems found!

darksidelemm:
There we go, CI passes!

PR #47, opened 2020-02-11
Added sideband detection to FreeDV's Hamlib support.

tmiw:
Per agreed convention, FreeDV uses LSB for bands under 10MHz and USB for bands above it. This differs from the usual "data mode" convention of USB regardless of band, This commit adds the ability for FreeDV to detect the sideband currently being used by the radio and emit a warning dialog if it differs from the expected convention.

Limitations:

1. Hamlib support must be enabled in FreeDV's settings.
2. Detection is only done when the user pushes Start. If the sideband changes during the session, FreeDV will not detect the change.
3. FreeDV/Codec2 will not attempt to mitigate incorrect sideband settings during the decode process.

## Test Plan
David, 21 Feb 2020
| Test | Description | OS | Radio | Who | Status |
| --- | --- |  --- | --- | --- | --- |
| 1 | Detects USB/LSB From Hamlib enabled radio | OSX | KX3 | Mooneer | **Pass** |
| 2 | Detects USB/LSB From Hamlib enabled radio  | Ubuntu 18  | IC-7200 | David | **Pass** |
| 3 | Detects USB/LSB From Hamlib enabled radio  | Windows 32 or 64  | Flex/IC-7000  | K5WH VK5APR  | **Pass**  |
| 4 | Hamlib PTT still works  | Any  | KX3 | Mooneer  | **Pass** |
| 5 | Serial PTT still works  | Any  | IC-7000  | Peter, VK5APR  | **Pass**  |

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -178,7 +178,9 @@ unplugged/plugged.
 
 ### USB or LSB?
 
-On bands beneath 10 MHz, LSB is used for FreeDV.  On 10MHz and above, USB is used. After much debate, the FreeDV community has adopted the same conventions as SSB, based on the reasoning that FreeDV is a voice mode.  
+On bands beneath 10 MHz, LSB is used for FreeDV.  On 10MHz and above, USB is used. After much debate, the FreeDV community has adopted the same conventions as SSB, based on the reasoning that FreeDV is a voice mode. 
+
+As an aid to the above, FreeDV will show the current mode on the bottom of the window upon pressing the Start button if Hamlib is enabled and your radio supports retrieving frequency and mode information over CAT. If your radio is using an unexpected mode (e.g. LSB on 20 meters), it will display that mode on the bottom of the window next to the Clear button in red letters. When a session is not active, Hamlib isn't enabled or if your radio doesn't support retrieving frequency and mode over CAT, it will remain grayed out with "unk" displaying instead of the mode (for "unknown").
 
 ## Common Problems
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,6 +98,7 @@ if(APPLE)
     add_custom_command(
         TARGET freedv
         POST_BUILD
+        COMMAND rm -rf FreeDV.* dist_tmp
         COMMAND mkdir ARGS -p FreeDV.app/Contents/MacOS
         COMMAND mkdir ARGS -p FreeDV.app/Contents/Resources/English.lproj
         COMMAND cp ARGS ${CMAKE_CURRENT_SOURCE_DIR}/info.plist FreeDV.app/Contents
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,7 +98,7 @@ if(APPLE)
     add_custom_command(
         TARGET freedv
         POST_BUILD
-        COMMAND rm -rf FreeDV.* dist_tmp
+        COMMAND rm ARGS -rf FreeDV.* dist_tmp
         COMMAND mkdir ARGS -p FreeDV.app/Contents/MacOS
         COMMAND mkdir ARGS -p FreeDV.app/Contents/Resources/English.lproj
         COMMAND cp ARGS ${CMAKE_CURRENT_SOURCE_DIR}/info.plist FreeDV.app/Contents
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -2825,14 +2825,6 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
         //printf("m_textEncoding = %d\n", wxGetApp().m_textEncoding);
         //printf("g_stats.snr: %f\n", g_stats.snr_est);
 
-        // attempt to start PTT ......
-        
-        if (wxGetApp().m_boolHamlibUseForPTT)
-            OpenHamlibRig();
-        if (wxGetApp().m_boolUseSerialPTT) {
-            OpenSerialPort();
-        }
-
         // attempt to start sound cards and tx/rx processing
         if (VerifyMicrophonePermissions())
         {
@@ -2841,6 +2833,14 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
         else
         {
             wxMessageBox(wxString("Microphone permissions must be granted to FreeDV for it to function properly."), wxT("Error"), wxOK | wxICON_ERROR, this);
+        }
+
+        // attempt to start PTT ......
+        
+        if (wxGetApp().m_boolHamlibUseForPTT)
+            OpenHamlibRig();
+        if (wxGetApp().m_boolUseSerialPTT) {
+            OpenSerialPort();
         }
 
         if (m_RxRunning)
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -734,6 +734,7 @@ MainFrame::~MainFrame()
 
     if (wxGetApp().m_hamlib)
     {
+        wxGetApp().m_hamlib->disable_sideband_detection();
         wxGetApp().m_hamlib->close();
         delete wxGetApp().m_hamlib;
     }
@@ -2614,14 +2615,7 @@ bool MainFrame::OpenHamlibRig() {
         wxMessageBox("Couldn't connect to Radio with hamlib", wxT("Error"), wxOK | wxICON_ERROR, this);
     else
     {
-        // Validate current sideband configuration of radio. FreeDV by convention uses the same sidebands
-        // as SSB voice.
-        wxString error;
-        status = wxGetApp().m_hamlib->is_correct_sideband(error);
-        if (!status)
-        {
-            wxMessageBox(error, wxT("Warning"), wxOK | wxICON_WARNING, this);
-        }
+        wxGetApp().m_hamlib->enable_sideband_detection(m_BtnSSBStatus);
     }
 
     return status;
@@ -2885,6 +2879,7 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
                 if (hamlib->ptt(false, hamlibError) == false) {
                     wxMessageBox(wxString("Hamlib PTT Error: ") + hamlibError, wxT("Error"), wxOK | wxICON_ERROR, this);
                 }
+                hamlib->disable_sideband_detection();
                 hamlib->close();
             }
         }
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -282,13 +282,6 @@ bool MainApp::OnInit()
 //-------------------------------------------------------------------------
 int MainApp::OnExit()
 {
-    // Note: sideband detection needs to be disabled here instead
-    // of in the destructor due to its need to touch the UI.
-    if (wxGetApp().m_hamlib)
-    {
-        wxGetApp().m_hamlib->disable_sideband_detection();
-    }
-
     //fprintf(stderr, "MainApp::OnExit\n");
     if (m_plugIn) {
         #ifdef __WXMSW__
@@ -2432,6 +2425,13 @@ void MainFrame::OnRecFileFromModulator(wxCommandEvent& event)
 //-------------------------------------------------------------------------
 void MainFrame::OnExit(wxCommandEvent& event)
 {
+    // Note: sideband detection needs to be disabled here instead
+    // of in the destructor due to its need to touch the UI.
+    if (wxGetApp().m_hamlib)
+    {
+        wxGetApp().m_hamlib->disable_sideband_detection();
+    }
+
     //fprintf(stderr, "MainFrame::OnExit\n");
     wxUnusedVar(event);
 #ifdef _USE_TIMER
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -2612,7 +2612,18 @@ bool MainFrame::OpenHamlibRig() {
     bool status = wxGetApp().m_hamlib->connect(rig, port.mb_str(wxConvUTF8), serial_rate);
     if (status == false)
         wxMessageBox("Couldn't connect to Radio with hamlib", wxT("Error"), wxOK | wxICON_ERROR, this);
- 
+    else
+    {
+        // Validate current sideband configuration of radio. FreeDV by convention uses the same sidebands
+        // as SSB voice.
+        wxString error;
+        status = wxGetApp().m_hamlib->is_correct_sideband(error);
+        if (!status)
+        {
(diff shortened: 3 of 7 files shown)


drowe67:
@tmiw thanks for you PR, it is a neat idea.

However I'm in two minds about this one, as I'm not sure we want the software to mandate the sideband convention.  I might pose that question to the list. So if someone chooses to use USB beneath 10MHz, will this pop up a dialog every time Start is pressed?

I'll test it today on my IC7200.

tmiw:
@drowe67, correct. Likewise, if your radio is in LSB above 10MHz, the same warning popup appears.

On that note, I wonder if the check should be more strict (that is, require PKTLSB/PKTUSB as opposed to just LSB/USB). That may affect radios without separate data modes, however.

drowe67:
Thanks @tmiw - lets wait a few days and see what the balance of opinions is on the list. The USB/LSB topic seems to generate a lot of traffic :-)

Another alternative would be something less intrusive than a pop up, like a LSB/USB indicator on the FreeDV main page, maybe going red if you're reversed from the convention.

drowe67:
@tmiw so a mixed response on the mailing list to this feature.  On balance I feel a pop up Dialog every time Start is pressed is a little strong, as some people may wish to use the opposite sideband.  However I'd be happy to have a feature that was a little more subtle, e.g. some sort of indication along the left hand side in Stats sub-window.

tmiw:
@drowe67, I was thinking next to the callsign text bar at the bottom, like so:

FreeDV USB indicator

I'm not sure yet if I want it as a button, however (to permit changes via Hamlib while a session is running?) or if it should just be a simple static field. Thoughts?

drowe67:
@tmiw Oh that does look quite nice!  Not sure if it should be on the LHS, closer to the Stats information?  I'd suggest just a static field rather than a button.  End users know how to swap sidebands.

tmiw:
@drowe67, here's what it looks like as a regular label instead of a button:

label

I'm hesitant to include it on the left hand side of the window since the controls on there already seem to be going all the way to the bottom. Unless you mean left hand side of that bottom bar (i.e. next to the Clear button)?

drowe67:
@tmiw yes I was thinking of either on the LHS next to clear button or in the Stats box.  BTW that level control isn't used much any more (and has never been that useful) so I was thinking or removing it.

tmiw:
@drowe67, I have the control wired up (and will follow radio changes, too, but I'm not sure about this yet as it seems to cause a slight issue on OSX the first time one uses FreeDV).  As the code is now, it'll appear as such when the sideband is correct:

correct

and like so when incorrect:

incorrect

So far it seems to work well with my KX3, but more testing would be good too. Moving its position should be straightforward once we're sure the backend logic works.

drowe67:
@tmiw OK well when you think it's ready for merging I can try it on Linux with my IC-7200.  We also need to test a Windows version.  There is some Docker support for Windows building on Ubuntu in freedv-gui/docker.

tmiw:
@drowe67, after more testing and bug fixes, I think it's good now. I settled on having it immediately to the left of the Clear button until the Level area gets removed.

drowe67:
@tmiw I've added a test plan at the tip of this page.  PTT support has been very problematic through the years so I'm inclined to test any changes in this area carefully. 

BTW well done for having a go at the GUI code, I find it really hard to work with myself, but something we are stuck with.  I think it originally came out of some code generator tool.

I've found a layout problem on Test 2, just seeing the last part of the LSB/USB string.  The colour changes OK with mode changes and PTT still works: nice :+1: 
Screenshot from 2020-02-21 07-10-12
Is it possible for you to build and test on Ubuntu? For these sorts of bugs you really need a fast compile/test loop.  I don't have the bandwidth to debug this one myself right now.

tmiw:
@drowe67, I have an Ubuntu VM on my MacBook Pro that I can try to use to debug this tonight. Offhand, though, something that you can try is to bump up the wxSize for that control (see [here](https://github.com/drowe67/freedv-gui/pull/47/commits/f09972844ca0e9084c95fe6041a507bedec62c6f#diff-38a5d30b16967e1d8943d1ca5ae1a5ebR241)) to something greater than 60; if that works, it's possible that we need to have specific #ifdefs for OSX vs. other platforms or potentially find another way to lay out that control.

drowe67:
@tmiw IIRC so far we've managed to avoid any OS-specific GUI layout code, that's one nice thing about WxWidgets, so I figure it's just some tweaks as you suggest.  BTW if you have a Linux machine you can use the Docker support in ```freedv-gui/docker``` to build Windows 32 and 64 bit installers.

tmiw:
@drowe67, I was able to figure out the UI issue on Ubuntu. In addition, I found a segfault issue when exiting without pushing Start first, so I fixed that as well. :)

drowe67:
@tmiw Travis (and I repeated it here) have a build issue with the latest commit:
```
/home/david/tmp/freedv-gui/src/hamlib.cpp:38:17: error: invalid conversion from ‘int’ to ‘rmode_t’ [-fpermissive]
     m_currMode(0)  {
                 ^
src/CMakeFiles/freedv.dir/build.make:350: recipe for target 'src/CMakeFiles/freedv.dir/hamlib.cpp.o' failed
```

tmiw:
@drowe67, oops. It should be fixed now though.

drowe67:
@tmiw I've done a little testing:

1. Layout looks good now thanks.
1. When freedv exits, there's a seg fault at the end:
   ```
   <snip>
   Hamlib::ptt: rig_set_ptt returned: 0
   rig:rig_close called 
   rig:rig_cleanup called   
   Segmentation fault (core dumped)
   ```
   This seg fault isn't present for me on freedv-gui/master.
1. On one test the serial port used for hamlib stopped working, and I had to reboot my laptop.  My rig was left in tx state. I'm not sure if that was just a random unlucky problem.  We need to make sure anything around hamlib is very stable, as it's controlling peoples PTT.  I haven't been able to repeat this issue.
1. Does the design poll hamlib constantly?  This might be adding significant load and increasing our changes of serial port problems in a high RF environment.  It would be OK (and perhaps preferrable) just to sample the mode once when PTT is pressed.

tmiw:
@drowe67, the segfault should have been fixed after 6c1faa4 if it's the same one I'm thinking of. What steps led up to it?

As for polling, hamlib has two modes: RIG_TRN_POLL and RIG_TRN_RIG. In my testing, the latter didn't update FreeDV when changing modes/bands on my KX3 (but the former did). I was concerned about using RIG_TRN_POLL too but FWIW there didn't seem to be much (if any) difference in CPU usage on my MacBook Pro between the two. That said, it might be worthwhile to see exactly what the difference is between those two modes.

drowe67:
> @drowe67, the segfault should have been fixed after [6c1faa4](https://github.com/drowe67/freedv-gui/commit/6c1faa40c1a4f4252cd8773fa6b2ee1d798a74c9) if it's the same one I'm thinking of. What steps led up to it?
With hamlib PTT enabled:

```
./build_linux/src/freedv
Click on Start (freedv starts receiving)
Click on Stop (freedv stops)
Exit freedv back to command line
```

tmiw:
@drowe67, figured out the issue. For some reason it wasn't happening on OSX but was happening on Linux.

Also, I can confirm PTT works on both Ubuntu and OSX, so I updated the table above.

tmiw:
BTW, I investigated a bit further on Hamlib's polling and it seems to default to 2x/second (every 500ms). Theoretically this interval can be extended further but I'm not sure what value would be appropriate, if any. I'm leaning towards extending that to every 1000ms, especially since there would be some value in knowing the current sideband status even on receive. Then again, the new code already initializes itself with the current state on start.

drowe67:
> @drowe67, figured out the issue. For some reason it wasn't happening on OSX but was happening on Linux.

Thanks :+1: Yes cross platform testing tends to throw up the corner cases.  Great work on testing on both OSes.  Can confirm the segfault issues is gone for me (Ubuntu 18/IC-7200), and the sideband-detect feature is working for me, tested on 40m/20m on USB/LSB.

drowe67:
> BTW, I investigated a bit further on Hamlib's polling and it seems to default to 2x/second (every 500ms). Theoretically this interval can be extended further but I'm not sure what value would be appropriate, if any. I'm leaning towards extending that to every 1000ms, especially since there would be some value in knowing the current sideband status even on receive. Then again, the new code already initializes itself with the current state on start.

Yes I'm still wondering if polling all the time is necessary/desirable, although it is kind of neat watching freedv-gui track the state of the radio.  However the current design is working, so lets throw it put for further testing on Windows and see how it goes.

Do you feel comfortable building the Windows installers with ```freedv-gui/docker``` or do you want me to build them?

tmiw:
@drowe67, I have to get Docker installed on my laptop and figure out how it works first. Whoever manages to do the build first I suppose?

tmiw:
@drowe67, any idea what's happening here with the Docker builds? Host VM is Ubuntu 18.04 LTS if that helps.

`Step 7/10 : RUN dnf install -y mingw{32,64}-speex mingw{32,64}-wxWidgets3 mingw{32,64}-hamlib mingw{32,64}-portaudio mingw{32,64}-libsndfile mingw{32,64}-libsamplerate.noarch svn
 ---> Running in e2970a660f4b
Warning: failed loading '/etc/yum.repos.d/_copr:copr.fedorainfracloud.org:hobbes1069:mingw.repo', skipping.
Last metadata expiration check: 0:04:50 ago on Sun Feb 23 08:37:21 2020.
No match for argument: mingw32-hamlib
No match for argument: mingw64-hamlib
No match for argument: mingw32-portaudio
No match for argument: mingw64-portaudio
No match for argument: mingw32-libsndfile
No match for argument: mingw64-libsndfile
No match for argument: mingw32-libsamplerate.noarch
No match for argument: mingw64-libsamplerate.noarch
Package subversion-1.12.2-3.fc30.x86_64 is already installed.
Error: Unable to find a match: mingw32-hamlib mingw64-hamlib mingw32-portaudio mingw64-portaudio mingw32-libsndfile mingw64-libsndfile mingw32-libsamplerate.noarch mingw64-libsamplerate.noarch
ERROR: Service 'fdv_win_fed_30_s' failed to build: The command '/bin/sh -c dnf install -y mingw{32,64}-speex mingw{32,64}-wxWidgets3 mingw{32,64}-hamlib mingw{32,64}-portaudio mingw{32,64}-libsndfile mingw{32,64}-libsamplerate.noarch svn' returned a non-zero code: 1`

db4ple:
Hi Moneer,
the the failing packages are from a repository @hobbes1069 maintains. Obviously this is not available from within your docker. Reason may be firewall or other issues such as your docker file references too old fedora releases (the repository holdes packages for fedora 30, 31 and 32 only)

In the docker-compose.yml and/or the dockerfile the fedora release to be used is defined.

drowe67:
Hmm, maybe the version of Fedora that Docker is using needs to be updated.  Looks like the freedv/docker scripts are currently configured for Fedora 30.

It probably runs OK for me as I have already installed the early packages in my Docker installation.

@hobbes1069 what version of Fedora should we be using from your repo?

tmiw:
@drowe67, @hobbes1069: I changed 30 to 31 in the Dockerfile so we'll see how it goes. It seems to have gotten farther in the process so far, anyway.

tmiw:
@drowe67, so I managed to build an installer with the Docker image: https://www.dropbox.com/s/h6mqqjwl3mnr35v/FreeDV-1.4.0-devel-20200224-f8d1eea-win64.exe?dl=0

Unfortunately, while PTT does seem to work with the above build, it's also not tracking changes to the radio as they're being made. However, it does seem to pick up the current state when you push Start. I'll investigate to see if there are any caveats to polling in Hamlib.

(Tested with KX3 + Microsoft Surface 3.)

tmiw:
Hmm, it looks like Hamlib uses sigaction() to implement rig polling: https://github.com/Hamlib/Hamlib/blob/master/src/event.c#L225. Based on what I was able to find, [Windows does not have sigaction()](https://stackoverflow.com/questions/32389905/sigaction-and-porting-linux-code-to-windows) so polling would need to be implemented via another mechanism for Windows. 

I see a few options here:

- Windows specific (or possibly for all platforms) polling mechanism FreeDV itself implements. (I can definitely see this conflicting with PTT unless implemented carefully. It'd probably be best not to poll at all while PTT is on if we went this route.)
- Disable polling entirely and only update state when Start's pushed.
- File a feature request with the Hamlib project to get polling working on Windows. (Note: assuming this got into their tree, the Docker container would likely need to be modified to use the nightly/beta builds of Hamlib until Fedora can catch up. And in the meantime, this would be a feature that only works on Linux/OSX.)

Thoughts?

drowe67:
@tmiw nice work in getting the Windows installer built, running some tests, and tracking down the issue :+1: 

I like the 2nd option, disable polling and just sample when Start is pressed.

tmiw:
@drowe67, I disabled the polling code via #if/#endif in the event that we ever want to re-enable it. I think it's also a good idea to submit a feature request to Hamlib, so I went ahead and submitted [this](https://github.com/Hamlib/Hamlib/issues/203) on the Hamlib project.

drowe67:
@tmiw nice work :+1:   When you're happy with you testing, suggest we get the Windows version out for a few other Hams to try with different radios.

tmiw:
@drowe67, a comment in the Hamlib issue I created seems to indicate that it's possible for a user to be using a radio that doesn't support reading the current frequency and mode. How should we handle that? I'm thinking not even bothering to enable the control if we can't perform the initial read but perhaps we could also have "Unknown" display in red instead?

drowe67:
@tmiw if the Hamlib features needed aren't supported on a particular radio (or Hamlib is not enabled) suggest the field be left blank.  That keeps us consistent with current versions of FreeDV.  Red is a bit strong, as it will signal a problem to the end users and make them ask us "what's wrong?".

tmiw:
@drowe67, it now uses "unk" (for "unknown") when inactive or if the current radio state can't be retrieved. The new Windows installer for this is [here](https://www.dropbox.com/s/1cbazo5hnf8u7zy/FreeDV-1.4.0-devel-20200225-f8d1eea-win64.exe?dl=0); I'm comfortable with letting others try it too.

drowe67:
@tmiw nice work :+1: Pls feel free to post to the digital voice group (perhaps 32 and 64 bit version).  It would be nice to get some coverage of non Hamlib radios too (test 5 above), and perhaps add a small section to the user manual explaining the new feature (and situations where it won't work).

tmiw:
@drowe67, email sent. 👍

tmiw:
@drowe67, added a paragraph about the mode display to the documentation. Let me know if there's anything I should change with that.

drowe67:
@tmiw just read it and it's great :+1: Also noting the fine feedback from end users on the mailing list - well done!

PR #49, opened 2020-03-06
Add minimum size for SNR text field to avoid it being chopped off in OSX and possibly other platforms.

tmiw:
@drowe67, found some weirdness with the SNR text field on OSX while testing 1.4.1. As the title states, I added a minimum size to that field to avoid the text being chopped off if the SNR is two digits or greater.

(Note: I think this has been around for a bit. I don't think this is related to the 1.4.1 changes.)


drowe67:
@tmiw OKcool  I just tried this branch on Linux and looks OK to me with a high SNR:
Screenshot from 2020-03-07 07-20-06

PR #55, opened 2020-05-07
Fix AVX detection for x86_64 platforms

### LINKED ISSUE #54: AVX detection via cpuid causes segfault on x86_64
This inline asm, while ok in 32-bit, corrupts the state of the RDI register in 64-bit mode:
https://github.com/drowe67/freedv-gui/blob/b097ce782ddcbbec6662dc0bb685d6d072b5b865/src/fdmdv2_main.cpp#L4498-L4503
Depending on compiler optimizations (e.g. inlining), this can cause a segfault when checking `isAvxPresent`:
https://github.com/drowe67/freedv-gui/blob/b097ce782ddcbbec6662dc0bb685d6d072b5b865/src/fdmdv2_main.cpp#L323

Rewriting to use `__cpuid()` as provided by `cpuid.h` solves the issue.

bwarden:
Use gcc's __cpuid() function to check for AVX support in a safer way.
Fixes #54

### DIFF
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -4502,10 +4502,10 @@ void MainFrame::checkAvxSupport(void)
     eax = ebx = ecx = edx = 0;
     __cpuid(1, eax, ebx, ecx, edx);
 
-    if (ecx & (1<<26) && ecx & (1<<27)) {
+    if (ecx & (1<<27) && ecx & (1<<28)) {
         // CPU supports XSAVE and AVX
-        uint64_t xcr0;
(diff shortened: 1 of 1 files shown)


drowe67:
Nice work @bwarden.  

Did you discover this problem on Windows or Linux? 

I'll try it out in the next day or two on a machine here

bwarden:
Linux. Our context is packaging codec2 for distribution in [Clear Linux](https://clearlinux.org/):
https://github.com/clearlinux-pkgs/codec2

drowe67:
Looks good :+1: Tested on an AVX and non-AVX machine

PR #59, opened 2020-05-30
Revert disabling of horus features, use horus_open_advanced

darksidelemm:
I've also made the modem stats (e.g. SNR display) update on every frame of samples, not just on every received packet, for a more fluid SNR display.

### DIFF
--- a/build_linux.sh
+++ b/build_linux.sh
@@ -9,7 +9,7 @@ export CODEC2DIR=$FREEDVGUIDIR/codec2
 export LPCNETDIR=$FREEDVGUIDIR/LPCNet
 
 # change this when working on combined codec2/freedv-gui changes
-CODEC2_BRANCH=master
+CODEC2_BRANCH=horus_api_fixes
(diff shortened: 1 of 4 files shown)


darksidelemm:
The CODEC2_BRANCH line in build_linux.sh will need to be reverted before merging. It's changed to match the branch in codec2 with the required changes, so that the tests pass.

drowe67:
@darksidelemm looks good.  

If you haven't done so please test the non-Horus voice modes too as a sanity check, e.g. Rx only config, and Tools-Start/Stop Play File from Radio, use the sample files in the wav directory for each mode.

If we merge the codec2 side first you can then change the branch in this PR to test before merging.

darksidelemm:
I can confirm that all the test files can be received using this branch. I was also able to transmit using all modes (except HorusB of course).

drowe67:
Horus support is being moved out of freedv-gui and into it's own fork.

PR #61, opened 2020-06-03
OpenSuSE-15.1 and make install instructions

drowe67:


tmiw:
@drowe67, I can confirm that this change still compiles on OSX Catalina.

we7u:
Looks good. I see you already merged it into master. Thanks!

PR #64, opened 2020-07-18
freedv-gui 1.4.2 release

drowe67:
Bumped version number to 1.4.2 and added some release notes.

Built against [codec2](https://github.com/drowe67/codec2/commit/8bfb6abdd727da6bc45852772cb45d3efc30a8e4) and [LPCNet](https://github.com/drowe67/LPCNet/commit/0a361f1e526d8b11d1904c3dab1cd2baaa84b4eb)

@hobbes1069 - codec2 has been refactored over recent months, so I wanted to get a maintenance release of freedv-gui out there to keep it up to date. If all is well nothing much should change from the point of view of end users.

PR #65, opened 2020-07-21
Tweak README and User Manual #64

jimt:
* add library name for building on Ubuntu 20.04
* fix link to example send mail script
* fix typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This document describes how to build the FreeDV GUI program for various operatin
   * [FreeDV GUI User Manual](USER_MANUAL.md)
   * [Building for Windows using Docker](docker/README_docker.md)
   
-## Building on Ubuntu Linux (16-19)
+## Building on Ubuntu Linux (16-20)
   ```
   $ sudo apt install libc6-i386 libspeexdsp-dev libsamplerate0-dev sox git \
   libwxgtk3.0-dev portaudio19-dev libhamlib-dev libasound2-dev libao-dev \
@@ -17,12 +17,14 @@ This document describes how to build the FreeDV GUI program for various operatin
   $ cd freedv-gui
   $ ./build_linux.sh
   ```
+  (For Ubuntu 20.04 the wxWidgets package is named `libwxgtk3.0-gtk3-dev`.)
+  
   Then run with:
   ```
   $ ./build_linux/src/freedv
   ```
   
-  Note this build all libraries locally, nothing is installed on your machine.  ```make install``` is not required.
+  Note this builds all libraries locally, nothing is installed on your machine.  ```make install``` is not required.
 
 ## Building on Fedora Linux
   ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -17,12 +17,14 @@ This document describes how to build the FreeDV GUI program for various operatin
   $ cd freedv-gui
   $ ./build_linux.sh
   ```
+  (For Ubuntu 20.04 the wxWidgets package is named `libwxgtk3.0-gtk3-dev`.)
+
```
drowe67, line 21: @jimt - have you actually built and run freedv-gui on Ubuntu 20?  I haven't tried this myself.

srsampson: As a side comment, I am now running well on Ubuntu 20. Mostly had to add in a bunch of Python2 and 3 stuff.

jimt: Since moving to Ubuntu 20.04 I've only used FreeDV-GUI in test mode.

I haven't encountered any problems with Python, but I don't use `send_email_on_sync.py` (which appears to be Python 3 aware).

drowe67: Thanks @srsampson and @jimt.

@jimt - what do you mean by "test mode'?  I'm sure I should know what that is ....

jimt: [Playing the supplied sample files](https://github.com/drowe67/freedv-gui/blob/master/USER_MANUAL.md#test-wave-files)  (in receive-only mode) to check the decode path without bringing the radio into the equation.

And it was then that I noticed there was a mode missing from this repo.  #66  :sunglasses:

drowe67: Yep that's the basic sanity check I do for freedv-gui too :+1: 

One day I'd like to add command line options for automated tests using freedv-gui, e.g. to decode a file without needed any mouse clicks.

drowe67: I've also had requests for PDF version of the manual.  There are tools for markdown->PDF however we'd need a way to automate the conversion every time we change the manual.  That's the nice thing about Markdown - always up to date!

hobbes1069: Needs some tweaking but how does this look?
https://hobbes1069.fedorapeople.org/USER_MANUAL.pdf

I generated it with pandoc, shouldn't be too hard to integrate it with the cmake config and generate it during building.

hobbes1069: For posterity here's the command line I used:
`$ pandoc USER_MANUAL.md -o USER_MANUAL.pdf "-fmarkdown-implicit_figures -o" --from=markdown -V geometry:margin=.4in --toc --highlight-style=espresso`

drowe67: Be worth putting that in the document.  It would be nice if we could have Travis rebuild a PDF version every time we push

hobbes1069: Although the file isn't "large", I would hate too keep checking in a 400-500k file into git every time the markdown is changed. Perhaps a simple bash script for convenience of the end user would be sufficient?

drowe67: Yep that's where I'm stuck too - don't want to manually maintain this step.  Many end users are Windows users who can't drive a Linux command line.  Some automation would be nice, e.g. rebuild PDF every time markdown changes (maybe bumping version number and release date or at least including hit hash) and pop it on the freedv.org web site.

drowe67: Oh well as a first step I added your command line to the manual in https://github.com/drowe67/freedv-gui/commit/82e0078059234d6e8cf462c21d9d0a5750e7dc02  @hobbes1069 

I figure someone in a local group of Windows FreeDV users will have the Linux skills could do the conversion for their friends.

Issue #67, opened 2020-08-04
wxWidgets 3.1.4


drowe67:
https://github.com/drowe67/freedv-gui/pull/64#issuecomment-667903077

tmiw:
Subscribing to this issue as we may need 3.1.4 to fix some macOS UI issues I noticed (namely the UI not properly updating for dark mode). I'll investigate a bit when I get a chance.

drowe67:
@tmiw is this still an open issue on macOS?

tmiw:
@drowe67, nope, I've been building with 3.1.4 for a while now and haven't seen any assertion failures recently. 👍

drowe67:
That's great @tmiw good work :+1:

Issue #27, opened 2019-08-13
macOS 10.14 darkmode bad view


ra1nb0w:
The following is an example of bad behaviour of the bars on macOS 10.14 with dark mode enabled.
<img width="1731" alt="Screenshot 2019-08-12 at 20 03 15" src="https://user-images.githubusercontent.com/1543838/62918681-b627f600-bda0-11e9-8377-dd8699741a31.png">
Seems a sort of transparency because in the background I see the window behind freedv-gui.

PR #68, opened 2020-08-05
Port FreeDV to wxWidgets 3.1

tmiw:
Updates FreeDV UI code to support wxWidgets 3.1. Resolves issues #67 and #27. 

@drowe67, this branch seems to look fine on Mac but it wouldn't hurt to test on Windows and Linux too. 

## Test Plan

Try all Dialogs and basic functionality (say decoding a test file from wav dir) on:

| Test | OS | wxWidgets | Who | Pass |
| --- | --- | --- | --- | --- |
| 1 | OSX | 3.0 | Mooneer | Pass |
| 1 | OSX | 3.1 | Mooneer | Pass |
| 2 | Linux | 3.0 | David | Pass |
| 3 | Linux | 3.1 | David | Pass |
| 4 | WIndows | 3.0 | not tested | - |
| 5 | Windows | 3.1 | not tested  | - |

Windows could be 32 or 64 bit.


src/dlg_filter.h
```diff
@@ -48,17 +48,17 @@ typedef struct {
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
 // Class AuiNotebookNoKbd
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
-class AuiNotebookNoKbd : public wxAuiNotebook
+class AuiNotebookNoKbd : public wxNotebook
```
tmiw, line 51: Had to move from wxAuiNotebook to wxNotebook due to a stack overflow bug when creating instances of this object. Maybe wx 3.1.5 or later will resolve?

drowe67:
@tmiw wow that's a big effort - thanks :smile: I haven't had to play with all the wx_align commands for a while - last time I did I found it very confusing.  Many of these screens were generated by some sort of layout tool in 2012, I have tried to maintain them manually since.

Yes we will need to test on Windows and Linux, and that the changes still work OK with wxWidgets 3.0.  I guess we know (via Travis), that they build OK on 3.0 :+1: 

I can see in CMakeList.txt there is an option to download a wxWidgets tarball.  Can you pls post the cmake command line you used to build against wxWidgets 3.1?

drowe67:
@ra1nb0w can you pls try this PR?

tmiw:
@drowe67, the version of wxWidgets that the build is using is hardcoded into cmake/BuildWxWidgets.cmake. I used the same commands that are in build_osx.sh, namely:

```
# Finally, build freedv-gui
cd $FREEDVGUIDIR && git pull
mkdir  -p build_osx && cd build_osx && rm -Rf *
cmake -DCMAKE_BUILD_TYPE=Debug  -DBOOTSTRAP_WXWIDGETS=1 -DUSE_STATIC_DEPS=1 -DUSE_STATIC_SPEEXDSP=1 -DHAMLIB_INCLUDE_DIR=${HAMLIBDIR}/include -DHAMLIB_LIBRARY=${HAMLIBDIR}/lib/libhamlib.a -DCODEC2_BUILD_DIR=$CODEC2DIR/build_osx -DLPCNET_BUILD_DIR=$LPCNETDIR/build_osx ..
make VERBOSE=1

# Rebuild now that wxWidgets is bootstrapped
cmake -DCMAKE_BUILD_TYPE=Debug  -DBOOTSTRAP_WXWIDGETS=1 -DUSE_STATIC_DEPS=1 -DUSE_STATIC_SPEEXDSP=1 -DHAMLIB_INCLUDE_DIR=${HAMLIBDIR}/include -DHAMLIB_LIBRARY=${HAMLIBDIR}/lib/libhamlib.a -DCODEC2_BUILD_DIR=$CODEC2DIR/build_osx -DLPCNET_BUILD_DIR=$LPCNETDIR/build_osx ..
make VERBOSE=1
```

The above causes CMake to download and build wxWidgets itself.

ra1nb0w:
Fix issue https://github.com/drowe67/freedv-gui/issues/67 but not https://github.com/drowe67/freedv-gui/issues/27. See the image:
untitled
it is interesting that when you use the screenshot tool the window starts to be correct. Maybe some redraw signal?
Anyway, thank you @tmiw!

@drowe67 when you make a release (bb9558cc496b7b253fb54bb28486b0bcd7516ae5) you should tag it; in that way I can make a release on macports and you get the release on github releases.

tmiw:
@ra1nb0w, what version of macOS are you using? On 10.15.5 I see the following:

Light mode:
Screen Shot 2020-08-05 at 11 33 45 PM

Dark mode:
Screen Shot 2020-08-05 at 11 33 57 PM

The UI also seems to dynamically change as expected when I change modes in System Preferences->General.

ra1nb0w:
Sorry, I was too concise.
Setup:
macOS 10.15.6 19G73
Xcode 11.5 11E608c
wxWidgets-3.2 @3.1.4 (from macports)
freedv-gui at commit bb9558cc496b7b253fb54bb28486b0bcd7516ae5

the image that I attached was at the end of the startup without touching anything. If I move around the window after a while it becomes clean but if I open a config panel then the main window becomes dirty again. Always without starting the capture and in dark mode; with "white mode" everything works correctly. if you prefer I can take a short video.

tmiw:
@ra1nb0w, OK, I was able to duplicate it on mine too and just submitted a commit to fix. Let me know if that helps.

ra1nb0w:
fixed! thank you @tmiw!

drowe67:
Nice work everyone!  I've added a test plan at the top of this PR. I think you guys may already have tried out the OSX cases, if so pls fill in your results.

We'll have to figure out how to modify the `freedv-gui/docker` scripts to bootstrap wxWidgets 3.1, it currently uses the Fedora 3.0 package.

tmiw:
Considering that macOS dark mode support didn't seem to get added to wxWidgets until v3.1.3, I'd not recommend running FreeDV with v3.0 on newer versions of the OS. It should still be okay for older versions without dark mode, however. I'll see about building a binary with MacPorts' wxWidgets to test 3.0.

For Docker, we could just have the build download/build its own wxWidgets like what build_osx.sh does for Mac. The downside of course is that it'll overwrite whatever's already installed on the user's machine (or will dramatically increase the size of the freedv binary).

tmiw:
@drowe67, I was able to build with MacPorts' wxWidgets (v3.0.4) and it looks like dark mode definitely doesn't behave well with freedv-gui. However, the app itself looks okay in the default "light" mode. Perhaps we can just enforce 3.1 with CMake when building on Mac unless one is building on an old enough version of the OS?

ra1nb0w:
I maintain freedv-gui on macports ([portfile](https://github.com/macports/macports-ports/blob/master/science/freedv-gui/Portfile)).
I went back to wxwidgets-3.0 to avoid bug https://github.com/drowe67/freedv-gui/issues/67.
To avoid bug https://github.com/drowe67/freedv-gui/issues/27 I use
```
<key>NSRequiresAquaSystemAppearance</key>
<true/>
```

When 1.4.2 will be tagged I apply this patch and I move back to wxwidgets-3.2 and enable dark mode (version 3.1.4 contains many fixes and enhancement for macOS).

drowe67:
> @drowe67, I was able to build with MacPorts' wxWidgets (v3.0.4) and it looks like dark mode definitely doesn't behave well with freedv-gui. However, the app itself looks okay in the default "light" mode. Perhaps we can just enforce 3.1 with CMake when building on Mac unless one is building on an old enough version of the OS?

:+1:  Good idea @tmiw - I'm happy for you to make the call on OSX issues like this.  Feel free to add the code to enforce wxWidgets 3.1 for OSX in this PR.

@ra1nb0w I'm new to tags - will work out how to add it over the weekend.  I'm unusually busy with $dayjob atm so sorry I'm a bit slow to respond.  

A lot of Linux people will continue to build with distro-supplied wxWidgets (e.g. 3.0) so we just need to make sure that isn't broken.  For Windows we can probably just settle on any version (e.g. fix it like for OSX), as that is built by very few people and tends to get distributed as an executable.

How about we bump the freedv-gui version to 1.4.3 in this PR?

@hobbes1069 do you have any thoughts on this PR?

tmiw:
Thanks @ra1nb0w for the tip regarding NSRequiresAquaSystemAppearance. Based on that, I was able to update CMakeLists.txt to disable dark mode support (by generating the appropriate info.plist) if building with wxWidgets 3.0 and confirmed that the app works properly with both 3.0 and 3.1 on Mac. I also bumped up the version number to 1.4.3.

hobbes1069:
I've been following the conversation. All seems reasonable. Fedora is still on version 3.0.4, Rawhide (rolling devel branch) is on 3.0.5.1, so it doesn't look like we're jumping on the 3.1 train just yet.

ra1nb0w:
> I also bumped up the version number to 1.4.3.

conflict. rebase because upstream has version 1.4.2. In fact you are using and tree.

@drowe67 git tag is easy `git tag 1.4.2 bb9558cc496b7b253fb54bb28486b0bcd7516ae5` then you can create the release from github. You can also do everything from the github web interface.

Just a side note: I am not sure that creating a new tag from the latest commit is a good idea since it requires latest codec2 `freedv_get_hash()` and lpcnet `lpcnet_get_hash()`

tmiw:
@hobbes1069, Debian Bullseye is only on 3.0.5.1 too. Not sure there are any Linux distros with 3.1.

@ra1nb0w, conflict merged.

drowe67:
@tmiw I managed to build this PR against wxWidgets 3.0 and 3.1 for Linux, and the dialogs modified look and (with a basic test) work OK.  Can decode some audio files OK.

I ran into problems building for Windows, due a local Docker configuration issue. I don't have time to look into that right now. So I'll merge this PR and if any Windows problems emerge we'll deal with them later.

drowe67:
@hobbes1069 - we spoke recently on the mailing list about a release for 1.4.2 but I understand you didn't think we needed one?  However @ra1nb0w has requested a tag (at least). Can you pls work with @ra1nb0w to work out a way fwd?

@ra1nb0w - @hobbes1069 has been managing releases for freedv-gui.

ra1nb0w:
note: I am not sure that creating a new tag from the latest commit is a good idea since it requires latest codec2 freedv_get_hash() and lpcnet lpcnet_get_hash(). Otherwise you should also tag a new version for codec2 and lpcnet.

hobbes1069:
Tags can be arbitrary (non-version related) if we just need a known state of code to work with but I agree with @ra1nb0w that if we want to tag a version we'll need to create tags/releases for codec2 and lpcnetfreedv for a known compatible combination.

ra1nb0w:
I mean tag then github version.

hobbes1069:
Sure, just need a commit hash that gets you what you need and agree on a tag name :)

PR #73, opened 2020-09-12
Add support for changing CI-V address for Icom radios

tmiw:
This adds support (both in the PTT configuration UI and in the Hamlib class) for using a different CI-V address on Icom radios. This may be necessary for allowing usage of new radios not yet supported by Hamlib, as well as existing radios that need to use the non-default value for whatever reason.

(Speaking of unsupported radios, apparently the Mac build has the IC-9700 listed but the Windows build does not. @drowe67, does Docker potentially need to be updated to install a newer version of Hamlib? The Mac build process currently uses the latest master on GitHub, so we could possibly modify the Windows/Linux builds to use a similar mechanism.)

Tests run so far (Mac only right now on a IC-7200):

- Ensuring that the PTT configuration UI only supports A-F and 0-9 for valid characters.
- Canceling out of the PTT configuration UI without modifying anything.
- Select non-Icom radio (Radio Address field becomes hidden).
- Select Icom radio (Radio Address field appears).
- Testing PTT settings with the following:
    - CI-V address of 00 (radio allows connection regardless of its actual address)
    - CI-V address equal to the radio's (radio connection allowed)
    - CI-V address not equal to the radio's (FreeDV complains of a Hamlib connection timeout)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -397,7 +397,7 @@ if(WXRC)
 endif(WXRC)
 set(WX_VERSION_MIN 3.0.0)
 set(WX_31_VERSION_MIN 3.1.3)
-find_package(wxWidgets REQUIRED core base aui html net adv)
+find_package(wxWidgets REQUIRED core base aui html net adv propgrid)
 execute_process(COMMAND sh "${wxWidgets_CONFIG_EXECUTABLE}" --version
     OUTPUT_VARIABLE WX_VERSION)
 string(STRIP ${WX_VERSION} WX_VERSION)
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -168,6 +168,12 @@ Hamlib comes with a default serial rate for each radio.  If your radio
 has a different serial rate change the Serial Rate drop down box to
 match your radio.
 
+If using an Icom radio, Hamlib will use the radio's default CI-V address
+when connecting. If this has been changed, you can specify the correct
+address in the "Radio Address" field (valid values are 00 through FF
+in hexadecimal). Note that "00" is the "wildcard" CI-V address and will 
+also work if there are no other Icom/CI-V capable devices in the chain.
+
 When **Test** is pressed, the "Serial Params" field is populated and
 displayed.  This will help track down any mismatches between Hamlib
 and your radio.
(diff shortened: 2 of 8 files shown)


hobbes1069:
FYI, Hamlib 4.0 should be released "soon". There's already been a RC1.

tmiw:
@hobbes1069, assuming their master is what 4.0 will ultimately be, it seems like FreeDV won't need any code modifications to work with it. So that's good at least :)

drowe67:
@tmiw nice work :+1:  and I appreciate the list of tests you have performed.  Looks like it's building for Linux OK as Travis is happy :slightly_smiling_face: 

At some stage we should build for Windows and ask some Windows users to try it out, perhaps after merge to collect a few recent changes in one Windows experimental release.

@tmiw:
1. Could you pls add a few lines to the User Guide on this feature and how to use it?
1. Have you tested any non-Icom radios?  I'd like to make sure this PR doesn't accidentally break HamLib PTT for any other radio.

tmiw:
@drowe67, I currently have a few VHF/UHF radios as well as the FT-817, so I'll give it a shot with those.

Speaking of VHF/UHF, I think I also want to make some changes to the sideband detection logic to allow FM if using 2400B. I'll do that in a separate PR if that's okay?

tmiw:
@drowe67, I updated the User Manual and made a minor tweak to the code in response to testing with my FT-817 (original, not ND). Seems that either rig_set_conf() for "civ_addr" isn't as transparently ignored for non-Icom radios like I thought or there was something weird that happened with the previous binaries I had on my system. In any case, not executing that function unless the radio is Icom is probably good practice anyway.

drowe67:
Looks good @tmiw - good to see you have it working OK with non-Icom radios too.  I'm happy the PR (and especially your testing), so pls let me know when you are finished and I will merge.  

Unfortunately I'm a bit busy atm due to my day job so can't test it myself right now, but I am satisfied your testing will have caught any big issues. When I get some time I'll build an up to date Windows version to capture the recent changes from this and one or two other PRs.

tmiw:
@drowe67, there are maybe some UI changes that would be nice, but those can be done in another PR. It's probably more important to get wider testing on the current changes.

PR #74, opened 2020-09-13
Allow FM for mode detection if using 2400B.

tmiw:
@drowe67, as mentioned in PR #73, here's the change to allow FM if using 2400B. With this change, if 2400B is chosen and the radio is in the 10 meter FM subband or higher, "FM" will not show as red. Tested on my FT-817 and seemed to work with no issues.

### DIFF
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -2445,7 +2445,7 @@ void MainFrame::OnExit(wxCommandEvent& event)
     // of in the destructor due to its need to touch the UI.
     if (wxGetApp().m_hamlib)
     {
-        wxGetApp().m_hamlib->disable_sideband_detection();
+        wxGetApp().m_hamlib->disable_mode_detection();
     }
 
     //fprintf(stderr, "MainFrame::OnExit\n");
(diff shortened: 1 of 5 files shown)


src/fdmdv2_main.cpp
```diff
@@ -2635,7 +2635,7 @@ bool MainFrame::OpenHamlibRig() {
         wxMessageBox("Couldn't connect to Radio with hamlib", wxT("Error"), wxOK | wxICON_ERROR, this);
     else
     {
-        wxGetApp().m_hamlib->enable_sideband_detection(m_txtSSBStatus);
+        wxGetApp().m_hamlib->enable_sideband_detection(m_txtSSBStatus, g_mode == FREEDV_MODE_2400B);
```
drowe67, line 2638: Maybe change the function name to `enable_mode_detection()` ?

drowe67:
Looks good @tmiw - I re-read the original sideband detection PR and am still happy with the design choice to make it fairly benign by just sampling (e.g. not too much polling) :+1:

src/fdmdv2_main.cpp
tmiw: Done. 👍

tmiw:
@drowe67, just pushed the renaming changes. And yeah, it's probably for the best considering the seemingly inconsistent radio/OS support for the non-polling method.

PR #75, opened 2020-11-16
Mac universal binary support for FreeDV

tmiw:
Adding support for ARM on FreeDV per https://github.com/drowe67/codec2/pull/154 and https://github.com/drowe67/LPCNet/pull/36. Requires -DBUILD_OSX_UNIVERSAL=1 to enable. Let me know if you have any questions!

Note: Looks like FreeDV broke on Big Sur (PortAudio errors on start). It's happening on the non-universal version, too, so it doesn't seem related to this change. I'll investigate further and create a separate PR for that issue.


tmiw:
@drowe67, looks like pointing CMake to the latest Git snapshot of PortAudio resolves the issue mentioned in OP. I included that change in this PR.

drowe67:
@tmiw nice work :+1: Looks like all the changes are OSX specific so it shouldn't break anything fpr other operating system builds.  Let me know when you ready to merge :slightly_smiling_face:

tmiw:
@drowe67, I did some on the air testing earlier tonight (albeit on an x86 MBP) and everything seems to work properly now. I'm fine with merging right now and creating additional PRs if ARM users end up running into issues.

BTW, out of curiosity, would I be able to force FreeDV/Codec2 to allow 2020 without AVX extensions? I wonder if the just-released Macs would be able to handle it without additional updates. (2020 does show as possible on x86 Macs and disabled on ARM.)

drowe67:
> BTW, out of curiosity, would I be able to force FreeDV/Codec2 to allow 2020 without AVX extensions? I wonder if the just-released Macs would be able to handle it without additional updates. (2020 does show as possible on x86 Macs and disabled on ARM.)

Yes you'd need to disable the AVX test in the freedv-gui source, and build freedv-lpcnet with the non-AVX Cmake flag.  As a sanity check, you try LPCNet on the command line first .  I think it's just real time on my laptop without AVX.

tmiw:
@drowe67, I did some more investigation tonight and it turns out that the universal binary version of LPCNet/Codec2 doesn't actually have AVX enabled by virtue of where it was built. Thus, it was as simple as ensuring that the main application didn't disable the option.

Anyway, it looks like 2020 decodes the sample .wav file with no issues on ARM Mac. ~60% CPU usage per Activity Monitor but if it's anything like on Intel Macs, that's just one core. Maybe we can consider whitelisting ARM Macs for 2020?

drowe67:
@tmiw OK that's an encouraging start.  Have you run the complete freedv-gui with 2020 on a ARM Mac?  Having said that, the 2020 decode will dominate most of the CPU.

Must be a fast platform. Sure, you are welcome to try white listing those machines if you like.  We'd need to add some explanation to the manual, and make it clear this is pretty experimental.

tmiw:
@drowe67, yep, full FreeDV UI. I ran test_vec in LPCNet and it seems that the NEON optimizations got enabled, which probably helped a fair bit. 

Anyway, I'll probably need to do another PR on LPCNet to also get AVX reenabled on the x86 side but after that, I'll see about whitelisting the ARM Macs in the main UI.

tmiw:
@drowe67, okay, needed changes to whitelist ARM Macs committed. Also, I have a binary up at https://www.dropbox.com/s/jnm0zn31xh35hp6/FreeDV-6374421.dmg?dl=0 for anyone else who wants to test.

drowe67:
> @drowe67, yep, full FreeDV UI. I ran test_vec in LPCNet and it seems that the NEON optimizations got enabled, which probably helped a fair bit.

OK cool - well that's SIMD acceleration, just like AVX.

> Anyway, I'll probably need to do another PR on LPCNet to also get AVX reenabled on the x86 side but after that, I'll see about whitelisting the ARM Macs in the main UI.

Would it be worth testing for neon?

tmiw:
> @drowe67: Would it be worth testing for neon?

Not sure. It seems to be getting enabled with LPCNet for the Raspberry Pi but I thought people haven't been able to get adequate performance on it yet? I could be remembering wrong, though.

drowe67:
@tmiw - that's an excellent point - we don't want it running on just any ARM machine, so the approach you have settled on looks good :+1:

PR #24, opened 2019-08-10
Improved USB sound device support

drowe67:
Windows often lists the same sound device under two names.  This PR removes duplicates from the list, and can also detect re-ordering of USB sound devices.

This was prototyped a few months ago and works great except for a compatibility issue with Virtual Audio Cable (VAC) software.  This PR seeks to resolve the remaining issues.

PR #77, opened 2020-12-13
Add PSK Reporter support + UX enhancements to FreeDV

tmiw:
This PR adds PSK Reporter reporting support to FreeDV. Requires Hamlib to be enabled to actually turn on--regardless of the PSK Reporter settings in the Options dialog--due to a dependency on the current frequency. Very much still a WIP but figured I'd open the PR for discussion/testing.

EDIT (28 December 2020): The scope of this PR has massively expanded and now includes various UX enhancements as well as fixes for bugs encountered during testing of the PSK Reporter feature.


tmiw:
@drowe67, it's possible that this may require RX checksum support to be anywhere close to 100% reliable. For instance, stderr showed the following earlier (testing using audio recorded from a local KiwiSDR):

```Adding callsign K6Aeo @ SNR 3 to PSK Reporter. ```

(Note: not my actual callsign :)

Any idea why it's compiled out in the code by default?

Additionally, I *think* the regex I'm using for the callsign extraction is correct per stuff I've read online but maybe we can tighten it up a bit?

```
wxRegEx callsignFormat("(([A-Za-z0-9]+/)?[A-Za-z0-9]{1,3}[0-9][A-Za-z0-9]*[A-Za-z](/[A-Za-z0-9]+)?)");
wxString wxCallsign = m_txtCtrlCallSign->GetValue();
if (callsignFormat.Matches(wxCallsign) && wxGetApp().m_pskPendingCallsign == "")
{
        wxString rxCallsign = callsignFormat.GetMatch(wxCallsign, 1);
        wxGetApp().m_pskPendingCallsign = rxCallsign.ToStdString();
        wxGetApp().m_pskPendingSnr = snr_val;
}
```

**Comment by tmiw** [comment]
FYI, here's the data that's actually sent to PSK Reporter per [their test server](https://pskreporter.info/cgi-bin/psk-analysis.pl):

Screen Shot 2020-12-12 at 10 24 24 PM

I've added the user's current mode to the "decoderSoftware" field which should show up on [the map](https://pskreporter.info/pskmap.html). I'm open to other ways of formatting this field if that's preferred.

API info [here](https://pskreporter.info/pskdev.html).

**Comment by drowemyriota** [comment]
@tmiw Cool - well done for having a go at this one.  Hmm, I can see how having free form text will make this tricky, depends on what people put in their tx string.  Hopefully the regex will pick it out.  Alternatively we could add a form for people to fill in, e.g. force the callsign to be in the first x characters or something like that.

From memory forcing checksums meant many txt strings were never printed, as due to channel errors the checksums were often never correct.   Also users tend to prefer (even errorr-ed txt) being printed out in real time. But it's been some time, I can't recall exactly.

For the feature to work, it might be worth introducing some simulated channel errors and find out how often you get a valid txt string, that's good enough to send to PSK reporter.  I think you can add AWGN noise on the Tools menu.

Future modes should probably put the txt string inside the FEC protected bits.  Or we could add some FEC as part of the text string.

**Comment by drowe67** [comment]
@tmiw Doh - used the $dayjob github account for the comment above - same laptop atm

**Comment by tmiw** [comment]
@drowe67, some hidden FEC will help with callsign decode for sure. I lean towards having the callsign always be at the beginning but I'm not sure what other users think. Maybe a discussion for the mailing list?

**Comment by tmiw** [comment]
@drowe67, question: any way to have Travis use C++11? Looks like that's the main reason why it's failing right now. Alternatively, if it'll be easier, I can go through pskreporter.cpp and remove the C++11isms.

**Comment by drowe67** [comment]
@tmiw sure - good question to ask on either/or digital voice or codec2 mailings lists.

You could put callsign/grid ref in fixed width fields, then have a free form txt bit.

Checksum or FEC **just** for the callsign/grid ref will make probability of sucesssful detection much more likely.

**Comment by drowe67** [comment]
> @drowe67, question: any way to have Travis use C++11? Looks like that's the main reason why it's failing right now. Alternatively, if it'll be easier, I can go through pskreporter.cpp and remove the C++11isms.

You could try a later Ubuntu release maybe?  Or would a change in CFLAGS set up gcc for C++11 ?

**Comment by tmiw** [comment]
@drowe67, sent email and updated the Travis config to use Bionic (18.04 LTS). Will see how it goes.

**Comment by tmiw** [comment]
I checked in a minor change that seems to make reporting the callsign a bit more reliable and tested with audio recorded from a WebSDR (both test and live PSK Reporter servers). Here's what the report looks like on PSK Reporter itself:

Screen Shot 2020-12-13 at 2 31 35 PM

Unfortunately I don't see "DIGITALVOICE" as a drop-down option but maybe it'll appear the more people report FreeDV to it.

**Comment by drowemyriota** [comment]
@tmiw looking pretty cool!  I haven't seen any comments on those checksum/fixed field questions.  Any feeling for how you would like to proceed with those design questions?  We can always just try a first pass - and see how that goes, and indeed if there is anyone using the feature, then choose to invest more time in it later.

**Comment by tmiw** [comment]
@drowe67, a few updates:

* I got confirmation from the guy running PSK Reporter that "FREEDV" is okay to use, so I updated that in the code. That should mean that the mode won't be truncated on the map.
* Added documentation about PSK Reporter reporting and its limitations.
* I made further tweaks today (including fixing a bug with the extraction logic discovered by playing audio back that was recorded through a better WebSDR). Based on letting it run for a few hours on 14.236, I didn't notice any false reports when there was no active signal--including when analog QSOs were in the passband. That said, there also wasn't really any FreeDV activity on 20 meters today, either.

Anyway, I wouldn't mind having people beta test it before the PR's merged. I'll see about getting a Windows build made that can be posted to the list.

**Comment by tmiw** [comment]
@drowe67, okay, so I managed to get the changes to build in Docker. However, I get this when I try to run it in my Windows 10 VM:

Screen Shot 2020-12-14 at 12 02 42 AM

Not sure what's going on offhand. FWIW, I updated the Docker config files to use Fedora 33 before building.

**Comment by drowe67** [comment]
@hobbes1069 do you have any thoughts on the issue @tmiw has hit above?

@tmiw could you perhaps try building with the Fedora 30 Docker system as a starting point?

**Comment by tmiw** [comment]
@drowe67, @hobbes1069, I'm actually having network issues with Docker now:

```Errors during downloading metadata for repository 'fedora-cisco-openh264':
  - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-33&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
Error: Failed to download metadata for repo 'fedora-cisco-openh264': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-33&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]``` 

I'll need to resolve those before I can try with an older Fedora version.

**Comment by drowe67** [comment]
@tmiw I've built a 64bit Windows version here http://www.rowetel.com/downloads/freedv/FreeDV-1.4.3-devel-20201214-650902f-win64.exe using the Fedora 30 config.

**Comment by tmiw** [comment]
@drowe67, thanks! That one seems to run with no issues on my VM. I'm guessing I still have inconsistent files floating around despite supposedly purging everything in Docker (not to mention the networking issue). I'll pass that link along to Walter K5WH and a few other people using the Windows version so they can test.

**Comment by tmiw** [comment]
@drowe67, so far so good:

Screen Shot 2020-12-14 at 12 05 34 PM

It did report some invalid callsigns during that test but it looks like PSK Reporter does the right thing and ignores them, FWIW.

**Comment by drowe67** [comment]
@tmiw while it's good that PSKreporter ignores invalid callsigns, it would be great if our design avoiding sending invalid data to the site.  However that probably means a checksum/fixed field for callsigns.

**Comment by hobbes1069** [comment]
> @drowe67, okay, so I managed to get the changes to build in Docker. However, I get this when I try to run it in my Windows 10 VM:
> 
> Screen Shot 2020-12-14 at 12 02 42 AM
> 
> Not sure what's going on offhand. FWIW, I updated the Docker config files to use Fedora 33 before building.

Some kind of library API/ABI mismatch but couldn't guess as to how it happened or how to fix it. I confess I don't know much about docker.

**Comment by tmiw** [comment]
@hobbes1069, my guess is that I just need to wipe everything and fix the networking issues and then I'll be able to make working builds.

@drowe67, I wonder if maybe there should be a higher SNR minimum for PSK Reporter reporting. Right now I'm using the current squelch level as the minimum but perhaps the minimum should be that or some positive number (whichever is higher). For instance, I believe K5WH was coming in at around +3 or so according to PSK Reporter above.

Also, a minimum sync time period might help too. How long does it take to transmit the maximum number of characters allowed in the text field? Maybe requiring that the transmission is at least that long to report will ensure we have the full text.

In any case, the ultimate solution will be FEC/checksums as you've mentioned.

**Comment by tmiw** [comment]
OK, looks like I had to follow the instructions in the accepted answer from [here](https://superuser.com/questions/1130898/no-internet-connection-inside-docker-containers) to reset networking inside Docker and permit docker-compose to rebuild the image.

Anyway, I can produce working builds with Fedora 31 but not 33. I imagine it's somehow packaging a different wxWidgets DLL than the one used for compilation but that might be something that can be fixed later.

**Comment by drowe67** [comment]
> @drowe67, I wonder if maybe there should be a higher SNR minimum for PSK Reporter reporting. Right now I'm using the current squelch level as the minimum but perhaps the minimum should be that or some positive number (whichever is higher). For instance, I believe K5WH was coming in at around +3 or so according to PSK Reporter above.
> 
> Also, a minimum sync time period might help too. How long does it take to transmit the maximum number of characters allowed in the text field? Maybe requiring that the transmission is at least that long to report will ensure we have the full text.
> 
> In any case, the ultimate solution will be FEC/checksums as you've mentioned.

Problem is SNR can bounce all over the place, so it's a pretty rough guide.  The bit rate is pretty slow, I think it's about 10 bits/s for the 700D text info (2 bits/ frame, 180ms frames).

**Comment by tmiw** [comment]
@drowe67, hmm, assuming the widest possible callsign in normal use (3 letters/numbers + 1 number + 3 letters), that would mean a ~6 second minimum sync time before we can start trying to extract a callsign. That should be reasonable but would need more testing to be sure.

**Comment by drowe67** [comment]
Hi @tmiw - just ping me when you feel ready to merge this feature :slightly_smiling_face:

**Comment by tmiw** [comment]
@drowe67, I'm thinking I want to update the Hamlib code to allow the current frequency to be refreshed from the radio after callsign extraction. As it is now, it'll grab the frequency once on start and cache it until the user pushes Stop. I don't know about anyone else, but I tend to move the VFO a few KHz up and down to avoid QRM, so keeping the reporting accurate in that case would be good.

Also, I'm giving more thought into enforcing a minimum sync time before trying to extract callsigns. However, it seems to be working well enough as is that it's something we can add later if it turns out to be necessary.

Anyway, I'll work on getting a build with the former made tonight for additional testing.

**Comment by tmiw** [comment]
@drowe67, I just saw your announcement. Let's wait on merging until 700E goes into master, then I can pull it down into my branch and update this PR to reflect its presence.

**Comment by drowemyriota** [comment]
@tmiw Cool - I'll merge the 700E branch in couple of days once a few more people have used it.  It's not under active development - I just did the minimum to allow 700E to be fired up.

In the mean time pls feel free to merge the 700E branch into this one - I think the changes are orthogonal.  You can even push out some more test versions if you like.  Maybe bump the version to 1.5.1

**Comment by drowe67** [comment]
@timw how about hooking up the checksum support for this feature?  Some ideas:
1.  Callsigns only get PSK-reported if the checksum is OK.
1.  I can't recall where I left the checksum code but one annoying part of it's previous incarnation was the text only got displayed if the checksum was OK.  So the text field was often blank or not updated for a long time. Instead we could modify such that text was printed as it arrived (including errors), but PSK reporter only received the string if the checksum was OK.

**Comment by tmiw** [comment]
@drowe67, I got a chance to enable the existing checksumming code tonight. There were a few compiler errors and wxWidgets assertions to resolve and some temporary disabling of UI elements in the Options dialog needed. I transmitted a nearly minute long transmission and recorded myself through a local KiwiSDR and basically ended up with the following:

```
rx callsign: e6AQ San DieboCA, USA c.
  checksum tx: 0c checksum rx: 3d
rx callsign: K6AQ San Diego- CA, USA ce5.
  checksum tx: e5 checksum rx: e6
rx callsign: X6AQ San Diego, CA, USA ce5.
  checksum tx: e5 checksum rx: f2
rx callsign: K6AQ San Diego, CA, USA ce5.
  checksum tx: e5 checksum rx: e5
rx callsign: KjAQ San Diego, CA, USA ce5.
  checksum tx: e5 checksum rx: 19
```

I'm not sure I'm a huge fan of needing to transmit for an entire minute to get a usable callsign, to be honest--and this was using groundwave. The checksum used in the code also seemed pretty basic; it adds all the bytes together and takes the least significant 8 bits. Might need to think on this more.

**Comment by tmiw** [comment]
Hmm, what about encoding the callsign using [Hamming(63,57)](https://en.wikipedia.org/wiki/Hamming_code)? There are 37 possible symbols for callsigns (26 letters + 10 numerals + forward slash), requiring 5 bits to encode. Typical maximum length (6 symbols) would mean 30 data bits, so we'd have another 17 bits to work with for future use. This way, we can treat the entire 63 bit block as a single unit and throw it out if we can't correct any errors.

**Comment by drowe67** [comment]
@tmiw yes I agree a minute is too long.

Re the use of FEC, how would be obtain "frame sync" on the codeword (ie determine it's location in the string of varicode bits)?  I guess we could put it in a known position, or try a  decode after every bit is received.

**Comment by tmiw** [comment]
@drowe67, I was going to say that we could just have the callsign always be at position 0 in the text, but I'm not sure how that'll work for people using older versions. At least with 1.5.1 and onward, we'll know about the new encoding (maybe always generate a 63 bit block of zeroes at the beginning if PSK Reporter support is disabled?) and be able to suppress its display.

**Comment by tmiw** [comment]
@drowe67, I just checked in a change to resolve the Icom CI-V issue from the digitalvoice Google group. Should I also include that change in another PR so that it can go up to master more quickly?

**Comment by drowe67** [comment]
@tmiw so will people using older version just see some odd characters?  That's probably acceptable - they are used to see characters received with errors in normal use.  Easy for them to upgrade too.

**Comment by tmiw** [comment]
@drowe67, hmm, I think I made a mistake in my math. We'll need 6 bits to encode all possible symbols in callsigns (5 bits would be a maximum of 31, which is less than the 37 needed). Thus, Hamming(63,57) would limit us to 9 symbols. Which, while enough for normal callsigns, would preclude many prefixes and suffixes (for instance, KA1BCD/QRP would be too long for a single block). The next one up is Hamming(120,127), which would let us store 20 symbols in a block--more than enough for any callsign+prefix+suffix combo but would also double the transmission time. Then again, ~12 seconds at 10 bit/sec is still significantly better than having to transmit for an entire minute.

Anyway, the expectation with FreeDV is that you still ID with your voice, right? The text field is secondary at best for ID purposes so maybe it doesn't matter that prefixes and suffixes can't be used? Thoughts?

(BTW, I have a working Hamming encoder/decoder class now that's fairly generalized, so using a different one should just be a matter of changing some constructor arguments. And it can do the extra parity bit for double bit error detection, too, if so desired.)

**Comment by drowe67** [comment]
@tmiw Hmm, this is a tricky one!  In hindsight I wished I'd placed the txt bits **inside** the FEC codeword we use for the payload voice data (on some modes).  Something to perhaps consider for new modes.  Yes I understand most countries require you to ID with your voice, the txt field is just a bonus.

To help brainstorm a solution perhaps we could set some requirements for this feature - pls feel to comment/edit:

1. We want the current txt system to work more or less as it does in terms of update rate (latency) and we don't care if there are a few bit errors in the free form txt. 
1. However it would be OK if we build a new txt system that breaks with txt from the current system, as (i) it's easy upgrade FreeDV and (ii) txt is just a secondary feature.
1. How often to we actually need to fire off a "spot" to PSK Reporter?  Would once per QSO be acceptable?
1. We want a valid call sign to send to PSK Reporter.
1. We don't want to spam PSK Reporter with too many reports, especially with bogus callsigns.
1. We should receive a valid callsign XX seconds after an over starts.  Hmm, is this essential? Is it OK if we get at least one valid callsign **some time** during a  QSO?
1. A variable length callsign string is desirable, fixed lengths are awkward in cases like KA1BCD/QRP.

One solution:
1. Ditch the current txt system.
1. Replace it with FEC protected codewords that can be cascaded to get any length string.
1. Terminate the decoded string with a checksum so we are super sure.
1. Make each codeword a few bytes long say (24,12) , so we get a real-time-ish update every few seconds, in chunks of two 6-bit characters.  Common Golay code too.

**Comment by tmiw** [comment]
@drowe67, to answer your questions/comments:

1. PSK Reporter's [developer documentation](https://pskreporter.info/pskdev.html) says that applications should be sending a packet every five minutes (or once it fills up, which I'm not sure would be likely for FreeDV given the typical size of each signal report in said packet and how often we'd likely see unique callsigns within that timeframe). I'm not sure how applications like WSJT-X do it exactly but I do know that I start seeing reports pretty quickly after sending a CQ on FT8 (within a minute or two at most?), so I imagine we'd want to at least queue up the packet as soon as we drop sync. The "seeing updates within a minute or so" deal probably won't come until there are a lot more users listening on 14.236 and elsewhere.
2. We'll probably need to figure out how often prefixes/suffixes are used. I know they're basically mandatory if you operate from another country than where you're licensed and you use your licensed callsign (vs. one granted for e.g. a DXpedition), but I feel like I don't see non-country ones all that often in general.
3. A fixed length was initially proposed as an attempt to make it more reliable by permitting an "all or nothing" approach. We could use smaller block sizes and just concatenate the decoded blocks together, but then we'd have to figure out how to handle one or more blocks becoming invalid during the transmission. And synchronization probably becomes more important as well.
4. Another plus for larger blocks: relatively more bits end up being used for data as opposed to parity (at least for Hamming, unsure about other methods), reducing performance loss. See the table in [this Wikipedia article](https://en.wikipedia.org/wiki/Hamming_code#Hamming_codes) for details on that. 
5. How would we handle UW becoming corrupt? Maybe duplicate it a few times and accept it when the receiver sees more than 50% of them?

**Comment by drowemyriota** [comment]
@tmiw thanks, good comments :+1: 

What I like about short code words is a psuedo-real time (every few seconds) update of text, which the end users seem to like.

If the UW is corrupt we don't get codeword frame sync and have to wait until the next UW to see **any** text.  Hmm, and that could take a long time, so that wont work :thinking: 

I think the problem of codeword sync exists for any FEC we use.  Varicode is neat in that it self re-synchronises after a bit error.

Another couple of ideas:
1.  Drop the occasional voice codec frame - replace it with packet data.  It will have the benefit of full FEC protection.  For 700D they are 112 bits of payload data, 56 for 700E.  Something similar is done for VHF Data modes.  We could use an inverse UW to signal data rather than voice.  The end user won't notice one 180ms voice packet missing say every 30s.  Keep the current varicode text channel running on the side for a visual update and backwards compatibility.
1. Transmit the FEC parity bits and a UW at the end of a "text codeword" or text line.  So the end user sees the regular text update, but every minute or so they get the codeword FEC parity bits and a UW to mark where they are.  The FEC codeword then runs on the data bits received already plus parity bits at the end.  The end user sees the current txt string get "corrected". If the UW is not detected, or FEC decode fails, then no valid string is obtained and we don't post a PSK Report.  We get another chance in a minute or so.
1. Your comments and the two points above seem to be converging on a fixed length, but long-ish codeword.  Say 30 or 60 seconds at 10 bit/s.  We could make it fixed length, and even use LDPC codes.  If it's a high rate code (say 0.8), the visual pause in txt update will be minimised.

**Comment by tmiw** [comment]
@drowe67, I took a look at the varicode stuff and it seems that it allows "normal" 7 bit ASCII. The current text system also seems to have an 80 character maximum, but there's nothing stopping FreeDV from having multiple 80 character buffers that it can switch between. Out of curiosity, I'm thinking of trying Hamming(7,4) (leaving the MSB of the resulting encoded byte as 0) and seeing what happens. Definitely not ideal, but might get us something good enough for the time being while we work on longer-term changes to the text system in general.

As for replacing one of the voice frames with data, 180ms might be enough of a gap to be detectable by ear. Then again, that could also happen due to HF propagation changes/QRM/etc., so it wouldn't be too out there.

**Comment by tmiw** [comment]
@drowe67, so I added Hamming(7,4) encoding to the text field. Seems to work okay when recording/playing back from the modulator but the local KiwiSDR wasn't cooperating today to see how it actually does on the air. I'll make new builds and email them out for testing.

**Comment by drowemyriota** [comment]
@tmiw OK cool idea hope it works well :+1:   So this would be a little slower than the regular varicode based txt due to the larger number of bits/character?  

Would it be possible to use varicode with the Hamming code to make it run a little faster?  Perhaps that would introduce it's own sync issues.

BTW we also have the Golay (23,12) code available in Codec 2 if that's useful.

**Comment by tmiw** [comment]
@drowe67, Hamming looks like it didn't help much, at least with the local KiwiSDR I usually use to listen to myself on the air when 20 meters isn't propagating :) I was actually in the process of switching it over to use the data frame approach but it looks like the callbacks are only used for VHF/UHF modes and not for 700D, etc. Would need to figure out what kind of data the incoming frame has in order to call the correct function for it.

Anyway, maybe Golay might work better. I'll have to take a closer look.

**Comment by drowemyriota** [comment]
@tmiw did you manage to work out why the Hamming system didn't work?  It might be worth running it to ground just to be sure.  Very easy to have a simple bug creep in.  It might also be useful to unit test it outside of FreeDV.  For example inject random bit errors, and make sure it works with no errors or a modest BER.  You could even inject burst errors typical of HF at the bit-stream level.

Also - this does seem to be turning into a major project - don't feel you need to work on it forever if it's getting too complex! But if you're having fun that's fine :slightly_smiling_face:

**Comment by tmiw** [comment]
@drowe67, so yeah, I think Golay is working a lot better. At the very least, I'm actually getting usable text decodes from the local KiwiSDR now, even if they're not that common. (I suspect KiwiSDR has issues that make FreeDV testing more difficult than it should be, since I recall WebSDR working better.) I'll generate some new builds and link them on the mailing list.

**Comment by tmiw** [comment]
@drowe67, so what I'm noticing is that text decoding with Golay seems reliable (for 700D at least) if we're at around 5-6 SNR (tested using a WebSDR ~400mi away on 40 meters). I'm not sure how much lower we can go before the text becomes a mess but maybe we can require a SNR minimum before scanning for callsigns? 

I'm also finding various FreeDV bugs while doing this testing/development, so I'll probably have more commits tonight with fixes for those.

**Comment by drowe67** [comment]
@tmiw nice work on the bug fixes :+1:  

I wouldn't recommend using SNR as a threshold, as it's an unreliable measure - especially on fading channels.  A checksum is a good way to determine if you have a valid callsign.

I'd also suggest breaking the system out as a module of code that can be unit tested.  For example see how well it works at a few percent random BER, make sure the Golay FEC is actually working as expected.  It's really hard to get objective results from a real world  HF channel, e.g. it's hard to repeat a test.  If you make the system part of the FreeDV API, we can unit test it through simulated (and repeatable) HF channels using the 700D modem.

Is it worth breaking the bug fixes out into a separate PR?

**Comment by tmiw** [comment]
@drowe67, if it takes much longer then yeah, I'd split out the bug fixes. What I'm thinking is the following:

1. Split out the callsign encode/decode logic into a separate class for later UT/enhancement (as suggested).
2. Update the encode/decode logic to use the new code *only* if PSK Reporter is enabled. Otherwise, we use pre-1.5.1 behavior and send/receive text without any special encoding. 
3. If PSK Reporter is enabled, ignore the existing user-configured text field during encoding. On TX, *only* send the callsign along with CRC8; decode during RX will only extract callsigns if CRC is correct. If we get sync on that field, we display everything even if the CRC is incorrect.

With all the above, I think it'll be okay enough for merge if you agree.

**Comment by tmiw** [comment]
@drowe67, OK, I emailed out the latest builds. Assuming there's no major issues, we can probably merge it up for now and just fix issues/improve the feature over time.

**Comment by drowe67** [comment]
Thanks @tmiw - quite a lot of code has changed so over the next few days I'll review the changes and run the Linux version myself.  I can't recall experiencing any issues with play/record file myself, perhaps it only happened on other OSes.  There was an issue with record file length I never solved.

You've done a lot of work!  Well done :+1:

**Comment by drowe67** [comment]
Hi @tmiw - I've built freedv-gui from this PR and it's running fine here on some test samples:

1.  I like the way yo have made Stop/Start change "on the fly" :+1: 
2. The vertical alignment of the  "Callsign" and "Grridsquare" static test fields isn't quite right on Ubuntu (this sort of thing confuses me all the time on WxWidgets BTW!):
   Screenshot from 2020-12-28 06-17-44
1. Also given PSK reporter and TxtMsg are closely related, how about moving the PSK Reporter box up to just under the TxtMsg Box?
1. I edited the PSK reporter section of the User Manual -see what you think :slightly_smiling_face:

File: src/dlg_audiooptions.cpp [hunk 1]
```diff
@@ -119,7 +119,7 @@ AudioOptsDialog::AudioOptsDialog(wxWindow* parent, wxWindowID id, const wxString
     // Rx Out -----------------------------------------------------------------------
 
     wxStaticBoxSizer* sbSizer3;
-    sbSizer3 = new wxStaticBoxSizer(new wxStaticBox(m_panelRx, wxID_ANY, _("From Computer To Speaker/Headphones")), wxHORIZONTAL);
+    sbSizer3 = new wxStaticBoxSizer(new wxStaticBox(m_panelRx, wxID_ANY, _("Output From Computer To Speaker/Headphones")), wxHORIZONTAL);
```
drowe67, line 123: Interesting innovation :thinking: - thanks for also modifying the user manual to match!  I'm open to ideas here, TBH I'm not sure how best the convey this issue to new users.  It's certainly a source of confusion.

src/dlg_filter.cpp
```diff
@@ -34,7 +34,7 @@
 #define MAX_GAIN           20
 
 #define MAX_LOG10_Q         1.0
-#define MIN_LOG10_Q        -1.0 
+#define MIN_LOG10_Q        -1.0
```
drowe67, line 38: Do you use the Atom editor?  I've just started using it and it's always making these sort remove-the-space of commits for me....

drowe67: Ahh - wait a minute - this might be the changes i made for the 700E PR :slightly_smiling_face: ?

src/fdmdv2_main.cpp
```diff
@@ -1163,70 +1195,72 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
...
     }
 
-
+    // We should only report to PSK Reporter when all of the following are true:
+    // a) The callsign encoder indicates a valid callsign has been received.
+    // b) We detect a valid format callsign in the text (see https://en.wikipedia.org/wiki/Amateur_radio_call_signs).
+    // c) We don't currently have a pending report to add to the outbound list for the active callsign.
+    // When the above is true, capture the callsign and current SNR and save it in a temporary location.
+    // Once sync is lost, add to the PSK Reporter object's outbound list.
+    if (wxGetApp().m_pskReporter != NULL && wxGetApp().m_callsignEncoder != NULL)
```
drowe67, line 1231: Good logic and ell worth documenting :+1: 

I can see is tweaking this over time as we get some real life experience with the feature.

drowe67:
@tmiw  probably a bit late but another way to test this is use Full Duplex and Tx to yourself on the same machine.  You can even inject channel noise from Tools-Options.  I use full duplex quite a bit when developing.

tmiw:
> Hi @tmiw - I've built freedv-gui from this PR and it's running fine here on some test samples:
> 
> 1. I like the way you have made Stop/Start change "on the fly" 👍

Thanks!

> 2. The vertical alignment of the  "Callsign" and "Gridsquare" static test fields isn't quite right on Ubuntu (this sort of thing confuses me all the time on WxWidgets BTW!):
> 3. Also given PSK reporter and TxtMsg are closely related, how about moving the PSK Reporter box up to just under the TxtMsg Box?

Confirmed in my Ubuntu VM. I went ahead and tweaked the spacing on the rest of the dialog while I was at it, too, and moved the PSK Reporter box:

Screen Shot 2020-12-27 at 3 44 31 PM

> 4. I edited the PSK reporter section of the User Manual - see what you think 🙂

Grid square isn't actually sent out over the air, just sent as part of the UDP packet to the PSK Reporter server(s). I removed that part of the update but otherwise good. 👍

> 5. Record and Play features work well - I like the pop up at the end of recording 👍

Yeah, it looked like it was supposed to show the popup all along but the bug previously mentioned was preventing it from happening. Glad that's working!

> 6. Wow you've done a lot of work 🥇 Thanks 😀

No worries!

src/dlg_audiooptions.cpp
tmiw: I figured explicitly mentioning where audio flows from towards would help clarify that a bit. Maybe longer term there can be a "simple" mode where you can select which sound card the radio is plugged into/which one has your headset and FreeDV can figure it out for you? Something like that's outside the scope of this PR though.

src/dlg_filter.cpp
tmiw: Yeah, that's probably the 700E stuff. I don't remember making that change. 👍

src/fdmdv2_main.cpp
tmiw: Taking another look, it looks like the logic is actually slightly off, so I'm going to clarify it in the lines below the comment (as well as update the comment).

tmiw: Actually, the code itself might be right but the comment still needs to be clarified. I'll update accordingly.

tmiw:
> @tmiw probably a bit late but another way to test this is use Full Duplex and Tx to yourself on the same machine. You can even inject channel noise from Tools-Options. I use full duplex quite a bit when developing.

Something to keep in mind for the future for sure. The requirements for PSK Reporter might mean that it'll need an actual radio, though, unless the code is updated to have it ignore that of course.

tmiw:
@drowe67, I also added "(Ignores Txt Msg)" to the PSK Reporter checkbox as another reminder to users that the Txt Msg field won't be used if it's enabled:

Screen Shot 2020-12-27 at 4 04 27 PM

drowe67:
> @drowe67, I also added "(Ignores Txt Msg)" to the PSK Reporter checkbox as another reminder to users that the Txt Msg field won't be used if it's enabled:
> 
> Screen Shot 2020-12-27 at 4 04 27 PM

Would it be possible to grey-out the TxtMsg field when PSK Reporter is checked?  That would make it very intuitive, and look a bit neater than the text in parens.

src/dlg_audiooptions.cpp
drowe67: Yep there is much that could be done here. A default Rx setup that uses the default sound device would be useful to get people started. 

About 12 months ago @ac0zj had a go at [filtering](https://github.com/drowe67/freedv-gui/pull/24) the long lists of sound devices but it was difficult to get consistent behavior across operating systems.

drowe67: Yes another suggestion for a simple GUI mode is a way to "disappear" many of the options to make it less bewildering for noobs.

drowe67:
@tmiw there was a digital voice mailing list comment that 800xa wasn't working.  It's working OK for me on Ubuntu, and I have pushed a test sample.

tmiw:
> Would it be possible to grey-out the TxtMsg field when PSK Reporter is checked? That would make it very intuitive, and look a bit neater than the text in parens.

Yep, just did it. 👍 

Also, we might as well do the same to the other text fields as well that don't make sense if their corresponding option is turned off. From what I can tell, those are the following:

* 700D Interleaver (if 700C Diversity Combine is disabled)
* SNR (Channel Noise)
* Carrier (Attn Carrier)
* Freq/Amplitude (Tone)
* UDP Port Number/Test (Enable UDP Messages)

Let me know if that's correct and I'll go ahead and make the needed updates.

drowe67:
> > Would it be possible to grey-out the TxtMsg field when PSK Reporter is checked? That would make it very intuitive, and look a bit neater than the text in parens.
> 
> Yep, just did it. +1

Oh wow that looks really nice now :1st_place_medal: 

>     * 700D Interleaver (if 700C Diversity Combine is disabled)

Actually that can be rm-ed entirely, as we no longer support the interleaver on 700D at the FreeDV API level.  Feature that didn't work out.

>     * SNR (Channel Noise)
> 
>     * Carrier (Attn Carrier)
> 
>     * Freq/Amplitude (Tone)
> 
>     * UDP Port Number/Test (Enable UDP Messages)
> 
> 
> Let me know if that's correct and I'll go ahead and make the needed updates.

Go for it :+1:

tmiw:
@drowe67, added code to gray out the previously discussed options. I also fixed a couple more bugs that I ended up finding during testing (one of which was a segfault when the app's idle while checking "Attn Carrier").

tmiw:
> @tmiw there was a digital voice mailing list comment that 800xa wasn't working. It's working OK for me on Ubuntu, and I have pushed a test sample.

Can't duplicate on macOS either. I even transmitted using 800XA and watched myself on an SDR and was able to confirm that audio was being sent out over the air. Might be worth getting more clarification on his exact issue.

drowe67:
@timw that looks great. If you're on a roll, the `PTT Config` and `Filter` menus could also benefit from the same gray-out improvement.  This is going a long way to making the various options less bewildering to new users :+1: 

However - you have done a huge amount of work on this PR so I totally understand if you want to leave it for now :smiley: .  We can always log this as further work.

tmiw:
@drowe67, I found another issue with the main GUI display. Basically, if you push Stop while transmitting, the waterfall view gets kinda wonky until you push Start and then Stop again. As a workaround, I just disabled the Stop button during transmit.

tmiw:
> @timw that looks great. If you're on a roll, the `PTT Config` and `Filter` menus could also benefit from the same gray-out improvement. This is going a long way to making the various options less bewildering to new users 👍
> 
> However - you have done a huge amount of work on this PR so I totally understand if you want to leave it for now 😃 . We can always log this as further work.

Done. I also fixed one more segfault I discovered (when exiting the app without having started a session first).

tmiw:
@drowe67, any ideas on that 800XA issue? I don't think anything I touched would have caused that to happen.

drowe67:
> Done. I also fixed one more segfault I discovered (when exiting the app without having started a session first).

Great!  Yeah that makes the PTT dialog in particular much more intuitive - thanks :slightly_smiling_face: .  

OK I haven't hit that segfault bug before (exits cleanly straight after start up on Linux).  Benefits of cross platform development :+1:

drowe67:
> @drowe67, any ideas on that 800XA issue? I don't think anything I touched would have caused that to happen.

I agree. It's probably user error or at worst an obscure corner case.  Old codec2 DLLs on his system, that sort of thing.  It's not uncommon to have single user report an error that they claim is the FreeDV software but really their side.  If it's a real problem some one will tell us how to reproduce it.

Suggest you push out some demo packages; we leave that for a few days just in case we missed something; then merge.

You've done a great job :100:

tmiw:
> > @drowe67, any ideas on that 800XA issue? I don't think anything I touched would have caused that to happen.
> 
> I agree. It's probably user error or at worst an obscure corner case. Old codec2 DLLs on his system, that sort of thing. It's not uncommon to have single user report an error that they claim is the FreeDV software but really their side. If it's a real problem some one will tell us how to reproduce it.
> 
> Suggest you push out some demo packages; we leave that for a few days just in case we missed something; then merge.
> 
> You've done a great job 💯

Thanks! I'll generate some new binaries now and email them out again.

As for the segfault, it was related to the PSK Reporter changes. I guess some platforms don't actually clear memory entirely when allocating objects? Though in fairness, that's officially not guaranteed by the C++ standards either.

tmiw:
@drowe67, looks like the changes to the Options dialog are causing the buttons on the bottom to not appear for at least one person. I'm going to investigate organizing the settings there into tabs (similar to what's done for the audio settings) later today.

tmiw:
@drowe67, I pushed a change to organize the options on the Options dialog into tabs to be more friendly to smaller displays. I'm open to different naming/organization for the tabs if that's preferred.

Also, I've heard about enough configuration issues with CI-V addresses on Icom radios that I clarified that in the user manual. Feel free to reword to make it more clear. 👍

tmiw:
BTW, I updated the OP in this PR to reflect the expanded scope. Also, here's what the Options dialog looks like now for reference:

Screen Shot 2020-12-28 at 3 41 17 PM

Screen Shot 2020-12-28 at 3 55 07 PM

Screen Shot 2020-12-28 at 3 55 09 PM

Screen Shot 2020-12-28 at 3 55 12 PM

Screen Shot 2020-12-28 at 3 55 15 PM

Screen Shot 2020-12-28 at 3 55 17 PM

drowe67:
@tmiw OK thanks that looks quite nice :+1:  I think you are right - these changes the Options page less "busy".  Layout comes up well on Ubuntu too. The Icom section of the manual read well - thanks for clarifying that.
 
How are you finding the wxWidgets coding?  I always struggle with it, especially the sizer stuff.  Dave Witten used some sort of tool for the original design, I've hacked it manually since then.

How would you feel about changing the `Interfacing` tab name to `UDP`?

tmiw:
@drowe67, I'm fine with renaming "Interfacing" to "UDP". Just pushed that change up.

As for wxWidgets, it took a bit but it's starting to make some sense in my head now. I still have to go to the online documentation pretty often, though.

tmiw:
@drowe67, so I decided to move the waterfall color options to a separate "Display" tab instead of making it a right-click thing. Seems more intuitive that way (as it matches how that setting is configured in other applications), especially as I don't recall the coloring thing ever having been documented well. "Blue Tint" seems to be the closest description for color number "2" in the code but I'm open to suggestions for color naming.

Screen Shot 2020-12-28 at 10 46 30 PM

Also, I'm going to do some 800XA tests on Windows just in case it's some Windows specific bug. Hopefully that's not the cae.

drowe67:
@tmiw I'd totally forgotten about that feature :facepalm: !  So you're right about it not being documented :grimacing: 

If you're on a roll - one long standing issue is the waterfall falls the wrong way, ie it should flow down rather than up :slightly_smiling_face:

tmiw:
@drowe67, done. Now to look into that one possible 800XA issue. 👍

tmiw:
@drowe67, so yeah, that sizing issue has now been resolved. I emailed out another build so we'll see how that goes.

tmiw:
@drowe67, one more build. I changed how the sound devices are saved/loaded from the config (via name instead of ID) and added some validation around those to try to detect first time execution/device removals. Dialog box examples (which display both on program start and when pushing the Start button):

Screen Shot 2020-12-30 at 2 19 26 AM

Screen Shot 2020-12-30 at 2 20 29 AM

I also tweaked the voice keyer a bit based on comments that the delays between PTTs weren't long enough (as well as fixed a bug resulting in the Stop button not being reenabled if one pushes the spacebar during voice keyer TX).

drowe67:
> @drowe67, one more build. I changed how the sound devices are saved/loaded from the config (via name instead of ID) and added some validation around those to try to detect first time execution/device removals. Dialog box examples (which display both on program start and when pushing the Start button):

Wow you've been busy :+1:  Would it be possible to move these most recent changes into a new PR?  Just `checkout -b` against the current branch and create a new PR.  That we can can stabilise the fine work you've done to date and get it released.  It also makes hunting down any issues much easier.

> Screen Shot 2020-12-30 at 2 19 26 AM Screen Shot 2020-12-30 at 2 20 29 AM

OK that's great - this is an area that causes a lot of confusion, so I appreciate your innovation here.  IIRC @ac0zj put some work into a similar idea (using names rather than device IDs) but we hit some issues with certain sound devices.  We should test this very carefully (e.g. feedback from several users on each OS) to make sure it's working OK.  Another good reason to move this to a new PR....

> I also tweaked the voice keyer a bit based on comments that the delays between PTTs weren't long enough (as well as fixed a bug resulting in the Stop button not being reenabled if one pushes the spacebar during voice keyer TX).

Cool - I didn't know about those issues :slightly_smiling_face: 

> BTW, Walter K5WH was saying something about how the voice keyer .wav file needed to be something different for 2020 vs. the other modes. Is there any possibility for improvement there?

Yes there was an issue there ... I can't recall exactly but probably we need a 16 kHz sample for 2020.  Perhaps ask Walter?

drowe67:
@tmiw credit where credit is due :slightly_smiling_face: https://github.com/drowe67/freedv-gui/commit/9eddbfd7406736a7ef0ca1d989129ba8bc46e127

tmiw:
@drowe67, looks like there are complications on Windows, specifically with Virtual Audio Cable (needed for SDR software on that platform). I'll back out that particular change and put it in a separate PR for further discussion on how to handle it.

drowe67:
@tmiw - yep that was exactly our experience last time!  Very tricky to debug...

tmiw:
@drowe67, moved the changes to PR #82.

drowe67:
@tmiw I resolved a small conflict from the merge of https://github.com/drowe67/freedv-gui/pull/83 - hope it's OK.

Travis is still off line - I have applied to them for a monthly allowance of credits.  However just let me know when you feel this PR is ready for merge.

tmiw:
@drowe67, I haven't heard anything from Siegfried (or anyone else) since I reverted the PortAudio changes, so we're probably good to merge this up to master later today or tomorrow.

tmiw:
@drowe67, oops, spoke too soon. Just checked in another change to reverse the Y axis labels on the waterfall view.

drowe67:
No problem @tmiw - Peter R who reported the axis flip is actually a neighbor of mine, lives a few km away.  If that's all people are finding you're doing great :100:

tmiw:
@drowe67, I tried updating fdmdv2_plot_scatter.cpp as follows:

```
        for(i = 0; i < scatterMemSyms; i++) {
            x = x_scale * m_mem[i].real;
            y = y_scale * m_mem[i].imag;
```

It looks like when clipping = off, the above behaves fine (AFAIK) but clipping = on causes the points to be centered in the upper left quadrant. Perhaps we should include the additional (m_rGrid.GetWidth()/2) inside the _scale variables only in certain instances?

tmiw:
@drowe67, I researched scatter plots some more and based on my understanding, I think max_xy is supposed to be based on the magnitude of the complex number, not the real and imaginary components individually. I checked in a change that looks like it resolves the display issue but please take a look and confirm.

drowe67:
> @drowe67, I researched scatter plots some more and based on my understanding, I think max_xy is supposed to be based on the magnitude of the complex number, not the real and imaginary components individually. I checked in a change that looks like it resolves the display issue but please take a look and confirm.

Hmm, I wonder if Peter encountered this problem as he had a rectangular scatter window with the width << height?  I guess the goal is here is to scale such that we don't hit the smallest dimension.  Maybe also include some asserts to fire if any coordinates land outside the window.

I'll think about this one some more ....

tmiw:
@drowe67, I think the default sizing of the main FreeDV window has height > width anyway. Still, it's weird that it only happens when clipping is off.

drowe67:
@tmiw can you pls move the commits for the scatter diagram clipping issue to a new "maintenance" PR?  From now on pls put any new commits (anything not directly related the PSK reporting) in that new PR.

I've just started back at $dayjob so it might be the weekend before I can put some time into the scatter diagram issue.  I still haven't had a response from Travis so we'll need to proceed with manual tests for now.

Are you happy with the PSK reporter functionality?  If so:
1. We can merge this PR.
1. We can then discuss merging https://github.com/drowe67/freedv-gui/pull/82

tmiw:
@drowe67, I think I'm good on PSK Reporter for now. We can revisit as more people use it (e.g. perhaps to see if a different encoding for the text field would work more reliably). 

Anyway, I created a new branch at https://github.com/tmiw/freedv-gui/commits/bugfixes, which should have this PR minus the scatter plot stuff. Hopefully it doesn't get overwritten next time I try to merge from master but maybe if I hold off until this and #82 are merged it won't be a huge deal (and the change I did wasn't big anyway, so it could be easily redone if it came to that).

drowemyriota:
Thanks @tmiw - I just pulled the latest version and it builds and runs OK against codec2/master so I'll merge now.  Thank you so much for all the work you put into this PR.  You've done a fine job, much appreciated!

drowe67:
> Anyway, I created a new branch at https://github.com/tmiw/freedv-gui/commits/bugfixes, which should have this PR minus the scatter plot stuff. Hopefully it doesn't get overwritten next time I try to merge from master but maybe if I hold off until this and #82 are merged it won't be a huge deal (and the change I did wasn't big anyway, so it could be easily redone if it came to that).

Cool.  Could you pls raise a PR for that branch?

tmiw:
@drowe67, the scatter plot stuff is now in #85.

PR #85, opened 2021-01-06
Resync + scatter plot fixes

tmiw:
@drowe67, as the title suggests, this is for the resync and scatter plot fixes.

1. The changes in https://github.com/drowe67/codec2/pull/167 require testing all FreeDV modes with a noise source to check squelch is working.  For each mode I tried the associated canned demo files in freedv-gui/wav, and added noise using Tools-Options-Simulate to about 2dB.  Tested if squelch worked with a signal present, and also if audio was squelched when the canned demo file had finished.

   | Mode | Tested By | Pass |
   |---|---|---|
   | 700C | DR | :heavy_check_mark: |
   | 700D | DR | :heavy_check_mark: |
   | 700E | DR | :heavy_check_mark: |
   | 800XA | DR | :heavy_check_mark: |
   | 2020 | DR | :heavy_check_mark: |
   | 2400B | DR | :heavy_check_mark: |

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -651,6 +651,11 @@ On Linux, using the Alsa loopback module:
                           - Transmit Tab - To Radio select   -> Loopback: Loopback PCM (hw:1,1)
 ```
 
+## Tips
+
+1. The space-bar can be used to toggle PTT.
+1. You can left click on the main window to adjust tuning, the vertical red line on the frequency scale will show the current centre frequency.  FreeDV will automatically track any drift once it syncs.
+
 ## Glossary
 
 Term | Notes
--- a/src/dlg_options.cpp
+++ b/src/dlg_options.cpp
@@ -44,7 +44,8 @@ extern int                 g_freedv_verbose;
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
 OptionsDlg::OptionsDlg(wxWindow* parent, wxWindowID id, const wxString& title, const wxPoint& pos, const wxSize& size, long style) : wxDialog(parent, id, title, pos, size, style)
 {
-
+    sessionActive_ = false;
+    
     this->SetSizeHints(wxDefaultSize, wxDefaultSize);
 
     wxBoxSizer* bSizer30;
@@ -941,18 +942,30 @@ void OptionsDlg::OnUDPTest(wxCommandEvent& event)
 
 void OptionsDlg::updatePSKReporterState()
 {
-    if (m_ckbox_psk_enable->GetValue())
+    if (!sessionActive_)
     {
-        m_txtCtrlCallSign->Enable(false);
-        m_txt_callsign->Enable(true);
-        m_txt_grid_square->Enable(true);
+        m_ckbox_psk_enable->Enable(true);
+        if (m_ckbox_psk_enable->GetValue())
+        {
+            m_txtCtrlCallSign->Enable(false);
+            m_txt_callsign->Enable(true);
+            m_txt_grid_square->Enable(true);
+        }
+        else
+        {
+            m_txtCtrlCallSign->Enable(true);
+            m_txt_callsign->Enable(false);
+            m_txt_grid_square->Enable(false);
+        }    
     }
     else
     {
-        m_txtCtrlCallSign->Enable(true);
+        // Txt Msg/PSK Reporter options cannot be modified during a session.
+        m_ckbox_psk_enable->Enable(false);
+        m_txtCtrlCallSign->Enable(false);
         m_txt_callsign->Enable(false);
         m_txt_grid_square->Enable(false);
-    }    
+    }
 }
 
 void OptionsDlg::updateChannelNoiseState()
--- a/src/dlg_options.h
+++ b/src/dlg_options.h
@@ -59,7 +59,13 @@ class OptionsDlg : public wxDialog
             }
         }
 
-
+        void setSessionActive(bool active) 
+        {
+            sessionActive_ = active; 
+        
+            updatePSKReporterState();
+        }
+        
     protected:
 
         // Handlers for events.
@@ -174,6 +180,8 @@ class OptionsDlg : public wxDialog
          void updateAttnCarrierState();
          void updateToneState();
          void updateUDPState();
+         
+         bool sessionActive_;
 };
 
 #endif // __OPTIONS_DIALOG__
--- a/src/fdmdv2_main.cpp
+++ b/src/fdmdv2_main.cpp
@@ -380,7 +380,7 @@ MainFrame::MainFrame(wxString plugInName, wxWindow *parent) : TopFrame(plugInNam
     {
         // Add Waterfall Plot window
         m_panelWaterfall = new PlotWaterfall((wxFrame*) m_auiNbookCtrl, false, 0);
-        m_panelWaterfall->SetToolTip(_("Left click to tune, Right click to toggle mono/colour"));
+        m_panelWaterfall->SetToolTip(_("Left click to tune"));
         m_auiNbookCtrl->AddPage(m_panelWaterfall, _("Waterfall"), true, wxNullBitmap);
     }
     if(wxGetApp().m_show_spect)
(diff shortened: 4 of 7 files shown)


tmiw:
Hang on, I'm only seeing the resync stuff.

drowe67:
@tmiw re the scatter plot issues - did you manage to reproduce them before introducing your fix?

drowe67:
@tmiw I agree with your resync fix, however I'm still getting the resync issue with internal noise.  When I turn on Options - API Verbose I can see `nans` getting into the modem:
```
207 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
OnReSync
Unsyncing per user request.
    ct_est:    0 foff_est:  2.0 timing_valid: 1 timing_mx: 0.5487
208 nin:    0 st: search euw:  0  0 mf:  0 f:   2.0 pbw: 0 snr:  4.8 eraw:   0 ecdd:   0 iter:   0 pcc:   0 rxst: ----
209 nin: 1280 st: trial  euw:  9  1 mf:  0 f:   1.7 pbw: 0 snr:  4.0 eraw:   0 ecdd:   0 iter: 100 pcc:  52 rxst: EBST
210 nin: 1280 st: trial  euw:  0  1 mf:  0 f:   1.5 pbw: 0 snr: -nan eraw:   0 ecdd:   0 iter:   1 pcc:   0 rxst: EBST
211 nin: 1280 st: trial  euw:  0  1 mf:  0 f:  -nan pbw: 0 snr: -nan eraw:   0 ecdd:   0 iter:   1 pcc:   0 rxst: EBST
212 nin: 1240 st: trial  euw:  0  1 mf:  0 f:  -nan pbw: 0 snr: -nan eraw:   0 ecdd:   0 iter:   1 pcc:   0 rxst: EBST
```

However, when I generate an external file with noise:
```
~/codec2/build_linux master $ ./src/freedv_tx 700D ../raw/ve9qrp.raw - | ./src/cohpsk_ch - - -20 --Fs 8000 | sox -t .s16 -r 8000 -c 1 - ~/Desktop/ve9qrp_700d_snr.wav
cohpsk_ch: SNR3k(dB):     2.20  C/No....:    36.97

```
... and use that as the file input for FreeDV, then no problem.  This might explain why no one has experienced any issues with real world signals.

tmiw:
@drowe67, yep, I was able to reproduce the scatter plot display issue prior to the fix.

As for nin=0, should the code perhaps be updated to not provide nin=0? Although simply checking for it as per the commit you pushed might be the easiest way to go.

drowe67:
@tmiw that's in interesting point. Looking at the relevant line in [ofdm.c](https://github.com/drowe67/codec2/blob/master/src/ofdm.c#L1113) it looks like nin=0 is valid - it's saying that we have identified a likely start of an OFDM modem frame at position 0 in the input buffer, so we have enough samples thanks, and no more are required on this call. It then jumps immediately back into sync.

I guess this make sense - although we have manually told the state machine to lose sync, but there are perfectly valid modem frames sitting in the buffer.

@srsampson do I have that right?

drowe67:
@tmiw I have copleted some squelch testsing (see table at top of PR).  There was another issue with 2400B that I fixed.  Some general comments:

1. The squelch works OK when squelching "no signal".  However when a signal is present, but has a low SNR, the squelch is not reliable, you occasionally get break through of audio as the SNR bounces around.  That might be an issue for future work.  I think the main use case for squelch is the former.

1. On the scatter and in particular in eye diagram mode, I occasionally see points outside of the scatter panel.  Wondering if we should check the coords of each point before allowing it to be plotted.

1. The Clock Offset `ClkOff` field occasionally displays too many digits that run off the panel.

tmiw:
@drowe67, I haven't tried the eye diagram stuff yet so didn't realize that was a separate thing. I just looked now, though, and it seems that we calculate min_y but don't use it. Perhaps that could be used?

In any case, adding bounds and forcing points to min/max x and y when outside of them is straightforward but not sure if that'd just mask another calculation issue.

drowe67:
> In any case, adding bounds and forcing points to min/max x and y when outside of them is straightforward but not sure if that'd just mask another calculation issue.

Thinking about the requirements of the scatter plot system, it's not unusual to have some outliers e.g. the scatter diagram blobs on a noisy signal will have a normal distribution.  We'd like the "AGC" to place the main blob inside the plot window, but don't mind if some of those outliers fall outside the plot window.  However is a few points do fall outside the window we would prefer them not to be displayed.  I guess the AGC system went a bit awry with some signals, which lead to the bug that kicked off this PR.

drowe67:
> Also, it looks like ClkOff is set to be 5 digits per fdmdv2_main.cpp:

Unfortunately for `%d` it's a minimum field width.  I found a modulus trick on Stack Overflow that's tamed it a little.  However ClkOff and FrqOff occasionally hit the RHS edge or stray just outside it.  I've taken a look at the `topFrame.cpp` sizer logic but can't figure out how to widen it.  Would you happen to know how to make that box a little wider?

tmiw:
@drowe67, for me at least, ClkOff seems to provide sufficient space:

Screen Shot 2021-01-07 at 6 07 55 PM

However, calling SetMinSize() on that control might help for other platforms. I'll check my Ubuntu and Windows VMs and play around a bit. (BTW, wiping all of your settings and returning to defaults may also help. I noticed that wxWidgets gave significantly more space when I wiped my settings while testing the code in the audio-ux branch.)

tmiw:
@drowe67, I updated the bounding for the scatter plot to hide points out of bounds. I'm not 100% sure about the eye chart but scatter seems to work, FWIW.

drowe67:
@tmiw - interesting that the layouts are different on OSX!  On Ubuntu it seems to be the Reset & Resync button widths that set the width of that whole LH panel.  

Thanks - the LH panel is wider.  Good work :+1:

Something funny going with the eye diagrams on 800XA and 2400B, they seem to keep scaling down to a single line, then slowly work their way up.  I'm not sure if this was happening before.

BTW after a week and numerous requests Travis support haven't replied to me - so I am working up GitHub Actions CI for codec 2 https://github.com/drowe67/codec2/pull/168 & freedv-gui https://github.com/drowe67/freedv-gui/pull/86

tmiw:
@drowe67, that reminds me, the Debian/Ubuntu package stuff in the tree should be updated at some point to fix whatever's preventing it from generating packages. I tried to do it pretty early on in the PSK Reporter work but I kinda left it to work on other stuff.

drowe67:
> Something funny going with the eye diagrams on 800XA and 2400B, they seem to keep scaling down to a single line, then slowly work their way up. I'm not sure if this was happening before.

Hmm, this is a big tricky to reproduce.  In 800XA mode it seems to happen when I move from my laptop mic as the Rx sound device to playing a stored 800XA file and back again.  It might not be worth persuing at this stage.

drowe67:
> @drowe67, that reminds me, the Debian/Ubuntu package stuff in the tree should be updated at some point to fix whatever's preventing it from generating packages. I tried to do it pretty early on in the PSK Reporter work but I kinda left it to work on other stuff.
>

Actually I've never been able to make that work.  Gosh, I can't remember last time I looked at it.  There is a gentleman (Maitland Bottoms AA4HS) who prepares FreeDV Debian packages but I don't have his email handy.

I guess my fall back until now has been the assumption that most Linux users should be able to run `./build_linux.sh`.  But some packaging would be nice or sure.

> It would also be nice to be able to auto-generate at least Linux and Windows packages with GitHub Actions so we have test packages every push, but that might be farther down the road.

Every push might be a bit often - as I think there is a limit of CPU minutes/month for GitHub actions.  I just ran into that problem with Travis and it's caused us a CI outage.  Also I am a bit nervous about Windows users running untested code.  Building packages on a merge might be a good point.

tmiw:
@drowe67, I haven't looked into GitHub Actions much but if it's possible to have it fire off stuff only on merges, that'd be good too. 👍

drowe67:
@tmiw so is this PR good to go?  I'm happy with it.

tmiw:
@drowe67, yep, I'm fine with it for now. We can look into the eye diagram weirdness later.

PR #88, opened 2021-01-08
Github action to render manual as PDF

drowe67:
Currently renders on every push.  We could make this just on "push to master". It's pretty heavy on CPU minutes due to all the Tex packages it needs.

ToDo:

1. Only run if `USER_MANUAL.md` has been changed :heavy_check_mark:   
1. Clean up PDF :heavy_check_mark:   
1. Section heading numbering :heavy_check_mark:   
1. Render to HTML as well, with section numbering :heavy_check_mark:   
2. Some combination of Git hash/date in PDF title  :heavy_check_mark:  
4. Link HTML and PDF versions from http://freedv.org


drowe67:
@hobbes1069 @tmiw @jimt :point_up:

tmiw:
@drowe67, nice! I just noticed there wasn't a mention of 1.5.1 in the changelog at the bottom, though, but I don't think it's been officially released outside of nightly builds I've emailed to the list either.

drowe67:
@tmiw should the 1.5.1 mention in the changelog have been merged with https://github.com/drowe67/freedv-gui/pull/77?

USER_MANUAL.md
```diff
@@ -90,9 +83,9 @@ your radio for transmission over the air.
 Tab | Sound Device | Notes
 --- | --- | ---
 Receive Tab | Input To Computer From Radio | The off air FreeDV signal **from** your radio rig interface to your computer
-Receive Tab | Output From Computer To Speaker/Headphones | The decoded audio from your computer to your Speaker/headphones
+Receive Tab | Output From Computer To Speaker | The decoded audio from your computer to your Speaker
```
tmiw, line 86: @drowe67, does the FreeDV app itself need to be updated to remove "Headphones" from the Audio Config dialog?

tmiw:
> @tmiw should the 1.5.1 mention in the changelog have been merged with #77?

Most likely, but it can be updated now (or a separate PR if the FreeDV application needs to be updated anyway per my other comment).

USER_MANUAL.md
drowe67: Hmmm.  This was actually an annoying hack I had to do.  When the PDF renders from markdown it can't break up table rows over multiple lines, it just truncates them.  Actually I think `Speaker/headphones` is best, we should stick with that.  I just need to reword to fit on one line or learn how to drive pandoc better :thinking:

tmiw:
@drowe67, it also looks like the page numbers on the bottom of the PDF are cut off (seen both on the GitHub PDF renderer and in Preview on macOS):

Screen Shot 2021-01-10 at 11 25 48 PM

Meanwhile, I'll go ahead and open a PR to update the manual's changelog to mention 1.5.1.

USER_MANUAL.md
drowe67: OK I have figured out how to do multi-line tables

tmiw:
@drowe67, looks good to me. 👍

PR #92, opened 2021-01-11
Fix manual typos

jimt:
use 'space bar' consistently
capitalization of RIGblaster and Hamlib


drowe67:
Thanks @jimt  - nice to see the automated rendering kicking off too!

jimt:
Yes, was cool to see that happen.

Issue #16, opened 2019-07-05
PulseAudio Support


drowe67:
A few people have requested pulse audio support on Linux.

drowe67:
Basic operation tested by @tmiw on 6 Feb 2021, posted to digitalvoice Goggle group:

> Here's another screenshot with multiple applications using PulseAudio simultaneously (WSJT-X and FreeDV grabbing 
> receive audio from SparkSDR). Again, it seemed to work properly in brief testing:

<img width="2048" alt="Screen Shot 2021-02-05 at 12 52 13 AM" src="https://user-images.githubusercontent.com/45574645/108639027-44c31a80-74e2-11eb-8013-e0efdf48a14a.png">

@timew asked for feedback and specific bug reports but there has been no response (up to 22 Feb).

mike-hb:
as stated before in the list

I use opensuse thumbleweed 64bit an the newest git pull from today (2021-03-26T14:52 Z)

with compile by build_linux.sh, I got an error
freedv-gui/src/hamlib.cpp: In Elementfunktion »bool Hamlib::connect(unsigned int, const char*, int, int)«:
freedv-gui/src/hamlib.cpp:140:57: Fehler: »FILPATHLEN« wurde in diesem Gültigkeitsbereich nicht definiert
  140 |     strncpy(m_rig->state.rigport.pathname, serial_port, FILPATHLEN - 1);
      |                                                         ^~~~~~~~~~
make[2]: *** [src/CMakeFiles/freedv.dir/build.make:225: src/CMakeFiles/freedv.dir/hamlib.cpp.o] Fehler 1
make[2]: Verzeichnis „freedv-gui/build_linux“ wird verlassen
make[1]: *** [CMakeFiles/Makefile2:134: src/CMakeFiles/freedv.dir/all] Fehler 2
make[1]: Verzeichnis „freedv-gui/build_linux“ wird verlassen
make: *** [Makefile:171: all] Fehler 2

After defining the missing FILPATHLEN in the head of that file hamlib.cpp setting it to 80, git commit and again "build_linux.sh" -- it was compiled.

So I tried to use it.

After starting pavucontrol and executing build_linux/src/freedv I was asked to configure audio.

the offered audio devices are a very limited list and the needed USB Soundcard for the microphone as well as the USB soundcard of the radio (mcHF) was _not_listed_ in the audio dialogs

Than I tried both soundcards to use "pulse" -- this was banned  after "apply"

Any help needed.
vy 73 
[...]

PR #103, opened 2021-02-05
Fix single device sound card configuration

tmiw:
@drowe67, I installed pavucontrol in my Ubuntu VM tonight and ended up accidentally running into an issue with how single sound card configs are saved in 1.5.2. This PR is to fix that at least. 

(BTW, PulseAudio in a single sound card configuration seems to work in brief testing, but I also wasn't trying to run other apps that use it along with FreeDV.)

EDIT: typo of "pavucontrol".


drowe67:
> (BTW, PulseAudio in a single sound card configuration seems to work in brief testing, but I also wasn't trying to run other apps that use it along with FreeDV.)

Yes I've had some success using the Pulse playback device, when some other device has grabbed the default device.

tmiw:
@drowe67, yeah, I'd possibly try to flesh out more info so that we have a defined scope of work for #16 (if there is in fact anything needed).

Anyway, returning to this PR, I emailed the build I sent to the digitalvoice tonight out to Walter and a few others for testing. I think someone else on Windows ran into this issue too but need to confirm for sure.

drowe67:
@tmiw is this one ready to merge?

tmiw:
@drowe67, yep, I haven't heard of any issues so we're good here.

PR #105, opened 2021-02-14
Fix radio mode GUI color when using macOS dark mode.

tmiw:
@drowe67, I started running dark mode full time on my MacBook Pro and noticed that the "USB" label on the bottom left was hard to read when I was using the FreeDV application. This fix resolves that issue on dark mode (and I've also confirmed that it displays properly in light mode as well).


drowe67:
@tmiw thanks for that fix - looks like we have a minor conflict in the user manual (I made some changes earlier today).

tmiw:
@drowe67, hmm, I tried to resolve the conflict here but that didn't seem to work. It might have to be resolved after approval.

drowe67:
@tmiw Unfortunately I can't merge unless the conflict is fixed (merge button is greyed out).  Interesting problem - that PDF is machine generated so shouldn't be involved in the merge process.  Also as you didn't modify the manual, we don't need or want the generation logic to kick in.

tmiw:
@drowe67, I'm going to try to force it to regenerate the user manual to see if that helps. Otherwise, it sounds like from googling that it should be possible to resolve the conflict using the command line or some sort of GUI client.

tmiw:
@drowe67, looks like we're good now, assuming it doesn't regenerate the PDF again.

drowe67:
Thanks @tmiw - I suspect I need to tweak the GitHub action to avoid this sort of issue in future.  However it is kind of cool how it auto-renders PDFs.

PR #108, opened 2021-03-23
Add required Ubuntu packages to README for building.

tmiw:
Updating README per JH0PCF's email to digitalvoice:

> FreeDV source code on Ubuntu 20.04
When building, the procedure in the manual
I couldn't handle it well
By adding module-assistant build-essntial
You can now process safely

\+ @drowe67

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -24,7 +24,7 @@ jobs:
       shell: bash
       run: |
            sudo apt-get update
-           sudo apt-get install libspeexdsp-dev libsamplerate0-dev sox git libwxgtk3.0-dev portaudio19-dev libhamlib-dev libasound2-dev libao-dev libgsm1-dev libsndfile-dev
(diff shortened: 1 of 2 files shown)


tmiw:
@drowe67, hmm, looks like the GitHub tasks have some issues at present. From [here](https://github.com/tmiw/freedv-gui/actions/runs/678403822):

> Package libwxgtk3.0-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

> E: Package 'libwxgtk3.0-dev' has no installation candidate
Error: Process completed with exit code 100.

Maybe it needs to be installing libwxgtk3.0-gtk3-dev instead?

tmiw:
OK, looks like I was able to fix the GitHub action on my end. Build works okay now with libwxgtk3.0-gtk3-dev instead of libwxgtk3.0-dev.

drowe67:
@tmiw OK that's interesting, I wonder how the GitHub action ran OK previously :thinking: ?

tmiw:
@drowe67, not sure. I'm guessing the package it was using before is a virtual package that only appears once you install using the actual package name. Oh well, as long as it works?

PR #109, opened 2021-03-25
Add support for dynamic HF mode switching 

tmiw:
This PR adds support for the following:

* Simultaneous decode of 2020, 1600 and 700C/D/E (without needing to push Stop first, change the mode and push Start again).
* Dynamic switching of the current TX mode between the aforementioned modes, again without needing to restart the session.
* A TX level slider on the bottom of the main screen to fine-tune transmit output (to more easily avoid clipping ALC and conflicting with other soundcard ham radio applications).

For reference, the main screen with the TX level slider is as follows:

Screen Shot 2021-03-25 at 1 12 23 AM

and the new option inside Tools->Options to enable dynamic mode switching:

Screen Shot 2021-03-25 at 1 12 13 AM

Note that 800XA and 2400B aren't included in the dynamic switching logic due to significant additional CPU usage (in the case of 2400B at least), but could possibly be added in the future.


tmiw:
@drowe67, I did some testing with Walter K5WH this afternoon and everything seemed to check out, so I'm going to send off builds to him and his group to play with some more before posting them to the list.

drowe67:
Woo you're on fire @tmiw :grinning: - great work.  I'll take a closer look at the code and try out your PR here over the next few days too.

1. An idea about the Tx level slider - perhaps it could replace the vertical Rx level bar on the lower LHS?  I'm not sure that's used by anyone.  Although I guess we should check that assumption.  The Rx level bar could maybe be replaced by a LED type indicator that blinks red when the Rx audio clips.  It's really not that critical unless the input audio is heavily clipped.  Hmm, I wonder what other applications do for Rx level indication?
2. Another idea - is it possible to have that lower LHS area be a slider during Tx and a Level indicator during Rx?  Not sure if that's possible in wxWidgets.
3. Is the value of the TX slider stored in the persistent config and stays the same if we restart freedv-gui?

drowe67:
Re 2400B not working with this PR  that's a bit of a mystery - it's actually a really simple modem so it' unlikely to be CPU. Perhaps a bug in the FreeDV API that makes it high CPU  However AFAIK 2400B is very rarely used so not a problem.

tmiw:
@drowe67, I thought about that, actually. It was quickest to do a separate thing but I imagine it might not be that big of a deal to have it swap out the RX one during TX. Maybe we'll see what others think of it as it is now first?

And yes, the value of the TX level slider is saved in persistent config. 👍

drowe67:
> @drowe67, I thought about that, actually. It was quickest to do a separate thing but I imagine it might not be that big of a deal to have it swap out the RX one during TX. Maybe we'll see what others think of it as it is now first?

Yep - good idea, lets get feedback from end users.  Another possible location in on the upper RHS next to squelch.

> And yes, the value of the TX level slider is saved in persistent config. +1

:+1:

tmiw:
This first round of updates basically involved the TX attenuation slider. Since it was linear before, it didn't behave as expected. The current behavior should now be more in line with what other ham radio apps use (and is delimitated in dB to make it clear).

Unfortunately, conditions on 20 meters weren't great today, so we couldn't test further. Hopefully tomorrow will be good enough to actually try having QSOs on the air.

tmiw:
@drowe67, so far so good this morning; I was able to do a QSO with WA5QPZ (who I think wasn't running the test build) and was also able to decode other people using 700E while mine was set to 700D. I might be able to send out the build for wider use today or tomorrow. 👍

tmiw:
@drowe67, I sent out the test build to digitalvoice now so we'll see how it goes. 👍

src/main.cpp
```diff
@@ -2661,19 +2661,23 @@ void txRxProcessing()
...
 
             if (g_analog)
                 nout = resample(cbData->outsrc1, outsound_card, outfreedv, g_soundCard1SampleRate, freedvInterface.getTxSpeechSampleRate(), 10*N48, nfreedv);
             else
                 nout = resample(cbData->outsrc1, outsound_card, outfreedv, g_soundCard1SampleRate, freedv_samplerate, 10*N48, nfreedv);
+            
+            // Attenuate signal prior to output
+            double dbLoss = g_txLevel / 10; // (-10) + (0 - (-10)) * ((double)g_txLevel)/100.0;
+            double scaleFactor = exp(dbLoss/20.0 * log(10.0));
+
```
drowe67, line 2676: I fprintf-ed `scaleFactor` and it check out dB value == linear :+1:

tmiw: The line you commented on didn't look exactly right so I had to make sure. Looks like 10.0 instead of 10 is used as expected (to avoid casting issues).

drowe67:
Hi @tmiw I tried this PR with your test file.  Works well! It was interesting to see the different signals roll past.

1. I did hit this assert just once, but couldn't reproduce it:
```/home/david/tmp/freedv-gui/codec2/src/fdmdv.c:1770: fdmdv_8_to_16: Assertion `(n % FDMDV_OS) == 0```

1. I had a few minor warnings when built on my Ubuntu 18 laptop, have fixed them.

1. I'd like to move that Tx level slider before merging, it seems to dominate the main UI at present.  It could be smaller and perhaps vertical (maybe there are UI standards for this sort of thing?).  I wonder if it should be on the main panel of another dialog?  Not sure on that myself - it might be the sort of thing that is often adjusted, like squelch, so the main panel might be the best place. As you can see from all the question marks I'm unsure exactly what to do :slightly_smiling_face: I would be interested to hear what the end users think :thinking:

DG9BFC:
2400b shows snr 16 with no input ... and 800xa shows around 5 to 8 ... should not both be at -2 with no signal??

tmiw:
@DG9BFC, I typically go on HF from home with a Hermes Lite 2 SDR and piHPSDR. With absolutely no output going into the virtual sound card (as in, piHPSDR not running at all), 800XA gives me -11 dB SNR and 2400B 0 dB with the latest test build. There weren't any recent changes to Codec2 that would have caused issues (and this PR doesn't seem like it would have impacted anything related to decode on those two) but @drowe67 may be better to ask.

As for the GUI in general, do you guys think we could rely on the coloring of the waterfall to determine proper receive levels? That way, we could just replace the vertical RX level indicator on the lower left with the new slider.

drowe67:
IIRC I haven't changed anything that would affect reported SNRs.  Lets keep an eye on that one, there are always corner cases that creep in, especially across operating systems. @DG9BFC:
1. What operating system were you using?
1. Do you actually use those modes on air?  I wasn't sure if anyone was using them.

@tmiw re levels I think the waterfall has an AGC on it, so that might not be a super reliable indication of Rx level.  Hmm, in the manual I noticed that we don't actually have much on how to set the Rx level:

```On receive, FreeDV is not very sensitive to the From Radio level, adjust so it is mid-range and not clipping. FreeDV uses phase shift keying (PSK) so is not sensitive to amplitude.```

I actually use the `From Radio` tab to monitor level, and make sure it's not clipping.  I can't recall the last time I looked at the Rx level side bar.  Some ideas:

1. Add some text to the manual about setting Rx level using the `Fom Radio` tab (the manual text feels a bit light for a noobie atm).
1. Keep the "Too High" alert that pops up over that Rx level bar just in case.  We'd have to think of somewhere to put it.
1. Ask people on the mailing list of they use that Rx level bar, and how they set Rx levels in practice.

DG9BFC:
operating system is win10 on an atom tabler (4x 1.92ghz) with 4 gig ram
using sdr console over qo100 and two virtual cables (works fine in fullduplex)
i tested all modes ... and here the 800xa works best (shows fatsest lock and good audio)
snr hovering between 15 and 17!! so a superb decode
level show is roughly three quarter of full scale .. maybe a bit more (just below the too high indicator showing up)
about screenlayout i will drop you a direct mail (rx level, tx level etc.)

tmiw:
@drowe67, I tried moving the slider to the right and it's actually not a bad placement:

Screen Shot 2021-03-28 at 11 52 02 PM

Let me know if there should be further improvements. This should hopefully at least let us kick the can down the road re: RX level improvements.

(FWIW, I do remember the colors on the waterfall actually changing in response to turning the AF dial on my KX3. I typically aim for something in the middle of the scale on the lower left for RX, though I'm not sure if that's the most optimal. VARA FM, for instance, is the best when the radio/soundcard interface is set such that the level is ~-13dB with no signal and squelch fully open on the radio.)

drowe67:
@tmiw yes that looks quite nice - especially the black you get on OSX :slightly_smiling_face: Does it add to the min height?  Just wondering if some users will be run out of screen vertically.  Looks OK on my laptop 1920x1200 laptop screen.

VARA_FM!  We'll have to get you using some open source data modes :wink: I'm making good progress lately on HF (and late last year) VHF data modes.

drowe67:
@tmiw all good for me, and it sounds like the beta testers aren't hitting any problems.  Just let me know when you're ready to merge.

tmiw:
@drowe67, no additional minimum height required. And I'm good for merging. 👍

DG9BFC:
any chance to get a test version for me?? ... i use only an old core2duo laptop with low screen resolution for my hf station or the atom tablet for qo100 (also a bit lower resolution used cause i connect that to a tv set via hdmi as big screen)
sorry no way to test 2020 mode (no avx) but the 700 modes and 800 xa run fine 
15 to 18 db snr is a perfect decode if you count in that its via that transponder and receiving is done with a "slightly drifting" lnb

tmiw:
@DG9BFC, see https://www.dropbox.com/sh/1gh766ixcguviwp/AADhiL5-IK0odbn08WynMUqwa?dl=0. I only generated Windows builds this time as there weren't any changes vs. the last build other than the placement of the slider.

DG9BFC:
thank you for that update.. works fine
but uses max height of my screen resolution (have to pull to minimum height that it fits
usable .. yes .. but if you add motre things please no additional height
greetz sigi
ps put that in the officila release position that other users may find it

drowe67:
@tmiw I've just pushed https://github.com/drowe67/freedv-gui/pull/109/commits/561af7fd0b2c8a9775fbbbdf8b251d52e0493507 - just let me know if you're happy with that and I'll merge.

Should we release 1.5.3 on freedv.org, or perhaps wait a little while to collect any other changes?

tmiw:
@drowe67, other than the comment on the user manual changes (TX level isn't on the bottom anymore), it looks good to me. The upcoming Pulse changes will likely affect a smaller set of people so we can take our time with those, so I'd say we can release after merging this PR.

drowe67:
@tmiw - Cool :sunglasses: - I'll pop 1.5.3 up on freedv.org this weekend.  When convenient can you pls email me a .dmg?

tmiw:
@drowe67, I uploaded it to [here](https://www.dropbox.com/sh/l8k8a0az5hc3qsj/AADXC_KskkkdbRXD_58sZffia?dl=0) since 30MB files are probably not a good idea to email :) Feel free to re-upload to the website.

PR #110, opened 2021-03-27
Add correct libraries for building static PortAudio on Linux.

tmiw:
@drowe67, here's a quick fix to CMake for something I discovered while investigating the PulseAudio stuff. (I was hoping that the latest PortAudio snapshot would work a bit better than the behavior I saw tonight, but it doesn't seem to be the case thus far.) Basically, this makes it so that the PortAudio static library enabled by -DUSE_STATIC_PORTAUDIO=1 is actually linked into the binary correctly along with ALSA. 

Anyway, I'll update #16 with what I found so far.

### DIFF
--- a/cmake/BuildPortaudio.cmake
+++ b/cmake/BuildPortaudio.cmake
@@ -54,10 +54,9 @@ if(WIN32)
     list(APPEND PORTAUDIO_LIBRARIES ${WINMM} ${DSOUND}
 )
 elseif(NOT APPLE)
-#else(WIN32)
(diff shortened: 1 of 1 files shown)


drowe67:
@tmiw would you like me to merge this one or leave it open as a PR for other Pulse experiments?

tmiw:
@drowe67, go ahead and merge. Pulse will likely go into a separate branch on my end anyway to avoid cluttering up e.g. the multi-decode one.

PR #98, opened 2021-01-14
Fix display bug when clearing null character + improve sync time.

tmiw:
As part of #97, this fixes a display bug I noticed while testing PSK Reporter functionality (and may also fix the occasional segfault others have noticed too based on the ones I've gotten). I also made some tweaks to improve sync time for the text field when PSK Reporter functionality is turned on.

PR #111, opened 2021-03-31
Sync status reporting improvements

tmiw:
This PR has some improvements to sync/mode reporting based on comments by Peter Reichelt VK5APR. It should now stay solid green when in sync (and return to red when sync drops out) in both auto and manual RX modes as long as there is typical noise on the bands.


tmiw:
@drowe67, I've seen it with 700D and E from what I remember. It might happen with others too.

drowe67:
> @drowe67, I did find something interesting when doing additional testing. It seems that when there's noise on the band, sync goes to red as expected after TX finishes. However, when there's no noise (e.g. with audio loopback), it seems to stay in sync until I push the "resync" button. I did a little playing on the Codec2 side but it didn't seem to make any difference. Any ideas?

Hi @tmiw, I've been trying to reproduce this using an ALSA loopback device on Ubuntu.  When in full duplex, I don't think the Tx button does much, it just toggles the PTT logic.  So I just see continuous sync as the Tx signal is always present ... can you pls tell me how to repeat the issue you saw (or maybe how you did it)?

tmiw:
@drowe67, basically, disable full duplex and play back the sample .wav files. I also used a VB-Cable virtual audio cable as my "radio" sound card (since that's what my Hermes Lite 2 SDR does audio I/O though), if that helps.

tmiw:
BTW, PiHPSDR (what I use to access the HL2) was not running at the time I did the testing, so I was receiving no RX audio from my actual radio.

drowe67:
OK I managed to repeat it - just as you suggested.  With the API debug option on I can see there are no unique word errors after the virtual sound card starts feeding it zeros, so it stays in sync.  I think this was because I used the not so smart selection of an all-0 bit unique word for some of the modes, and through some fluke all zero input samples demodulates to the QPSK symbol for 0 bits.

Adding a tiny random component fixes it, e.g. in freedv_interface L457:
```
                rx_fdm[i].real = (float)input_buf[i] + 1E-6*(float)rand()/RAND_MAX;
```

However I guess it's a bug affecting practical operation, more a quirk we hit when testing. I'll add it as a codec 2 issue and think about a better solution.

tmiw:
@drowe67, sounds good. In the meantime, the build I sent out on the 31st seems to have helped with Peter's issue he mentioned on the list.

src/freedv_interface.cpp
```diff
@@ -122,10 +144,29 @@ void FreeDVInterface::setRunTimeOptions(int clip, int bpf, int phaseEstBW, int p
 {
     for (auto& dv : dvObjects_)
     {
-        freedv_set_clip(dv, clip);   // 700C/700D/700E/2020
-        freedv_set_tx_bpf(dv, bpf);  // 700D/700E
-        freedv_set_phase_est_bandwidth_mode(dv, phaseEstBW); // 700D/2020
-        freedv_set_dpsk(dv, phaseEstDPSK);  // 700D/2020
+        switch(freedv_get_mode(dv))
```
drowe67, line 147: I think the original implementation above is OK and less complicated.  Take a look at the implementation of these functions in https://github.com/drowe67/codec2/blob/master/src/freedv_api.c - they don't do anything unless they are called in the relevant mode.  If that's not the case, I'm happy to tweak `freedv_api.c`

src/freedv_interface.cpp
```diff
@@ -464,7 +505,7 @@ int FreeDVInterface::processRxAudio(
             }
 
             // Optional frequency shifting
-            freq_shift_coh(rx_fdm_offset, rx_fdm, rxFreqOffsetHz, freedv_get_modem_sample_rate(dv), rxFreqOffsetPhaseRect, nin);
+            freq_shift_coh(rx_fdm_offset, rx_fdm, rxFreqOffsetHz, freedv_get_modem_sample_rate(dv), rxFreqOffsetPhaseRectObjs_[index], nin);
```
drowe67, line 508: Yep this is the correct way to handle the phase state variable while looping through modes :+1: 

Each mode will have a different `nin`, so the phase will be rotated by a different amount for any `rx_fdm_offset` other than 0.  So you need a separate phase state variable for each mode.

Also note the modem sample rate can be mode specific (although it's 8000 Hz for most modes).  This (or a a related issue) might be why the system was choking on 2400B.

src/freedv_interface.cpp
```diff
@@ -15,14 +36,137 @@ FreeDVInterface::~FreeDVInterface()
...
+        struct FIFO* errFifo = codec2_fifo_create(2*freedv_get_sz_error_pattern(dv) + 1);
+        assert(errFifo != nullptr);
+        
+        errorFifos_.push_back(errFifo);
+        
+        auto tmpPtr = new COMP();
+        tmpPtr->real = cos(0.0);
+        tmpPtr->imag = sin(0.0);
+        txFreqOffsetPhaseRectObjs_.push_back(tmpPtr);
+
```
drowe67, line 71: Not sure if you need mode specific tx phase variables, as you're only tx-ing one channel at a time, and the state get's reset at the start of every PTT "over".

drowe67:
@tmiw re Tx echo issues, should main.cpp:2679:
```
nout = resample(cbData->outsrc1, outsound_card, outfreedv, g_soundCard1SampleRate, freedv_samplerate, 10*N48, nfreedv);
```
Use `FreeDVInterface::getTxModemSampleRate` rather than `freedv_samplerate`.  The latter will be set to the maximum sample rate of all modes.

src/freedv_interface.cpp
tmiw: Reverted back to the prior implementation.

src/freedv_interface.cpp
tmiw: I tried reenabling 2400B after making this change and the sample rate and it seems that receiving 2400B works properly (albeit no sync indication) but other modes don't work. I left 2400B and 800XA disabled for auto-RX for now.

src/freedv_interface.cpp
tmiw: Yep, looks like a single one is fine. 👍 I consolidated the multiple ones for TX into one.

tmiw:
@drowe67, agreed on getTxModemSampleRate. I also made updates to re-initialize Speex on TX mode changes as well as added locking around the TX areas in case there's some sort of race condition involved.

src/freedv_interface.cpp
drowe67: Another complication with 2400B is the sample rate (48kHz) is > some sound card rates.  This might be causing a buffer overflow somewhere.

Not sure what's going on with 800XA....

tmiw:
@drowe67, I was able to consistently reproduce the PSK Reporter crash (previously reported by others) on this branch with the following procedure:

1. Turn on radio/SDR application and set AF to zero to suppress receive audio.
2. Turn on FreeDV in full duplex mode and PSK Reporter functionality enabled and wait until your full callsign appears on the bottom.
3. Turn up AF audio on the radio to 100% and wait a few minutes.

I tried to partially port the fix in #98 but that only lengthened the time before it crashed. I was able to fully fix it (at least for the 30 minutes or so I let it run to test) for [9572a3f](https://github.com/drowe67/freedv-gui/pull/111/commits/9572a3fdda8b5625013475be9a28ef459013249b).

Anyway, this was while I was trying to duplicate the 1600 text issue Siegfried Jackstien reported. _Now_ I'm going to see if I can duplicate the actual issue I was checking out. :)

drowe67:
Sounds like good progress :+1:

So I was wondering - if you are running full duplex, why do you need the radio/SDR connected at all?

>     1. Turn on radio/SDR application and set AF to zero to suppress receive audio.

When I test in full duplex I use a loopback sound device for the radio tx/rx audio and the laptops internal sounds card for mix/speaker.

tmiw:
@drowe67, PSK Reporter functionality currently requires Hamlib to be enabled, hence the need for a radio. That's another thing that seems like should be fixed long term.

Additionally, the radio's audio provides enough interference to stop proper decode of characters from the TX side.

drowe67:
> Additionally, the radio's audio provides enough interference to stop proper decode of characters from the TX side.

You could also use the AWGN noise generator built into freedv-gui.

tmiw:
Yep, that should be viable once PSK Reporter's extended to allow usage without Hamlib. 

Anyway, I'm going to send out one more build for testing. Hopefully we'll be able to merge this PR if there are no other major findings. 👍

tmiw:
@drowe67, I'm going to create a separate PR for Keith Prosser's request after this one's merged to avoid cluttering this one up further if that's okay with you. Otherwise it doesn't seem like there are any major issues with the latest build.

(BTW the changes for his request are already in https://github.com/tmiw/freedv-gui/tree/ui-enhancement if you want to take an early look. The branch is based on the current bugfixes one, so it'll end up also showing all of the previous 17 commits in this PR if I created a new PR now.)

drowe67:
> @drowe67, I'm going to create a separate PR for Keith Prosser's request after this one's merged to avoid cluttering this one up further if that's okay with you. Otherwise it doesn't seem like there are any major issues with the latest build.

Yep - way to go :slightly_smiling_face: Good news that the build is stabilizing :+1: 

I'm attempting repeat Peter vk5apr's 1600 test issue that he mentioned on the mailing list on & April.  I think it's a ctrl code thing.

tmiw:
@drowe67, I fixed that issue in 8cfa2b9, actually, if it's the same one I'm thinking about. 👍 

Meanwhile, I haven't been able to duplicate Kanda's latest issue. Any thoughts on that one?

drowemyriota:
> Meanwhile, I haven't been able to duplicate Kanda's latest issue. Any thoughts on that one?

Which issue was that?  I can't recall :thinking:

tmiw:
From 5:44am PDT (UTC-7) this morning on the digitalvoice list:

> Thank you for your continued support.

> I noticed that I was communicating with a friend today.
I'm using FreeDV 1.5.3, which was published on the date 20210406
Pressing the analog button while communicating in 700D mode
When I checked the original signal that was being received,
Mixed with pure digital signals in 700D mode,
You can see a fairly large amount of noise.

> When the same signal was received by the program of Ver 1.5.2,
Only pure digital sound is received, no noise is confirmed in the digital signal.

> Is there any obstacle to noise when it is converted to analog?
Can you confirm it there as well?

> This symptom occurs when you click on Modem's Multible RX Operation Simultaneously Decode All HF Modes.
You can confirm that you are receiving.

I played a previous recording that I made and pushed the Analog button during playback (both with "Simultaneously Decode All HF Modes" on and off) and it sounded the same to me both times. I could be misunderstanding his email, though.

drowe67:
OK, I think he's suggesting you get simultaneous decoded DV and off air Rx audio, or perhaps chopping between the two.  This could perhaps be a corner case in that same chunk of `freedv_bits_to_speech()` that has some tricky logic.  Lets see if it crops up for anyone else of becomes repeatable ...

tmiw:
I asked for CPU usage and his system specs as well in case it's something involving limited system resources (although I'm thinking that's probably not the case).

tmiw:
@drowe67, I haven't heard anything else; Siegfried seems to think it's an issue with the machine's performance so he's going to try another setup soon. 

In the meantime, should we merge now and generate the final 1.5.3 build?

PR #116, opened 2021-04-19
Avoid clipping of final speech data when stopping transmit.

tmiw:
I noticed that I would occasionally hear the last tiny bit of speech from myself when I stopped transmitting. This PR should ensure that this bit is actually transmitted prior to stopping PTT.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2368,6 +2368,9 @@ void txRxProcessing()
 
     paCallBackData  *cbData = g_rxUserdata;
 
+    // Lock the mode mutex so that TX state doesn't change on us during processing.
+    txModeChangeMutex.Lock();
+    
     // Buffers re-used by tx and rx processing.  We take samples from
     // the sound card, and resample them for the freedv modem input
     // sample rate.  Typically the sound card is running at 48 or 44.1
@@ -2575,11 +2578,8 @@ void txRxProcessing()
     //
     //  TX side processing --------------------------------------------
     //
-
-    if (((g_nSoundCards == 2) && ((g_half_duplex && g_tx) || !g_half_duplex))) {
-        // Lock the mode mutex so that TX state doesn't change on us during processing.
-        txModeChangeMutex.Lock();
-        
+    
+    if (((g_nSoundCards == 2) && ((g_half_duplex && g_tx) || !g_half_duplex))) {    
         // This while loop locks the modulator to the sample rate of
         // sound card 1.  We want to make sure that modulator samples
         // are uninterrupted by differences in sample rate between
@@ -2648,7 +2648,7 @@ void txRxProcessing()
             g_mutexProtectingCallbackData.Unlock();
 
             resample_for_plot(g_plotSpeechInFifo, infreedv, nout, freedvInterface.getTxSpeechSampleRate());
-
+            
             nfreedv = freedvInterface.getTxNNomModemSamples();
 
             if (g_analog) {
@@ -2716,9 +2716,9 @@ void txRxProcessing()
             }
             codec2_fifo_write(cbData->outfifo1, outsound_card, nout);
         }
-        
-        txModeChangeMutex.Unlock();
     }
+
+    txModeChangeMutex.Unlock();
 
     if (g_dump_timing) {
         fprintf(stderr, "%4ld", sw.Time());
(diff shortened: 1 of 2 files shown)


tmiw:
@drowe67, I'm still seeing a bit of the final transmitted signal in the waterfall after TX stops, but I no longer hear myself in the speakers. Not sure if/how we should suppress that, though. Any ideas?

drowe67:
Hi @tmiw - OK that's a good catch.  Yes we the swap between Tx/Rx should be orderly.

Just thinking out loud - but is there any way to do it with Mutexes?  I Like to avoid synchronisation objects when I can, although I realise they are sometime unavoidable.

I'll think about it some more too ... this one needs fresh coffee in my AM :slightly_smiling_face:

tmiw:
I added some additional logic to make the UI part of the TX/RX switching wait until txRxProcessing executes. That doesn't look like it helped with the waterfall artifact (below) but it does seem to avoid the clipping of the last ~20-40ms of audio I've noticed from others:

Screen Shot 2021-04-20 at 1 19 21 AM

Any ideas on that would definitely be appreciated. 👍

tmiw:
BTW, I've also noticed a bit of noise at the very beginning of my transmissions even with these changes. Here's an example playing back a recording from one of the web-based SDRs I test through:

Screen Shot 2021-04-20 at 1 25 08 AM

It kinda looks like it's encoding the last bit of RX audio just before switching to the modulated output but it's possible it's some issue with my radio setup locally. @drowe67, are you able to duplicate similar?

tmiw:
@drowe67, so I think the extra noise at the beginning is related to how I had my SDR set up. I managed to install a second set of virtual audio cables and set separate devices for both RX from and TX to the radio and I no longer see the noise in the web SDR's waterfall. (I also tried clearing all the FIFOs that it could have possibly been prior to enabling TX but that didn't seem to do anything, implying it was probably a driver or SDR software issue.)

drowe67:
HI Mooneer - I'm a bit busy with my day job atm but will look into this one in more detail over the weekend :+1: I'd like to see if I can reproduce the behavior here.

drowe67:
Hi @tmiw - I'll try to reproduce this issue myself today.  Reading the comments above there seems to be a few related issues, possibly combined with site-specific configuration.

Can you pls confirm if this procedure is OK to produce the issue:
1. Setup: Set up for an over the air test with my radio (in my case a IC7200) Any particular mode I should use?  Should the auto-mode switching be enabled?
1. Start Tx
2. Switch to Rx
3. Listen for the end of my own Tx audio in the Rx audio.

drowe67:
BTW you should now be able to commit directly to the drowe67/freedv-gui repo, if you want to raise PRs there.  That way it's a bit easier to try different branches. However it's also fine to continue working off your Github repo :slightly_smiling_face: - whatever works best for you.

drowe67:
Hi Mooneer, I just tried the procedure above using 700D:

1. With multi-rx disabled I get just a short click when I transition from Tx/Rx.
2. With multi-rx enabled I get a short burst (maybe 100ms?) of garbled audio, but not a proper decode, more like the decoder working with random bits.

In both cases the sync LED lights very briefly as I enter the Rx state.  In both cases it doesn't seem very noticeable for me, and even at a low (1dB) squelch setting doesn't break squelch - so I hear nothing if squelch is enabled. I don't think it would bother me in normal operation.

If there is a common buffer in use might it be enough just to use a separate buffer for TX and Rx that is cleared when we transition?

If you still feel a need to fix the problem with a PR, I'm happy to help, but would like to explore a way that uses fewer synchronisation objects as I feel they complicate the code and can lead to problems of their own (like deadlocks).  An alternative might be a way for the main thread to message the TxRx thread when a PTT up/down event occurs.

tmiw:
@drowe67, yeah, I suspect a lot of it does come down to local setup (though a lot of people do use SDRs and virtual sound cards). 

As for the PR, it does seem to fix the clipped audio but not the remaining waterfall artifact with the existing wxMutex already used for TX mode switching. I'm honestly not sure it's worthwhile to try to fix the artifact, especially since it might be a good confirmation that one isn't currently transmitting. I'll keep in mind locking issues when considering additional mutexes in the future, though.

PR #119, opened 2021-05-02
Adds support for alternative configuration files.

### LINKED ISSUE #104: [Feature Request] Ability to use a different configuration depending on location
I was resetting my FreeDV configuration back to my "home" setup (Hermes Lite 2 SDR via a virtual sound card) after operating portable this past weekend and realized that this would be something that might be useful for others as well. Figured I'd put it here as a TODO for me (or someone else) to get to eventually. :)

Anyway, here's what I'm thinking so far. Feel free to update as needed.

Configuration options that should be updated on a per-location basis:
* Sound card configuration
* PTT configuration
* PSK Reporter grid square
* Regular text field (if PSK Reporter not used)

Configuration options that should remain the same regardless of location:
* PSK Reporter callsign (?)
* Audio filters
* All remaining options in Tools->Options.

### LINKED ISSUE #118: command line freedv-gui to support testing
A way to run freedv-gui from the command line, put it into a mode (e.g 700D), set up input and output sound cards, and feed a signal through it.  This way basic freedv-gui operation can be tested using a GitHub action.  

1. Not sure what happens to the actual screen on a headless GitHub action machine.
1. Rx only is OK for a first pass.  
1. ALSA loopback sound cards could be used to simulate real sound card hardware (see `codec2/unittest/tnc*` tests for examples).
1. Sox could be used for command line sample I/O from sound cards to files.

tmiw:
This fully resolves #104 and partly resolves #118 by providing a -f (or --config) command line argument to allow usage of a different configuration file.


drowe67:
Nice work @tmiw  - that's a very neat solution :+1:

PR #122, opened 2021-05-03
Early Tx termination

drowe67:
The last few 100ms of Tx audio tends to be cut off when we transition from Tx to Rx.

### DIFF
--- a/portaudio-macos.patch
+++ b/portaudio-macos.patch
@@ -1,7 +1,7 @@
 diff -ur portaudio.orig/configure portaudio/configure
 --- portaudio.orig/configure	2020-11-19 17:23:04.000000000 -0800
-+++ portaudio/configure	2020-12-01 20:56:35.000000000 -0800
-@@ -15884,13 +15884,19 @@
++++ portaudio/configure	2021-04-29 01:25:21.000000000 -0700
+@@ -15884,13 +15884,25 @@
                elif xcodebuild -version -sdk macosx10.15 Path >/dev/null 2>&1 ; then
                   mac_version_min="-mmacosx-version-min=10.4"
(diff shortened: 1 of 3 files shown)


drowe67:
@timw thinking about the way the signals flow, and where the latency is:

Mic -> Mic input FIFO A - > TxRx thread -> modulated Tx FIFO B -> sound card output C to Rig Mic

I'm assuming the operating system sound card driver and Rig latency C is small and can be ignored.  So I think we actually have a delay in A & B.  When we press PTT to end the over, there may be unprocessed mic audio samples in A.

So how about this design:
1. On press PTT to Tx-->Rx, we set `endingTx=true`, and sample the current value of `g_outfifo1_empty`. We then wait on `g_outfifo1_empty` changing in a sleep(20) while loop (or we could use the syncronisation objects as per the current design).
2. The thread filling A monitors `endingTx` and stops writing samples.
3. The existing samples in A flow through the TxRx thread into B until A empties.
4. B eventually empties after all the samples have been processed and output.  It then increments `g_outfifo1_empty`
5. The while loop/sync object in step (1) falls out and we tell Hamlib to take the rig out of Tx.

tmiw:
@drowe67, I added step 2 above to the current PR and it looks like that covered pretty much all the remaining audio that was truncated. I think there's still about ~20ms cut off but that seems within limits given what I understand about how my HL2 (and similar SDRs) work. Feel free to give it a try and let me know how it goes.

drowe67:
@tmiw - thanks.  I have an idea for a simpler implementation that I will try out in another branch in a few days so I don't mess this one up!  Will get back to you once I know if it works .... :slightly_smiling_face:

drowe67:
Alternate implementation based on this design  used in https://github.com/drowe67/freedv-gui/pull/123

PR #123, opened 2021-05-04
Early Tx Termination - 2 

drowe67:
Hi @tmiw pls tell me what you think of this version.  I've tested it on air and it seems to capture all of the Tx audio.  

I've left some debug code in so you can see it working. 

I'm open to using a synchronization object instead of the while loop.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2628,11 +2628,9 @@ void txRxProcessing()
             //     * If nothing is remaining in the FIFO and we're ending TX, we exit this processing loop.
             int available = codec2_fifo_used(cbData->infifo2);
             int toRead = ( !endingTx || available >= nsam_in_48 ) ? nsam_in_48 : available;
-            int nread = 0;
 
             if (toRead == 0 && endingTx) break;
-            nread = codec2_fifo_read(cbData->infifo2, insound_card, toRead);
-            assert(nread == 0);
+            codec2_fifo_read(cbData->infifo2, insound_card, toRead);
             
             nout = resample(cbData->insrc2, infreedv, insound_card, freedvInterface.getTxSpeechSampleRate(), g_soundCard2SampleRate, 10*N48, nsam_in_48);
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2623,13 +2623,11 @@ void txRxProcessing()
             // special case - if we are ending Tx then only process if we have samples available from mic
             if (endingTx && nread != 0)
             {
-                if (codec2_fifo_used(cbData->infifo2) == 0) break;
-                
-                short* tmp = (short*)&insound_card;
-                while (codec2_fifo_used(cbData->infifo2) > 0)
-                {
-                    codec2_fifo_read(cbData->infifo2, tmp++, 1);
-                }
+                int nused = codec2_fifo_used(cbData->infifo2);
+                fprintf(stderr, "mic nusued: %d\n", nused);
+                if (nused == 0) break;
+                // final read to empty fifo and consume last few mic samples
+                codec2_fifo_read(cbData->infifo2, insound_card, nused);
             }
 
             nout = resample(cbData->insrc2, infreedv, insound_card, freedvInterface.getTxSpeechSampleRate(), g_soundCard2SampleRate, 10*N48, nsam_in_48);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2618,17 +2618,15 @@ void txRxProcessing()
 
             // zero speech input just in case infifo2 underflows
             memset(insound_card, 0, nsam_in_48*sizeof(short));
-            int nread = codec2_fifo_read(cbData->infifo2, insound_card, nsam_in_48);
+            
+            // We try to read nsam_in_48 samples here. If we're stopping TX, we're okay
+            // with fewer so we can finish with whatever's already in the FIFO.
+            int available = codec2_fifo_used(cbData->infifo2);
+            int toRead = ( !endingTx || available >= nsam_in_48 ) ? nsam_in_48 : available;
+            int nread = 0;
 
-            // special case - if we are ending Tx then only process if we have samples available from mic
-            if (endingTx && nread != 0)
-            {
-                int nused = codec2_fifo_used(cbData->infifo2);
-                fprintf(stderr, "mic nusued: %d\n", nused);
-                if (nused == 0) break;
-                // final read to empty fifo and consume last few mic samples
-                codec2_fifo_read(cbData->infifo2, insound_card, nused);
-            }
+            if (toRead == 0 && endingTx) break;
+            nread = codec2_fifo_read(cbData->infifo2, insound_card, toRead);
 
             nout = resample(cbData->insrc2, infreedv, insound_card, freedvInterface.getTxSpeechSampleRate(), g_soundCard2SampleRate, 10*N48, nsam_in_48);
 
@@ -2806,7 +2804,10 @@ int MainFrame::rxCallback(
     // OK now set up output samples for this callback
 
     if (wptr) {
-         if (codec2_fifo_read(cbData->outfifo1, outdata, framesPerBuffer) == 0) {
+        int bytesUsed = codec2_fifo_used(cbData->outfifo1);
+        int bytesToRead = ( !endingTx || bytesUsed >= framesPerBuffer ) ? framesPerBuffer : bytesUsed;
+        memset(outdata, 0, MAX_FPB);
+        if (bytesToRead > 0 && codec2_fifo_read(cbData->outfifo1, outdata, bytesToRead) == 0) {
 
             // write signal to both channels if the device can support two channels.
             // Otherwise, we assume we're only dealing with one channel and write
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2632,7 +2632,8 @@ void txRxProcessing()
 
             if (toRead == 0 && endingTx) break;
             nread = codec2_fifo_read(cbData->infifo2, insound_card, toRead);
-
+            assert(nread == 0);
+            
             nout = resample(cbData->insrc2, infreedv, insound_card, freedvInterface.getTxSpeechSampleRate(), g_soundCard2SampleRate, 10*N48, nsam_in_48);
 
             // optionally use file for mic input signal
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2620,17 +2620,9 @@ void txRxProcessing()
             memset(insound_card, 0, nsam_in_48*sizeof(short));
             
             // There may be recorded audio left to encode while ending TX. To handle this,
-            // first we grab how many bytes are currently used by the FIFO. Then we determine
-            // the amount to read from the FIFO as follows:
-            //     * If transmit is not in the process of ending, we attempt to read nsam_in_48 from the FIFO.
-            //     * If transmit is ending and more than framesPerBuffer remain in the FIFO, we attempt to read nsam_in_48 from the FIFO.
-            //     * If transmit is ending and less than framesPerBuffer remain in the FIFO, we read the number currently in the FIFO.
-            //     * If nothing is remaining in the FIFO and we're ending TX, we exit this processing loop.
-            int available = codec2_fifo_used(cbData->infifo2);
-            int toRead = ( !endingTx || available >= nsam_in_48 ) ? nsam_in_48 : available;
-
-            if (toRead == 0 && endingTx) break;
-            codec2_fifo_read(cbData->infifo2, insound_card, toRead);
+            // we keep reading from the FIFO until we have less than nsam_in_48 samples available.
+            int nread = codec2_fifo_read(cbData->infifo2, insound_card, nsam_in_48);
+            if (nread != 0 && endingTx) break;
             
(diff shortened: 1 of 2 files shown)


tmiw:
@drowe67, this doesn't seem to introduce any new bugs. However, with my normal test script, it seems to cut off after "four".

drowe67:
> @drowe67, this doesn't seem to introduce any new bugs. However, with my normal test script, it seems to cut off after "four".

Thanks for trying that.  The cut off is just right when I test here with my IC7200.  I think the design guarantees there is no audio left in freedv-gui (both the mic and tx audio FIFO must be empty), although I may have missed something. I wonder what could be different on our two systems?  Perhaps additional delay in the operating system or radio?
 
> EDIT: actually, I'm seeing squelch issues with "Simultaneously Decode All HF Modes" enabled. I think that might be at least partly a Codec2 issue but I haven't looked into it yet (and I saw it on the other branch too).

I did recently refactor the SNR estimator for the modes that use the OFDM modem (700D/E, 2020).  Can you tell me some more about the problem and how to reproduce it?  I might also be able to test at the FreeDV API level.

src/main.cpp
```diff
@@ -2614,7 +2618,10 @@ void txRxProcessing()
 
             // zero speech input just in case infifo2 underflows
             memset(insound_card, 0, nsam_in_48*sizeof(short));
-            codec2_fifo_read(cbData->infifo2, insound_card, nsam_in_48);
+            int nread = codec2_fifo_read(cbData->infifo2, insound_card, nsam_in_48);
+
+            // special case - if we are ending Tx then only process if we have samples available from mic
+            if (endingTx && (nread != nsam_in_48)) break;
```
tmiw, line 2624: codec2_fifo_read() seems to return -1 if there isn't enough data and 0 otherwise. Could that be why significant chunks of audio still seem to be getting cut?

tmiw:
> I did recently refactor the SNR estimator for the modes that use the OFDM modem (700D/E, 2020). Can you tell me some more about the problem and how to reproduce it? I might also be able to test at the FreeDV API level.

Basically, what I'm seeing is that FreeDV seems to sync (green text) and display the correct mode being decoded but no audio. Additionally, the SNR is seemingly greater than the squelch level. I hear decoded audio once I uncheck the squelch box. I'm not sure it's reproducible 100% of the time, either, which may mean that it won't necessarily matter whether "Simultaneously Decode All HF Modes" is enabled.

tmiw:
OK, it looks like I can trigger it after playing back recorded audio 4 or 5 times or so. SNR seems to be stuck at 0dB when this happens, too, and there's no audio even when I set the squelch to -5 or so (but audio does play once squelch is fully disabled). Hopefully this helps!

tmiw:
@drowe67, I was able to improve things a bit by processing whatever bytes remain in the FIFO (not just a multiple of nsam_in_48). I think that was why the other implementation seemed to do a bit better as well.

src/main.cpp
drowe67: Good catch @tmiw - I should check the return codes on stuff I wrote 10 years ago :grinning:

drowe67:
> @drowe67, I was able to improve things a bit by processing whatever bytes remain in the FIFO (not just a multiple of nsam_in_48). I think that was why the other implementation seemed to do a bit better as well.

Cool :+1: I've refactored your design a little to remove the while loop.  I've left the debug `fprintf()` in to see if your system is any different to mine, I get:
```
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
mic nusued: 1760
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 462
mic nusued: 0
mic nusued: 0
sample: 462 g_outfifo1_empty: 470
mic nusued: 0
```
So there are 1760 unused samples in the "mic" fifo at the start on my machine, which I think is about 40ms.  That's probably not noticeable but hey its good to be thorough :slightly_smiling_face:

drowe67:
@tmiw BTW I'm really appreciating some of the ease of use features you have developed.  Plugging in (and re-plugging ) USB sound devices is now "painless", as your features maps them correctly despite re-ordering.  Well done :100:

drowe67:
> OK, it looks like I can trigger it after playing back recorded audio 4 or 5 times or so. SNR seems to be stuck at 0dB when this happens, too, and there's no audio even when I set the squelch to -5 or so (but audio does play once squelch is fully disabled). Hopefully this helps!

Thanks for the heads up - I will try to repeat and attempt to debug this weekend :thinking:

tmiw:
This last test with the commit I just pushed took 13 times through the loop before we ran out of bytes in the FIFO:

```
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2916
sample: 2916 g_outfifo1_empty: 2920
rig.c(2705):rig_set_ptt entered
```

(and seemed to produce similar results as with the other branch)

I also added an increment for i because it apparently got missed before (and caused an infinite loop). One upside of the synchronization objects is that we don't have a 50ms minimum wait, but it's also not a time period that would be really noticeable as you mention.

> @tmiw BTW I'm really appreciating some of the ease of use features you have developed. Plugging in (and re-plugging ) USB sound devices is now "painless", as your features maps them correctly despite re-ordering. Well done 💯

No worries 👍 I think it can still be improved a bit but that might be due to some quirks macOS has (or my local setup).

drowe67:
> Audio was still getting cut off for me as well, so I figured it was getting cut off here too. I added a comment here to clarify the logic.

I'm puzzled by that :thinking: as I don't get any cut off, and the "fractional" parts of the sample remaining in those FIFOs seem very small on my machine (40ms or so) - which should be inaudible.  Can pls you confirm that the additional code to "use/drain" the remaining samples from the two FIFOs actually helps? I happy for us to employ this code if it helps, but want to make sure it's actually doing something useful.

Thanks for the comments - that spells it out for me well :slightly_smiling_face: Another approach is to write the actual source code in a way that makes the logic easy to follow (this also aids maintenance), but I'm OK with either approach. I tend to avoid the `?` constructs myself as I find them harder to understand than if-then, but that's personal style I guess.

I had a similar problem in some of the `freedv_api.c` logic around squelch and audio pass through - and needed to list all the use cases it needed to handle to get is straight in my head.

src/ongui.cpp
```diff
@@ -430,6 +432,19 @@ void MainFrame::togglePTT(void) {
...
+        endingTx = true;
+        //while (g_outfifo1_empty == sample) wxThread::Sleep(20);
+        int i = 0;
+        while ((i < 20) && (g_outfifo1_empty == sample)) {
+            fprintf(stderr, "sample: %d g_outfifo1_empty: %d\n", sample, g_outfifo1_empty);
+            i++;
+            wxThread::Sleep(50);
+        }
+        fprintf(stderr, "sample: %d g_outfifo1_empty: %d\n", sample, g_outfifo1_empty);
+
```
drowe67, line 447: I figure we just need to:
1. Answer the question on if the fractional FIFO code helps the cut off issue.
1. Clean this up, ie remove the fprintfs and decide on the time out (I suggest 1000ms)
... and we're done :+1: 

Is there anything else you think we need to do?  It's been a real joint effort!

tmiw:
> I'm puzzled by that 🤔 as I don't get any cut off, and the "fractional" parts of the sample remaining in those FIFOs seem very small on my machine (40ms or so) - which should be inaudible. Can pls you confirm that the additional code to "use/drain" the remaining samples from the two FIFOs actually helps? I happy for us to employ this code if it helps, but want to make sure it's actually doing something useful.

I added some debug prints in rxCallback as follows:

```
        int bytesToRead = ( !endingTx || bytesUsed >= framesPerBuffer ) ? framesPerBuffer : bytesUsed*;
        
        if (endingTx && bytesToRead != framesPerBuffer)
        {
            fprintf(stderr, "Bytes remaining in the TX FIFO: %d\n", bytesToRead);
        }
```

Most of the runs gave me stuff similar to the following:

```
sample: 11822 g_outfifo1_empty: 11822
Bytes remaining in the TX FIFO: 0
Bytes remaining in the TX FIFO: 0
Bytes remaining in the TX FIFO: 0
Bytes remaining in the TX FIFO: 0
sample: 11822 g_outfifo1_empty: 11826
```

However, there was one that gave me the following, which indicates that it could help in some instances:

```
sample: 944 g_outfifo1_empty: 944
Bytes remaining in the TX FIFO: 137
Bytes remaining in the TX FIFO: 0
Bytes remaining in the TX FIFO: 0
sample: 944 g_outfifo1_empty: 946
```

Whether that's actually audible, though... /shrug (I tried recording from a local KiwiSDR with and without the partial FIFO handling in rxCallback and it was really hard to tell the difference, if there was one at all. Both seemed to end at the end of "five" in the test script.)

src/ongui.cpp
tmiw: Nope, that sounds about it. I removed the debug prints that I could see, and the timeout seems to already be 1000ms based on the current logic (50ms * 20 tries). 👍

tmiw:
Actually, I recorded myself saying the test script in Audacity and it seems to take ~250ms for me to say "six". I've been using 700D for tests; could it be simply that there's not enough time after pushing the PTT button to stop transmitting to actually encode that last bit? Or something inherent to that mode? I'll try some other modes to confirm whether that could be the case.

tmiw:
@drowe67, just removed the partial FIFO logic. Would there be any value in adding a 160-180ms sleep just prior to setting endingTx = true? This would only happen if the current TX mode is one of the ones with long frames (700D/E/2020).

drowe67:
> @drowe67, just removed the partial FIFO logic. Would there be any value in adding a 160-180ms sleep just prior to setting endingTx = true? This would only happen if the current TX mode is one of the ones with long frames (700D/E/2020).

Hmm, interesting idea :thinking: I guess what we want is enough time for any samples in the OS, and PortAudio subsystems, to make it into the mic FIFO. Once it gets to that FIFO, I think we're pretty sure the samples are being processed and dispatched OK, with our careful work to date.

It would be good to avoid **too** much delay as we don't want unnecessary latency in modes like 1600/700E/700C that have smaller processing frames.  Maybe we could try a delay that is a fraction (say half) of the current modes processing block size?  That is available (in samples at Fs=8000 Hz) from functions like `freedv_get_n_speech_samples()` which I think you've seen before.

tmiw:
@drowe67, I added a bit of code to sleep as per your suggestion. I think it helped a little bit but I'm not 100% sure. Can you give it a shot?

drowe67:
> @drowe67, I added a bit of code to sleep as per your suggestion. I think it helped a little bit but I'm not 100% sure. Can you give it a shot?

Hi @tmiw I just tried it on my IC7200, and it works great for me.  However the previous versions works well for me too :slightly_smiling_face: Perhaps we could be down to rig or other site specific issues like the OS/sound drivers etc?  However overall it does work great - much better than before - so I'm pleased with our joint effort.  Sound card stuff is tricky ....

I'm happy to proceed with or without https://github.com/drowe67/freedv-gui/pull/123/commits/3c238f6dbc52310ea3ce1f6921bd14d2984e8dc2 - pls let me know if you have a preference, and we can merge.

drowe67:
> OK, it looks like I can trigger it after playing back recorded audio 4 or 5 times or so. SNR seems to be stuck at 0dB when this happens, too, and there's no audio even when I set the squelch to -5 or so (but audio does play once squelch is fully disabled). Hopefully this helps!

Hi @tmiw - I've tried to reproduce this with a few off air files but I can't seem to make it happen for me.  Can you pls email me the file you were using?  If you can reproduce it, also try the "API Verbose" option - that will print out some stats (including SNR) from the FreeDV API which might give us a clue.

tmiw:
> I'm happy to proceed with or without [3c238f6](https://github.com/drowe67/freedv-gui/commit/3c238f6dbc52310ea3ce1f6921bd14d2984e8dc2) - pls let me know if you have a preference, and we can merge.

Let's proceed with it. It should be trivial to revert if it ends up causing issues.

> I've tried to reproduce this with a few off air files but I can't seem to make it happen for me. Can you pls email me the file you were using? If you can reproduce it, also try the "API Verbose" option - that will print out some stats (including SNR) from the FreeDV API which might give us a clue.

Let me see if I can reproduce it again and get back to you.

PR #124, opened 2021-05-08
Bootstrap wxWidgets for Windows builds

tmiw:
This also updates the Docker container to Fedora 34.

@drowe67, unfortunately this significantly lengthens compile times since it has to compile wxWidgets. Eventually it would be good if the wxWidgets packages used before are updated (or we find some sort of way to only compile updated files instead of everything).


drowe67:
Thanks @tmiw, I'm part way through running it here.  I imagine the built-in packages will catch up in time, I recall when we started we would have to build wxWidgets from scratch even for Linux.

Are we running the same wxWidgets version on Linux and MacOS?  
Is that wxWidgets version a pre-compiled package on those operating systems?

drowe67:
@tmiw so I built the Fedora 34 environment:
```
docker-compose -f docker-compose-win.yml build
```
However I hit this when I tried to run it:
```
david@xps15:~/freedv-gui/docker ms-packaging $ ./freedv_build_windows.sh 64
+ '[' -z ']'
+ FDV_GIT_BRANCH=master
+ '[' -z ']'
+ FDV_GIT_REPO=https://github.com/drowe67/freedv-gui.git
+ FDV_CMAKE=mingw64-cmake
+ '[' 1 -eq 1 ']'
+ '[' 64 -eq 32 ']'
+ log=build_log.txt
+ FDV_CLEAN=
+ FDV_CMAKE=mingw64-cmake
+ FDV_GIT_REPO=https://github.com/drowe67/freedv-gui.git
+ FDV_GIT_BRANCH=master
+ docker-compose -f docker-compose-win.yml up --remove-orphans
Starting fdv_win_fed34_c ... 
Starting fdv_win_fed34_c ... done
++ cat build_log.txt
++ sed -n 's/.*package: \(.*exe\) .*/\1/p'
+ package_docker_path=
+ echo

+ docker cp fdv_win_fed34_c: .
Error response from daemon: bad parameter: pathcannot be empty
```

tmiw:
@drowe67, so you'll probably need to provide FDV_CLEAN=1 to the script for the first initial build, e.g.:

```
david@xps15:~/freedv-gui/docker ms-packaging $ FDV_CLEAN=1 ./freedv_build_windows.sh 64
```

After the initial build finishes, you should be able to leave off FDV_CLEAN=1 and have it only do a git pull on freedv-gui and a make/make package. I'm testing to make sure that's actually the case on my end.

drowe67:
> ```
> david@xps15:~/freedv-gui/docker ms-packaging $ FDV_CLEAN=1 ./freedv_build_windows.sh 64
> ```

:thinking: that didn't seem to work either, have attached the log file: [build_log.txt](https://github.com/drowe67/freedv-gui/files/6446606/build_log.txt).

hobbes1069:
Fedora already has wxWidgets 3.0.4 built for Windows, which we should be able to use:
https://src.fedoraproject.org/rpms/mingw-wxWidgets3

tmiw:
@drowe67, can you uncomment the following line in freedv_build_windows.sh (and comment the other one) and try again:

```
FDV_CLEAN=$FDV_CLEAN FDV_CMAKE=$FDV_CMAKE FDV_GIT_REPO=$FDV_GIT_REPO FDV_GIT_BRANCH=$FDV_GIT_BRANCH docker-compose -f docker-compose-win.yml up --build --force-recreate --remove-orphans > $log
```

Once that's done, you can undo the changes in freedv_build_windows.sh and execute the script without FDV_CLEAN=1.

@hobbes1069, I was mainly trying to use wxWidgets 3.1 as part of the accessibility work I'm doing (in case it fixes stuff with it vs. 3.0). Additionally, the 3.0 package it was previously trying to use doesn't work in 34; FreeDV complains on startup about the wxWidgets ABI versions being incorrect despite no indication that there's more than one version installed in the Docker container.

tmiw:
BTW I forgot to mention that this would bring the Windows builds in line with macOS, as the latter also uses 3.1 (due to required support for dark mode).

hobbes1069:
I can see about getting the package updated if that helps.

tmiw:
@hobbes1069, that would definitely be good longer term. 

Actually, I wonder if a general "FDV_EXTRA_OPTIONS" would be good along with FDV_CLEAN. Then we could specify other command arguments in addition to BOOTSTRAP_WXWIDGETS. Thoughts?

drowemyriota:
@tmiw as a goal for this PR I'd like to make the Docker build system really easy to use, e.g:
```
cd freedv-gui/docker
docker-compose -f docker-compose-win.yml build
./freedv_build_windows.sh 64
```
should "just work" without any special command line options, or manual edits of scripts.  IIRC the version currently in master meets that goal.

If there are any reasons it should be harder than that, lets carefully document why and how to use them.  I'll see if I can get this use case running here, hopefully just some tweaks to the scripts.

drowe67:
@tmiw OK so how about the following use cases:

1. Default build:
   ```
   docker-compose -f docker-compose-win.yml build
   ./freedv_build_windows.sh 64
   ```
   fresh clone of freedv-gui, build wxWidgets and freedv-gui.  Always works, even if it is slow.  No manual edits or special variables rqd.  Works on greenfields machines, or those where we have built before.

1. Subsequent builds, for more knowledgeable users:
   ```
    CLEAN=0 ./freedv_build_windows.sh 64
   ```  
   Just a git pull and rebuild of freedv-gui, doesn't rebuild wxWidgets, faster, for those making small edits who want to build and test quickly. Or if we want to get really clever:
    ```
    ./freedv_build_windows.sh 64 --noclean
   ```  
  

Does that work for you?  I'm happy to help with the scripting edits (or you are welcome to do them).

tmiw:
@drowe67, works for me if you have the time to make the changes. Otherwise, I can get to this when I have a chance. 👍

hobbes1069:
It took a lot of tweaking, but I have wxWidgets 3.1 packages built:
https://copr.fedorainfracloud.org/coprs/hobbes1069/mingw/

drowe67:
@tmiw I've added a CLI and set up the default behavior as above.  I've tried:
```
./freedv_build_windows.sh -d --build 64
and:
./freedv_build_windows.sh -d --noclean 32
```
which worked OK...

tmiw:
@drowe67, nice 👍 I also made bootstrapping wxWidgets an opt-in thing instead of being forced to do so.

@hobbes1069, is the package supposed to auto-update in the container if docker-compose uses --build --force-recreate? I just tried to test the new package and FreeDV still thinks it's using 3.0:

Screen Shot 2021-05-11 at 1 24 30 AM

hobbes1069:
Since the build is in my COPR, not in the standard repositories, we'll have to figure out how to enable them within docker.

hobbes1069:
Never mind, I see this in the Dockerfile:
https://github.com/drowe67/freedv-gui/blob/063b285fe3279de318fe27713a558838b1369b41/docker/fdv_win_fedora/Dockerfile#L15

It should be picking up 3.1.5 then.

tmiw:
docker-compose rm followed by docker-compose --no-cache seems to have pulled down the correct packages. Should that be in the documentation somewhere, or maybe thrown into the script somehow?

Anyway, it looks like it's behaving as expected now. Anything else that might be useful to add? (For instance, being able to specify Codec2/LPCNet repos/branches might be good.)

tmiw:
@drowe67, added and confirmed working. 👍

docker/freedv_build_windows.sh
```diff
@@ -1,22 +1,89 @@
...
+    print_help
 fi
 
 log=build_log.txt
-FDV_CMAKE=$FDV_CMAKE FDV_GIT_REPO=$FDV_GIT_REPO FDV_GIT_BRANCH=$FDV_GIT_BRANCH docker-compose -f docker-compose-win.yml up > $log
+
+if [ $FDV_REBUILD -eq 1 ]; then
+    docker-compose -f docker-compose-win.yml rm -f > $log
+    docker-compose -f docker-compose-win.yml build --no-cache >> $log
+else
```
drowe67, line 82: @timw should split these steps up?  For example say we make a minor change to the docker script, should we have to rebuild the entire image (installing every rpm)?  I watching it run atm `tail -f buld_log.txt` and it's pretty slow.

drowe67: Hmm, guess I'm assuming there is a docker command to just install the latest docker script, while keeping everything else ...

tmiw: I think that's technically what build without --no-cache is supposed to do. On the other hand, I'm not sure if people will necessarily know that there might be new packages available; I thought it would have upgraded wxWidgets on its own, for instance.

drowe67:
@tmiw I've added a `--rebuild` option, so say if you just modify `fdv_win_fedora/entrypoint.sh` then:
```
./freedv_build_windows.sh --build 32
```
... will update the Docker container with that script.  However if you would like to re-install everything then:
```
./freedv_build_windows.sh --rebuild 32
```
BTW `--noclean` isn't really doing much, as the codec2 and LPCNet repos are always rebuilt from scratch by `build_windows.sh`.  But it's still reasonably fast (a few minutes), and I figure we've done enough work on this PR.

Pls let me know if the current version of this PR is OK for you, and we can merge.

tmiw:
@drowe67, works for me. 👍

PR #125, opened 2021-05-09
Fix bugs found during testing of multiple config support

tmiw:
Looks like it wasn't actually saving config info after all when specifying a config file with -f. Tested this PR and it seems to resolve the issue.


drowe67:
@tmiw I've been doing some tests with master 063b285 and noticed it's forgetting a few config items like sound cards, and items checked on Options/modem and Options/debug flags when I restart.  This PR seems to have fixed that, it's remembering OK :+1: 

I tried it with:
```
./src/freedv -f my-freedv_conf.txt
```
and it set up a new config OK, and the file was stored in `~/my-freedv_conf.txt`, as per the relative location you specified in the manual.  Was there any reason for using a location relative to `~` under Linux?  It's kinda counter intuitive (I had to check the manual after I couldn't find the file).  I guess it makes sense for Windows registry paths.

drowe67:
I also noticed some unused variables being saved to the config in https://github.com/drowe67/freedv-gui/pull/125/commits/313277955b623833c108463162acfa2edd0bcaed

tmiw:
@drowe67, the ~ thing is a wxWidgets decision. We could potentially make it use elsewhere if we wanted, but it probably doesn't matter as long as it's documented somewhere.

Also, there seems to be options dialog code as well that uses the options that were removed, so that should probably be cleaned up too.

tmiw:
@drowe67, done. 👍

Issue #126, opened 2021-05-13
On RaspberryPI, FreeDV GUI goes in a locked status when switching off 700C


UgoPoddine:
Hello David,

FreeDV GUI is now much more stable on RaspberryPI3 than it was two years ago, and that's really appreciated.
Also if minor issue, I would like to make you aware of the fact that, on Raspbian 5.10.17-v7+ #1414 (but it's longtime that i's happens), also FreeDV GUI 1.5.3 devel  c93345e crashes (or - better - comes in some kind of neverending loop status) when, after having worked in 700c mode, the "Stop" control key is pushed. 
This happens only on 700C, all other modes can be stopped as expected.
Before pushing "Stop" no issue : you can transmit, receive or going in "analog" mode without problem.
For switching off the FreeDV-gui it's then necessary to kill the process.

As usual, thank-you for your effort
Best 73s
IU1IPB ugo

tmiw:
How long do you typically have FreeDV running before you run into the issue? If possible, can you try running it from the terminal (ideally under gdb but just on its own might produce debug output that can help)?

UgoPoddine:
Hello Tmiw,

Thank-you for your so fast answer.
The issue happens always and also immediately after the startup of the gui : I start 700C and, I do nothing more, if I push the "stop" the neverending loop starts. It happens since longtime (perhaps one year or more) : I'm usually keep FreeDV up-to date.
At the moment I will attach the log to from the terminal, made starting FreeDV-GUI, starting 700C, doing nothing more, stopping 700C and finally killing the GUI by CTRL+C : if you need the debug, I will do tomorrow :-)

Regards
Ugo

[freedv_debug.log](https://github.com/drowe67/freedv-gui/files/6474104/freedv_debug.log)

drowe67:
Hi @UgoPoddine! 

Could you please tell me:
1. Were you using `build_linux.sh` to build freedv-gui?
1. What Git hash of codec2 is being used (just want to confirm its the current master https://github.com/drowe67/codec2/commit/306e424282b0941caf018e408b91e4e3385948d2).

drowe67:
@tmiw those internal ALSA messages are so annoying (on all Linux operating systems), I wonder if there is a way to make them quiet?

tmiw:
@drowe67, I imagine the messages will be less of an issue once the PulseAudio PR finally gets accepted into PortAudio master.

UgoPoddine:
Hello mates,

[...]

PR #127, opened 2021-05-15
Optimization 

drowe67:
Some optimization of freedv-gui in response to 700C not quite working on the RPi 3: https://github.com/drowe67/freedv-gui/issues/126

### DIFF
--- a/src/plot_scalar.cpp
+++ b/src/plot_scalar.cpp
@@ -191,7 +191,6 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
     // a e.g. 500 pixel wide display.
     int samplesPerPixel = m_samples / plotWidth;
     float compressedSamples[m_channels*plotWidth];
-    std::deque<float> runningSamples;
     int compressedIndex = 0;
     for (int cIndex = 0; cIndex < m_channels; cIndex++)
     {
--- a/src/plot_scalar.cpp
+++ b/src/plot_scalar.cpp
@@ -204,7 +204,7 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
             runningTotal += a;
             if (sIndex > 0 && sIndex % samplesPerPixel == 0)
             {
-                compressedSamples[compressedIndex++] = runningTotal; // / samplesPerPixel;
+                compressedSamples[compressedIndex++] = runningTotal;
                 runningTotal = 0;
             }
         }
--- a/src/plot_scalar.cpp
+++ b/src/plot_scalar.cpp
@@ -90,7 +90,6 @@ PlotScalar::~PlotScalar()
 //----------------------------------------------------------------
 void PlotScalar::add_new_sample(int channel, float sample)
 {
-    int i;
     int offset = channel*m_samples;
 
     assert(channel < m_channels);
--- a/src/plot_scalar.cpp
+++ b/src/plot_scalar.cpp
@@ -178,8 +178,8 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
     ctx->SetPen(wxPen(BLACK_COLOR, 0));
     ctx->DrawRectangle(plotX, plotY, plotWidth, plotHeight);
 
-    index_to_px = (float)m_rGrid.GetWidth()/m_samples;
-    a_to_py = (float)m_rGrid.GetHeight()/(m_a_max - m_a_min);
+    index_to_px = (float)plotWidth/m_samples;
+    a_to_py = (float)plotHeight/(m_a_max - m_a_min);
 
     wxPen pen;
     pen.SetColour(DARK_GREEN_COLOR);
@@ -202,7 +202,7 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
 
             // invert y axis and offset by minimum
 
-            y = m_rGrid.GetHeight() - a_to_py * a + m_a_min*a_to_py;
+            y = plotHeight - a_to_py * a + m_a_min*a_to_py;
 
             // regular point-point line graph
 
@@ -226,9 +226,9 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
                     assert(m_a_max > 0.0);
 
                     float norm = (log10(a) - log10(m_a_min))/(log10(m_a_max) - log10(m_a_min));
-                    y = m_rGrid.GetHeight()*(1.0 - norm);
+                    y = plotHeight*(1.0 - norm);
                 } else {
-                    y = m_rGrid.GetHeight() - a_to_py * a + m_a_min*a_to_py;
+                    y = plotHeight - a_to_py * a + m_a_min*a_to_py;
                 }
 
                 // use points to make a bar graph
@@ -237,7 +237,7 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
 
                 x1 = index_to_px * ((float)i - 0.5);
                 x2 = index_to_px * ((float)i + 0.5);
-                y1 = m_rGrid.GetHeight();
+                y1 = plotHeight;
                 x1 += PLOT_BORDER + XLEFT_OFFSET; x2 += PLOT_BORDER + XLEFT_OFFSET;
                 y1 += PLOT_BORDER;
                 path.MoveToPoint(x1, y1);
@@ -272,6 +272,9 @@ void PlotScalar::drawGraticule(wxGraphicsContext* ctx)
     float    sec_to_px;
     float    a_to_py;
 
+    int plotWidth = m_rGrid.GetWidth();
+    int plotHeight = m_rGrid.GetHeight();
+
     wxBrush ltGraphBkgBrush;
     ltGraphBkgBrush.SetStyle(wxBRUSHSTYLE_TRANSPARENT);
     ltGraphBkgBrush.SetColour(*wxBLACK);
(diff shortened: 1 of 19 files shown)


src/plot_scalar.cpp
```diff
@@ -148,59 +156,107 @@ void PlotScalar::draw(wxAutoBufferedPaintDC&  dc)
...
+    a_to_py = (float)plotHeight/(m_a_max - m_a_min);
 
     wxPen pen;
     pen.SetColour(DARK_GREEN_COLOR);
     pen.SetWidth(1);
-    dc.SetPen(pen);
-
+    ctx->SetPen(pen);
+
+    // Compress the samples as we don't need to draw thousands of lines for
```
tmiw, line 190: @drowe67, I'm not sure the below code (using the average) is a good way of compressing the samples but it seems to look okay in the GUI. This drops the CPU usage of the "From Radio" tab from ~55-60% on my 2019 MBP to ~30%.

tmiw: Or I guess running total right now, not average. 👍

srsampson:
Since freedv_gui was probably designed for a fast desktop or laptop, I wonder if changing all the double precision trig functions to single-precision might speed-up many math operations?

tmiw:
> Since freedv_gui was probably designed for a fast desktop or laptop, I wonder if changing all the double precision trig functions to single-precision might speed-up many math operations?

At least for me, a lot of the issue seemed to be GUI related and was manifesting in Instruments (macOS profiling tool) as a lot of code being executed deep in Core Graphics. The commits done so far reduce that to the point that I don't go above 30-35% CPU even before taking into account the latest Codec2 PR.

I also noticed that the bulk of what freedv-gui does is on two threads. It may be useful to parallelize more assuming it's possible; that may make devices like the Raspberry Pi work better.

drowe67:
@tmiw The OP bug in 

> Since freedv_gui was probably designed for a fast desktop or laptop, I wonder if changing all the double precision trig functions to single-precision might speed-up many math operations?

I imagine the PI does double precision float in hardware, also, there's not much DSP at the freedv-gui level.  Suspect the main bottleneck is high speed cache memory.

drowe67:
@timw, as per https://github.com/drowe67/freedv-gui/issues/126#issuecomment-841726593 - on the Pi 3B the OP bug seems to be fixed without the changes in this PR, and the optimizations in this PR don't seem to have much effect.  It might just be the Pi has some other bottleneck. Although it's still unclear if the Pi 3B can use 700C for real QSOs. 

On my Ubuntu laptop, 700C no signal, using `top`, looking at peak over a few seconds

| Center Panel | master CPU Load % | this PR CPU load % |
| --- | --- | --- |
| waterfall | 66 | 46 |
| spectrum | 44 | 57 |

So waterfall looks a bit faster, but spectrum has slowed down for me.  Can you pls add some measurements for your machine?

Can we pls roll back any changes that affect the current appearance (e.g. the plot scalar and scatter plots)? They don't seem to help the OP Issue on the Pi, and I can't recall any CPU load issues on other platforms.

tmiw:
@drowe67, I reversed the following changes from scatter and scalar plots:

1. add_new_samples_* using memcpy/memmove -- this is probably why the scatter plot looked weird before. Weirdly, it seems like it would work intuitively, but I guess not.
2. The "sample compression" stuff I tried to do in PlotScalar.

However, I did replace the optimization in PlotScalar with another one that actually does seem to help without screwing up the display. With 700C and the latest in drowe67/codec2#126:

Tab|CPU Usage with PR (%)|CPU Usage with freedv-gui master
---|---|---
Waterfall|28.0%|47.3%
Spectrum|21.4%|21.1%
Scatter|22.1%|25.2%
From Radio|34.0%|53.4%
From Mic|24.1%|28.8%
To Spkr/Hdphns|26.2%|28.5%
Timing|23.8%|19.9%
Frequency|20.9%|20.4%
Test Frame Errors|8.8%|5.7%
Test Frame Histogram|9.3%|8.7%

BTW if we want to continue with the new PlotScalar stuff I'm trying, we may need to suppress it for the Timing tab as well as that tab still renders a bit weird compared to master.

drowe67:
HI @tmiw - I just tried 6adad9f59, and get similar results to my post above for spectrum/waterfall on my machine, i.e. just the waterfall seems to be improved, and spectrum is worse.  The plot_scalar based plots (from radio, timing, freq) all run at about 50% in both master and 6adad9f59. On the Pi3B freedv-gui/master works OK for me with codec2 341fb3f.

In 6adad9f59 scatter doesn't seem to be working at all when I run a 700C sample through it.  Has that been restored to the original scatter code?  Actually even in the original code the "AGC" of scatter seems a bit slow, and it can take a few 10's of seconds to fill the screen - but lets leave that as a future issue for now.

So based on your table above you see improvements on just waterfall and from radio?  I'm reluctant to merge changes that don't consistently/significantly improve performance (unless there is some other compelling reason, such as a dramatic reduction in lines of code, or better functionality). 

How would you feel if we just include changes that have measured improvement across a few operating systems? I think that's definitely waterfall at this stage, and possibly "From Radio", although that appears to have some issue with other plots that require mode specific flags.

Re "From Radio" - we might be able to improve that outside of plot_scalar, e.g. increase the decimation rate in txRxProcessing, perhaps mixed with some averaging of the samples.  Having said that, I suspect most people spend their time on waterfall/spectrum/scatter - "From Radio" is probably rarely used and I can't recall anyone having problems due to it's CPU load.

tmiw:
@drowe67, I restored the add_sample code for scatter (forgot to earlier) but it still looks wonky on 700C. 700D seems to work better, though, not sure why.

As for the changes in general, there were some code cleanups/consolidation done as well so I would like to at least keep those. For instance, all of the OnPaint() events basically had the same code, so all of the plot classes now just use the one in the base class. Also, PlotPanel's draw() was basically never used, so it became abstract instead.

Anyway, as PlotScalar's also used by the "From Mic" tab (which FreeDV switches to on TX), there may be value in keeping those changes for that tab too; the measurements above were made in RX with full duplex disabled, meaning zero inbound audio. Some measurements during TX (or at least with full duplex turned on) might be useful.

tmiw:
Update: for "From Mic" (with 700D when I whistle into the mic while full duplex is turned on):

**This PR**: 34.3%
**master**: 41.9%

So, definitely a bit of improvement on that tab at least, FWIW.

drowe67:
> @drowe67, I restored the add_sample code for scatter (forgot to earlier) but it still looks wonky on 700C. 700D seems to work better, though, not sure why.

Thanks it's working well for me on 700C :+1:  Lol, 700D the scatter is hard to see on the test files (with no compression), 4 tiny dots.  But it's working fine.

> As for the changes in general, there were some code cleanups/consolidation done as well so I would like to at least keep those. For instance, all of the OnPaint() events basically had the same code, so all of the plot classes now just use the one in the base class. Also, PlotPanel's draw() was basically never used, so it became abstract instead.

Sure, that's a good reason to merge changes.

> Anyway, as PlotScalar's also used by the "From Mic" tab (which FreeDV switches to on TX), there may be value in keeping those changes for that tab too; the measurements above were made in RX with full duplex disabled, meaning zero inbound audio. Some measurements during TX (or at least with full duplex turned on) might be useful.

I'm happy to consider that if you think it's a significant improvement and worthwhile for end users. 

As a way forward, how about:
1. Let put the changes we are both 100% comfortable with (waterfall, plot scalar cleanup/consolidation) in a new PR.
1. In that PR I'll try some init-time and DSP sample rate ideas I have (outside of plot scalar) to address the From Radio/To Radio CPU.
1. We can then jointly decide the best way forward...

I figure this is just like that Tx PTT ending early thing, sometimes we need to run a few laps around the problem before the best solution pops up.  I am pleased you've dived into this code, I've had some lingering concerns for years about it's CPU load :+1:

tmiw:
@drowe67, at this point, backing out the caching code in PlotScalar should basically be all that's necessary to achieve point 1 (there are other changes in the Plot* classes involving switching to wxGraphicsContext that happened as a result of the code cleanup; however, as its API is substantially similar to wxAutoBufferedPaintDC I don't think there's any issue keeping those changes).

If that's okay, I can do that and the other changes you were considering can go into the separate PR.

tmiw:
OK, backed out the caching stuff from PlotScalar and added some additional code cleanup. 👍

drowe67:
Looks good to me @timw, just let me know when you're ready to merge.

Curiously, when it's in sync with 700C, freedv is <25% on all plots, but the Xorg process can get as high as 80%, so loadav is around 1.0.  I'm not sure if that's spread across a few cores.  I guess other operating systems have different graphics back ends.

tmiw:
@drowe67, I think we're good for merging this PR now. I played the entire 700C sample file with the latest commits in here and Codec2 master and got [these results](https://github.com/drowe67/freedv-gui/files/6499017/freedv_profile.txt) on my Ubuntu VM. There might be more that can be done on PlotWaterfall but since it took playing the entire sample file for PlotWaterfall to show as using more than 0.00s, it's probably going to have diminishing returns.

> I guess other operating systems have different graphics back ends.

CPU usage on macOS still seems a bit high to me relative to Linux and Windows but it might be related to [this wxWidgets/macOS SDK bug](https://trac.wxwidgets.org/ticket/18516); changing my system's color space to "Generic RGB" from "Color LCD" cut CPU usage by ~7 percentage points or so.

> I'm not sure if that's spread across a few cores.

From what I can tell with Instruments on my laptop, FreeDV is mainly using two threads: the "main" thread (for GUI updates and event handling) and the RX/TX thread. I think it's possible to at least have it use more cores for RX--mainly for simultaneous RX--but I'll experiment separate from this PR.

drowe67:
Thanks @tmiw - the `gprof` results are interesting, not sure how they can be reconciled with the results from `top`.  Re two cores, I was speculating the `freedv` and `Xorg` (X server) processes might run on separate cores, rather than freedv itself, which is just multi-threaded. 

When in sync on 700C, the CPU for `Xorg` usually exceeds that of `freedv` (which is 20-25% for all centre panels).  I guess that's the graphics subsystem doing all the drawing

PR #129, opened 2021-05-21
Multi-threaded multiple RX support

tmiw:
Converts the single-threaded receive handler in FreeDVInterface into a multithreaded one (one thread per listening mode). The intention is to allow some machines that were borderline for this feature in the past (e.g. older dual or quad core Intel machines or Raspberry Pis) to better handle the feature, as well as achieve some improvements in sync performance and general thermals/power consumption.


tmiw:
@drowe67, unfortunately it looks like we still can't include 800XA and 2400B even with the latest Codec2 changes. At least for 2400B, it seems to always be in "sync", meaning that the other threads have little ability to inject audio into the playback stream. 

OTOH, I'm getting about 40% of a core's worth of processor usage with all modes enabled, so this PR and/or the Codec2 changes did help. 👍

src/freedv_interface.cpp
```diff
@@ -447,84 +459,184 @@ int FreeDVInterface::processRxAudio(
...
+                futIndexWithSync = futIndex;
+                break;
             }
             
-            nin = freedv_nin(dv);
+            lastSyncRxMode_ = nullptr;
         }
         
-        if (done)
+        if (snr > maxSyncFound && tmpStats.sync != 0)
```
tmiw, line 583: @drowe67, using both sync and highest SNR _should_ be okay if 800XA and 2400B aren't included in multi-RX. Based on my understanding, if there are somehow two or more modes that are in sync (due to blips), the highest SNR should be the "correct" one. This could be a problem for signals that are within the noise but there's also have code farther up to not select any other modes if we're already in sync on one of them.

Anyway, let me know if my understanding is incorrect or if there are additional pitfalls I didn't think of. 👍

drowe67: Its a bit hard to say, as the SNR is filtered/averaged so may rise slower in some modes than others.   However if a mistake is made it will fall out of sync eventually and you'll get another chance.

drowe67:
@tmiw pls let me know when you're happy with this PR.  I'm afraid the code is beyond my current skills so I'm not sure I can effectively review it :thinking:

tmiw:
@drowe67, it seems okay here so far but this is the kind of change we'd probably want to do test builds for. Maybe we can get the other pending PRs in first and then generate some?

tmiw:
@drowe67, I added an additional option in Tools->Options (if multiple RX is enabled) to return FreeDV to the pre-PR behavior. Once you're done on the codec2 side I'll generate builds based on this PR so there's runtime on this code at least.

drowe67:
Hi @tmiw I'm happy with codec2/master, so pls feel free to generate test builds of freedv-gui any time you like.

tmiw:
@drowe67, sounds good. It might be a bit because I ran into some MacPorts issues here but I'll get them out ASAP. 👍

drowe67:
@tmiw - suggest asking end users to measure CPU load with this feature on/off, to measure the benefits of the multi-threaded approach.

tmiw:
@drowe67, yep, I asked in the latest email I sent out. 👍

tmiw:
@drowe67, I haven't heard anything indicating that the most recent build caused another issue related to this. However, #146, #145 and #142 will need to merge first before this one.

drowe67:
@tmiw thanks, I agree there have been no recent issues reported with this PR.

However it does add complexity, and as I mentioned above I would struggle to maintain the multithreaded code in this PR if you were unavailable. So I want to make sure there is good reasons for merging it, i.e. some benefit or a bug fix.

Some questions:

1. Can we obtain the same functionality and performance without a multithreaded approach (ie the original single threaded code)?
1. Do you have any measurements that indicate the goals of this PR at the top were reached?
1. Failing that are we sure it doesn't make operation slower? 
1. Can the option to disable please be retained, such that if there is an issue (and you are unavailable to fix it), we can revert to the previous, single threaded operation?

tmiw:
> @tmiw thanks, I agree there have been no recent issues reported with this PR.
> 
> However it does add complexity, and as I mentioned above I would struggle to maintain the multithreaded code in this PR if you were unavailable. So I want to make sure there is good reasons for merging it, i.e. some benefit or a bug fix.

Once #142 merges we can see if this PR shows a lot fewer file changes. It may very well look more complicated than it is given that I've been merging the other PRs into this one in order to generate test builds.

> Some questions:
> 
> 1. Can we obtain the same functionality and performance without a multithreaded approach (ie the original single threaded code)?
> 2. Do you have any measurements that indicate the goals of this PR at the top were reached?
> 3. Failing that are we sure it doesn't make operation slower?

I should still have a Raspberry Pi 3 here that may be slow enough such that the multi-threaded version makes a difference (but not too slow such that it can't run FreeDV at all). I definitely have a rPi4 that does run FreeDV, albeit spiking the thermals after a bit. I'll see if I can get answers for you for those questions.

> 4. Can the option to disable please be retained, such that if there is an issue (and you are unavailable to fix it), we can revert to the previous, single threaded operation?

Yep, I have no intention on removing that setting (at least for now).

tmiw:
@drowe67, FWIW, Brian G8SEZ just wrote the following to the digitalvoice list:

> On a different note I tried the current freedv-gui master version and the ms-optimization build, the latter is much faster to start up and enable the modem so it certainly gets my vote for being merged into master, it claims to be 40 commits ahead I think.

drowe67:
> @drowe67, FWIW, Brian G8SEZ just wrote the following to the digitalvoice list:
> 
> > On a different note I tried the current freedv-gui master version and the ms-optimization build, the latter is much faster to start up and enable the modem so it certainly gets my vote for being merged into master, it claims to be 40 commits ahead I think.

Thanks @tmiw yes I saw that.  Not sure how this version could be faster to start up?  Maybe I'm misunderstanding the comment :thinking: . It would be good to show any improvements in the form of a repeatable test, and have some numbers around it.

tmiw:
I got some CPU usage numbers per thread on my Raspberry Pi 4 tonight (freedv-gui a few seconds after it finishes processing the Start button press but no audio going in):

*ms-optimization*:

IMG_20210722_231403

*master*:

IMG_20210722_231801

CPU usage seems to be similar between the two, just across more threads with the former. However, the former also seemed to take a bit longer to start after pushing the Start button. It'd be nice to get confirmation from more than one system in case it's something specific to my Pi.

drowe67:
OK nice work @tmiw - the start time seems an odd one, I would have expected the same time.  Might be a coincidence.  I've been reading a bit about how threads spread across cores.  I suspect (but am not sure) that more modern processors have better hardware to support this.

I'm happy to merge it seeing you have (4) in place.

tmiw:
Another test of multithreaded mode, this time on a VM running on my home NAS:

Screen Shot 2021-07-24 at 3 00 56 AM

Anyway, this is ready for merge.

PR #130, opened 2021-05-22
2400B Analog passthrough

drowe67:
WIP to debug 2400B and 800XA issues, in conjunction with https://github.com/drowe67/codec2/pull/188

1. Analog pass through working in 2400B.
2. Refactored out some unneeded globals.
3. Updated some comments.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -149,10 +149,6 @@ wxWindow           *g_parent;
 float               g_RxFreqOffsetHz;
 float               g_TxFreqOffsetHz;
 
-// buffer sizes dependent upon sample rate
-int                 g_modemInbufferSize;
-int                 g_speechOutbufferSize;
-
 // experimental mutex to make sound card callbacks mutually exclusive
 // TODO: review code and see if we need this any more, as fifos should
 // now be thread safe
@@ -1562,11 +1558,6 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
             g_error_histn[i] = 0;
         }
 
-        // Set processing buffer sizes, these are FRAME_DURATION (20ms) chunks of modem and speech that are a useful size for the
-        // various operations we do before and after passing to the freedv_api layer.
-        g_modemInbufferSize = (int)(FRAME_DURATION * freedvInterface.getRxModemSampleRate());
-        g_speechOutbufferSize = (int)(FRAME_DURATION * freedvInterface.getRxSpeechSampleRate());
-
         // init Codec 2 LPC Post Filter (FreeDV 1600)
         freedvInterface.setLpcPostFilter(
                                        wxGetApp().m_codec2LPCPostFilterEnable,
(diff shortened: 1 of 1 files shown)


drowe67:
@tmiw when convenient (no rush) could you pls try this PR together with https://github.com/drowe67/codec2/pull/188?

I've made some progress on the issues in https://github.com/drowe67/codec2/pull/188, latest status [here](https://github.com/drowe67/codec2/pull/188#issuecomment-845754530).  I've tested several uses cases, but still a bit worried I've missed something ....

tmiw:
@drowe67, with this branch and the corresponding codec2 one:

1. Push Start with 2400B selected while on 20 meters. SNR is 0 as expected.
2. Play 2400B sample file. The sample file begins playing but SNR is still 0 and there's no audio played back regardless of squelch level.
3. Disable squelch -- audio plays back normally (SNR still 0).
4. Push Stop and then Start again without manually stopping playback--audio begins playing and now has an SNR around 40+. Audio continues playing with squelch enabled, too.
5. Stop playback of sound file. SNR drops down to 4-5 or so.

Hope that helps!

drowe67:
Thanks @tmiw - I managed to repeat your test (after some experimentation).  Can you pls try again with codec2 58ccc973 ?

A SNR of 4-5dB with no signal is actually expected for this demod, at least at the moment. This demod is unusual, in that it's design to work on the baseband output of an analog FM demodulator (a bit like VARA-FM and other AFSK modems used with FM radios on VHF).  As such the mapping between the SNR estimator of this demod and the SNR at the input of the FM demodulator is complex, and I haven't figured it out yet.  

All the other modes use SSB, which is much more straight fwd - a SSB radio just shifts chunks of spectrum up and down from RF to baseband.

tmiw:
@drowe67, looks like that scenario works with 58ccc973. The same scenario seems to work fine with 800XA too.

Question about analog passthrough and 2400B--I hear a lot of artifacts with squelch off because it seems to very briefly sync pretty often (with 20m band noise). Normal?

drowe67:
> @drowe67, looks like that scenario works with 58ccc973. The same scenario seems to work fine with 800XA too.

Yayyy!  Phew - that took a few goes.  Thanks for your testing, helped me catch a few corner cases.  Keep an eye on the other modes too, as I refactored some non-mode specific generic code too.  I'll merge that PR soon.

> Question about analog passthrough and 2400B--I hear a lot of artifacts with squelch off because it seems to very briefly sync pretty often (with 20m band noise). Normal?

Yes, and also with 800XA.  The Unique Words on those modes are quite short (800XA is just 8 bits), so the probability of a false sync with random noise is quite high.  Those modes are rarely used, so given the number of false syncs it might not be worth including them in the multi-rx group.

tmiw:
@drowe67, agreed on multi-RX, especially since another PR is proposing to use highest SNR to determine the mode to use (vs. just sync). That may need more testing to be sure we don't run into any problems even without including 800XA/2400B.

PR #131, opened 2021-05-23
Additional optimizations for PlotWaterfall

tmiw:
Additional optimizations for PlotWaterfall:

1. macOS: Force main window's color space to be the same as wxWidgets (sRGB). This prevents background color space conversions and reduces CPU usage. Based on info and code from https://github.com/OpenTTD/OpenTTD/issues/7644 and https://trac.wxwidgets.org/ticket/18516.
2. Cache each displayed waterfall segment using wxGraphicsBitmap instead of re-rendering the entire visible area as a single bitmap.

These changes result in FreeDV CPU usage dropping to ~15% on both macOS and Linux when the waterfall's displayed (2019 MacBook Pro, latter using a VM on said machine).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -855,7 +855,7 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
         m_panelWaterfall->setRxFreq(FDMDV_FCENTRE - g_RxFreqOffsetHz);
         m_panelWaterfall->m_newdata = true;
         m_panelWaterfall->setColor(wxGetApp().m_waterfallColor);
-        m_panelWaterfall->Refresh();
+        m_panelWaterfall->RefreshWaterfallOnly();
     }
 
     m_panelSpectrum->setRxFreq(FDMDV_FCENTRE - g_RxFreqOffsetHz);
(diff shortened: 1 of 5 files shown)


drowe67:
Hi @tmiw - CPU loads measured in my Ubuntu machine 700D, decoding `ve9qrp_700d.wav`:

| hash | freedv | Xorg |
| --- | --- | --- |
| freedv-gui/master d28681 | 33 | 25 |
| b212784 |  38 | 33 |

tmiw:
Just tried again. 25% with master for me, 17% with this PR while decoding the sample 700D file (macOS). This is with multi-RX disabled and once the waterfall fully fills up.

Think it would be better to revert all but the macOS-specific changes? Those should be able to be included without needing to include the others.

drowe67:
> Think it would be better to revert all but the macOS-specific changes? Those should be able to be included without needing to include the others.

Yes, it would be good not to slow down operation on any other operating systems.  Might be useful to check Windows too.

tmiw:
@drowe67, on further testing the non-colorspace changes didn't seem to make much of a difference on macOS either (except right on app start/resize when the waterfall isn't full yet). I reverted all but the macOS specific changes as a result.

drowe67:
Hi @tmiw - is this PR ready for merge?

tmiw:
@drowe67, correct, this is ready. 👍

PR #134, opened 2021-05-26
macOS: Remove dependency on building our own PortAudio

tmiw:
Looks like MacPorts' PortAudio now has a universal binary for both ARM and x86, so no need to build our own on macOS anymore. The patch file was still updated, though, in case some prefer to do so (and eventually will probably be removed).


drowe67:
Hi @tmiw - I can't really test this one, but I'm happy to merge it, if you're finished.

tmiw:
@drowe67, yep, I'm good. 👍

Issue #133, opened 2021-05-24
undefined layouts when re-sizing main window on Ubuntu Linux


drowe67:
While testing 304de7875 I noticed an odd screen layout:

![Screenshot from 2021-05-25 07-19-44](https://user-images.githubusercontent.com/45574645/119411918-55c23880-bd2a-11eb-95fe-57709155de80.png)

It can also be reduced in size right down to zero.  Quite a few GTK warnings being issued in `stderr`.

@tmiw - 607b827 seems OK, so this issue might have appeared with #121?

tmiw:
@drowe67, oh yeah, I switched the sizers used by the left and right panels to allow them to take horizontal space when the display resolution isn't tall enough to allow the "normal" UI. The default window size for the main window should allow it to still look the same otherwise, though.

Anyway, maybe some sort of minimum size should still be enforced and not allow resizing down to zero?

drowe67:
Yes it seemed to make a good choice of moving some buttons to the RHS with a "wide" window.  However after re-sizing a few times I found many "illegal" arrangements, for example the one above has a zero size center panel, and reducing the entire Window down to zero size. 

I guess this could be handled by setting a minimum for the outside window, or perhaps a minimum for other GUI elements.

tmiw:
Apparently the GTK warnings are "normal":

http://wxwidgets.10942.n7.nabble.com/GTK-warning-and-critical-error-overkill-td95569.html
https://stackoverflow.com/questions/61486307/supressing-warnings-in-production-for-wxwidgets-application

(among other sources)

The justification seems to be that a) they don't seem to impact execution and b) users don't typically start GUI apps from the terminal. If there aren't any issues with FreeDV other than the warnings on the console, we probably don't need to suppress them, especially since I haven't been able to get a minimum size to stick for the various plots (nor does the minimum size behavior for the main window do 100% of what we'd like it to do).

drowe67:
Yes I agree it's just a minor annoyance to have the warnings, a bit like ALSA.  I guess we could filter them using pipes and grep if we really needed to, e.g. if they were obscuring other warnings.

As a fall back, the previous behaivour like https://github.com/drowe67/freedv-gui/commit/607b827ef5d8fc789f500c0b9ebe5aa7300c76ce is OK with me.

tmiw:
I'd say leave as is _unless_ there are bugs as a result of the controls moving horizontally. If that's the case, we can revert back to 607b827's behavior. 👍

[...]

PR #135, opened 2021-05-28
Restrict resize to a minimum size that won't cause GTK warnings.

tmiw:
Per #133, here are some tweaks to ensure that the window can't be resized too small (and causing GTK warnings among other isuses).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -402,7 +402,7 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent)
     // Add to the "best" width and height to prevent hiding of the waterfall
     // and the controls just below it. The values here were determined by 
     // experimentation.
-    size.SetWidth(size.GetWidth() + 375);
+    size.SetWidth(size.GetWidth() + 350);
     size.SetHeight(size.GetHeight() + 100);
 
     if (w < size.GetWidth()) w = size.GetWidth();
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -321,9 +321,6 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent)
     g_SquelchLevel = pConfig->Read(wxT("/Audio/SquelchLevel"), (int)(SQ_DEFAULT_SNR*2));
     g_SquelchLevel /= 2.0;
 
-    Move(x, y);
-    SetClientSize(w, h);
-
     if(wxGetApp().m_show_wf)
     {
         // Add Waterfall Plot window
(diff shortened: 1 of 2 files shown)


tmiw:
@drowe67, BTW, here's the smallest I can resize the main window with this PR:

Screen Shot 2021-05-27 at 7 40 53 PM

drowe67:
@timw - that's great and works really well, and resolves all of my concerns in https://github.com/drowe67/freedv-gui/issues/133 :grinning: .  Any time I tried to resize - the layout was fine (even after Start was pushed) :+1: 

I even tried the "tiling trick" and it worked Ok:

Screenshot from 2021-05-28 14-57-30

Some super small things that you don't have to tackle in this PR, but I wanted to mention:
1. Occasionally, when I drag the spectrum plot to the top of the centre panel, freedv-gui bombs, I've noticed this occasionally for the last few weeks.  But I can't figure out how to repeat it reliably.
1. When the window is re-sized - the waterfall often gets filled with "noise" - probably some un-initialised screen buffer memory.  Example in the waterfall above.
1. I was trying to figure out how to get the stats numbers  to use a justification other than LH, but I think it's a proportional font, so hard to say RH align them.

tmiw:
Just pushed a change to PR #131 that should help with point 2 above. Center justifying the stats numbers shouldn't be a big deal, I'll work on that in this PR now.

drowe67:
Thanks :slightly_smiling_face: Actually I'm not sure what justification would be best, ie center or right, or perhaps left but with the digits in each field aligned to a new column in the center of the stats window.  Here's one possibility:
```
Bits:         6390
Errs:            8
BER:         0.001
Resyncs:         8
```
or maybe:
```
Bits:        6390
Errs:        8
BER:         0.001
Resyncs:     8
```
:thinking: not sure if that's 100% right either.  Maybe there is a convention for this?

tmiw:
Right now the stats display is one static text field per stat. We'd probably need to split each into two (one for the label, one for the value) to accomplish either of those options. For now, I center aligned each of the existing static text fields, but perhaps another issue/PR would be best for this one.

drowe67:
Hi @tmiw - I reverted the center alignment of the Stats text as I felt it didn't really address the (minor) core issue, as discussed above in https://github.com/drowe67/freedv-gui/pull/135#issuecomment-850263755.

Were there any other changes you wanted to make in this PR, or are we OK to merge?

tmiw:
@drowe67, this is OK to merge.

Issue #51, opened 2020-03-24
Feature Request: PTT Input


drowe67:
Sigi dg9bfc writes:

in freedv soft you have ptt out on a comport (to key external rig) ... so far so good

can you add also a PTT IN via CTS on a comport??

reason is ... i now use freedv on qo100 with sdr console ... and in normal ssb i key the sdr radio with usb/serial converter (ftdi chip) with a pullup resistor and pulling the cts pin to ground

oh3gai:
First of all  thanks for great application.

Serial PTT in would be great improvment and / or wild idea to support for "zello" bt mic which would make IE. mobile operation simple. Audio in /out and ptt in same package.  

https://rover.ebay.com/rover/0/0/0?mpre=https%3A%2F%2Fwww.ebay.com%2Fulk%2Fitm%2F164035444072

Good work !!
Br,Jarno / OH3GAI

DG9BFC:
Hello David
any news about adding the PTT IN ?
a similar setup like we have for ptt out 
select comport and select pin to be used

drowe67:
@DG9BFC this feature would involve significant effort to implement.  We would need to be arbitrate between the same serial port being used for PTT control.  Our experience has been anything around serial ports involves many support issues.

So some questions:
1. How any other people want this feature? Pls show this is not a one-off feature.  Wide spread email support on the digital voice list would be useful.
1. Is there any other way to do it with 3rd party software?  Is anyone else using a foot switch or external PTT control with FreeDV already?

drowe67:
Here is one other request: https://github.com/drowe67/freedv-gui/issues/52

DG9BFC:
There is no way for a foot pedal(YET!) You need to use the mouse... Should be a comport in on cts pinA usb to serial converter like ftdi or profilic and you can use any desktop mike with a ptt switch with pulling cts to ground So no more need of mouse switchI have no idea how many would use it.. But i guess a lot users would love to have a normal ptt button and do not need to use mouse to press ptt on screenA real ham radio mike can be wired that way to the PC.....similar as what i do with my sdr setup a normal user with a rig can wire gis handheld or desktop mike to pc... Ptt in on one comport... Ptt out on another port?? Or wire pc and rig in parallel?? So... I think a ptt in is a good ideaGreetz dg9bfc sigiAm 06.01.2021 23:23 schrieb drowe67 <notifications@github.com>:
[...]

Issue #29, opened 2019-10-12
automatic detection of serial port permissions issues on Linux


drowe67:
A common trap for PTT using a serial port (HamLib or Serial PTT) is lack of permissions. This fix is usually adding your user to the `dialout` group. It would be nice to automatically detect permissions issues  and report by a pop up window during serial port configuration/testing.

tmiw:
#136 updates Linux serial port detection to use a combination of globbing /sys and stat() on /dev/tty* to determine whether the port is valid. Since stat() returns -1 for ports that aren't usable, is the port not appearing in the port list sufficient?

drowe67:
I think it's useful to know what physical ports are actually available :+1: BTW there are many variations in Linux serial port names, does #136 capture them all in the list?

:thinking: However I think the port not appearing (when a user knows it's there and connected to the PTT device) might be confusing to a user.  Like the current situation, it would require a note in the documentation to explain, so is a sideways step/variation on this issue, rather than a solution.

The behavior above in OP above would be ideal - let the computer take the user straight to the problem.

tmiw:
I'm not too familiar with serial ports in Linux but my understanding is that /dev/ttyS* and /dev/ttyUSB* should cover the vast majority (if not all) of the common ones. Those are definitely covered by /sys, IIRC.

Anyway, I tweaked the OK/Apply actions to test chosen ports and display a message: https://github.com/drowe67/freedv-gui/pull/136/commits/113c366dffb70640d10c5c99e5933426c9255d2e. The message right now is "Couldn't open serial port ___. This is likely due to not having permission to access the chosen port." but I'm open to suggestions.

drowe67:
Hi @timw, thanks for working on this.  I took a look at https://github.com/drowe67/freedv-gui/commit/113c366dffb70640d10c5c99e5933426c9255d2e:
1. The error message is presented on all operating systems, whereas this issue deals with a specific Linux gotcha.  Hmm :thinking: , could it also occur on other operating systems?  I don't know enough about them myself....
1. It doesn't handle the permissions issue when encountered using a serial port with HamLib (I'll make that clearer in the problem description above).
1. The error message that's been added could be misleading - as it doesn't specifically detect the permissions issue as described in OP, so it will be displayed for many other reasons.

[...]

PR #136, opened 2021-05-31
Add support for PTT input

tmiw:
Adding support per #51. Only supports CTS pin monitoring using a separate port at this time.


src/dlg_ptt.cpp
```diff
@@ -116,24 +116,10 @@ ComPortsDlg::ComPortsDlg(wxWindow* parent, wxWindowID id, const wxString& title,
     wxStaticBoxSizer* staticBoxSizer17 = new wxStaticBoxSizer( serialSettingsBox, wxVERTICAL);
     mainSizer->Add(staticBoxSizer17, 1, wxEXPAND, 5);
     
-    wxStaticBox* pttBox = new wxStaticBox(serialSettingsBox, wxID_ANY, _("PTT Port"));
+    wxStaticBox* pttBox = new wxStaticBox(serialSettingsBox, wxID_ANY, _("PTT Out"));
```
drowe67, line 119: Can we pls use the nomenclature for PTT/PTT In, rather than moving to a PTT In/PTT Out?  I don't want the end users have to think about this step (am I configuring PTT In or Out?), as it will lead to confusion.

I'd like to emphasis that the PTT in feature is very much a low demand, one off feature, and should have minimal impact on the current PTT feature which is very important.  I want to make sure the tail doesn't wag the dog.

tmiw: Reverted back to "PTT Port".

drowe67:
@tmiw re https://github.com/drowe67/freedv-gui/issues/29#issuecomment-851897918 - I've also seen other serial port names like `/dev/ttyACM0`, but you've used a combo box, so we should be able to type in anything unexpected.  When I set up serial I use a console and `dmesg`, to find out what serial port I'm using.

Minor thing but on my machine the combo boxes for the Serial Device seem a little large (e.g. larger than the Serial Device combo box in the hamlib section):

Screenshot from 2021-06-01 20-37-43

tmiw:
@drowe67, yeah, the combo boxes seemed pretty small for me before (and were cutting off a fair bit of the text). The aim was to allow the field to resize when the dialog box gets resized.

Also, a larger field in the Hamlib section is useful for those radios that use TCP/IP (e.g. the Hermes Lite 2 with piHPSDR).

tmiw:
@DG9BFC, @oh3gal: have either of you had a chance to try this PR? How well is it working?

DG9BFC:
yes it does work ... but i have to start freedv (and serial port splitter) before sdr softif i want to use it with the sdr and with an external rig (usage on qo100 with a pluto and on shortwave with a "normal" rig) i need a different setup
... but yes the ptt in DOES WORK FINE
means now you can connect any "normal" ham mike to your pc (or add a footswitch or  whatever to freedv soft)
well done ... superb job adding the cts line as input
with a serial to usb converter (ftdi dongle) now you can switch freedv to tx ... GOTCHA... it works fine

DG9BFC:
when port is already in use it gives an error (can not the comport be set to be "shared"?!?)
i check that ... should then give no error but can be read from two aps in parallel
greetz sigi dg9bfc

DG9BFC:
ok ... sharing a comport does not work in windoze (grmbl)
so virtual serial port emulator (with a splitting port) needs to be used (but thats a special setup for using with an sdr)
in freedv i then can either use the shared port towards sdr (or vox) ... or use another virtual port and a cat command (my sdr soft talks (ts2000)
same on external rig (ft990) i can either use cat command or rts on com1 (switching transistor) ...
so ... that serial port in added opens up "a can of worms" hi hi
... com 7 (ftdi in) as ptt in ... and com1 with rts as ptt out to switch external rig is what most users will use and that works superb
that a "used port" can not be found by the soft is just a normal thing (and can be solved with virtual port)
greetz sigi dg9bfc

tmiw:
@DG9BFC, glad to hear that it's working, even if there are some Windows quirks.

@drowe67, perhaps we can proceed with this PR and make sure the code changes don't interfere with the existing Hamlib/serial PTT support. If other bugs in PTT input end up being encountered later, those can always be fixed.

drowe67:
@tmiw OK cool.  Is there anything you would like to add to the manual before merging?

tmiw:
@drowe67, yep, I just pushed an additional paragraph covering PTT support in the manual. 👍

DG9BFC:
i guess we can close the issue :-)
again a BIG THANK YOU!!!

PR #137, opened 2021-05-31
Resolve waterfall flickering issue on Linux.

tmiw:
Per discussion on the digitalvoice list, I was able to duplicate the flickering issue by repeatedly clicking on the Ubuntu desktop while FreeDV was running. This PR should resolve that issue.


tmiw:
@drowe67, looks like this PR resolves the issue posted on digitalvoice, so I'm ready for merge.

PR #138, opened 2021-05-31
Fixed additional issues with macOS and MacPorts/Homebrew PortAudio

tmiw:
Turns out I spoke too soon and there were still issues with not using the preexisting PortAudio. This should now resolve that but it would still be nice to make sure Windows and Linux builds still work. 👍


tmiw:
@drowe67, this is okay to merge too.

PR #140, opened 2021-06-19
Remove unneeded checks to reflect codec2 PR #194.

tmiw:
As the subject says. Since codec2 is the one doing the checks before applying the equalizer setting, freedv-gui doesn't need to do them too. The other stuff in FreeDVInterface also doesn't do checks for the most part (except for setCarrierAmplitude() as Codec2 currently asserts on mode == 700C).

### DIFF
--- a/src/freedv_interface.cpp
+++ b/src/freedv_interface.cpp
@@ -307,15 +307,9 @@ int FreeDVInterface::getSync() const
 
 void FreeDVInterface::setEq(int val)
 {
-    int index = 0;
     for (auto& dv : dvObjects_)
     {
-        int mode = enabledModes_[index];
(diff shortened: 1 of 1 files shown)


tmiw:
@drowe67, should the assertion be removed from freedv_set_carrier_ampl() too?

drowe67:
Good catch @tmiw :smile: 

> @drowe67, should the assertion be removed from freedv_set_carrier_ampl() too?

Yes please, an assert is perhaps a bit heavy handed!  Medium term I'd like to rm some of those test options (e.g. error histogram).  I used them for some initial 700C development work many years ago but they don't have any practical use.  Removing them would simplify the source code and GUI and manual  :thinking:

tmiw:
@drowe67, done. 👍

PR #141, opened 2021-06-29
Var update for all modes using Codec 2 700C

drowe67:
freedv-gui side of https://github.com/drowe67/codec2/pull/196

I tested all modes using full duplex as per https://github.com/drowe67/codec2/pull/194 - Var was updates in FreeDV 700C/D/E and 800XA, and not updated in the other modes (which still worked OK).


tmiw:
@drowe67, looks good here. 👍

Issue #144, opened 2021-07-09
FreeDV crashes when using Bluetooth headsets


tmiw:
Tried using one the other day and ran into a lot of random crashes immediately on pushing Start.

Example (from macOS 11.4):

```
Process:               freedv [95781]
Path:                  /Users/USER/*/freedv
Identifier:            org.freedv.freedv
Version:               1.0 (1)
Code Type:             X86-64 (Native)
Parent Process:        bash [21514]
Responsible:           Terminal [433]
User ID:               501

Date/Time:             2021-07-07 16:03:09.475 -0700
OS Version:            macOS 11.4 (20F71)
Report Version:        12
Bridge OS Version:     5.4 (18P4663)
Anonymous UUID:        A45FADFF-EF47-4050-91B2-5ECE3531D6ED

Sleep/Wake UUID:       CC9B3AC5-0DB5-4F9F-99B6-FE650560CB11

Time Awake Since Boot: 570000 seconds
Time Since Wake:       5100 seconds

System Integrity Protection: enabled

Crashed Thread:        1  com.apple.NSEventThread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
Performing @selector(controlAction:) from sender wxNSButton 0x7fc962599630
abort() called
freedv(95781,0x700005ab2000) malloc: Incorrect checksum for freed object 0x7fad03afb648: probably modified after being freed.
Corrupt value: 0x0

...

Thread 1 Crashed:: com.apple.NSEventThread
0   libsystem_kernel.dylib        	0x00007fff2035a92e __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff203895bd pthread_kill + 263
2   libsystem_c.dylib             	0x00007fff202de411 abort + 120
3   libsystem_malloc.dylib        	0x00007fff201be1f5 malloc_vreport + 548
4   libsystem_malloc.dylib        	0x00007fff201d2809 malloc_zone_error + 183
5   libsystem_malloc.dylib        	0x00007fff201b34a4 tiny_malloc_from_free_list + 1797
6   libsystem_malloc.dylib        	0x00007fff201b281d tiny_malloc_should_clear + 233
7   libsystem_malloc.dylib        	0x00007fff201b1727 szone_malloc_should_clear + 66
[...]

PR #145, opened 2021-07-11
Convert samples from voice keyer file to sound card 2's sample rate.

### LINKED ISSUE #8: voice keyer wave file format
Voice keyer will use wave file even if it's format (8 or 16 kHz) is not correct
1. Trap if wave file format not correct, and notify user.
1. Note a different wave file is required for 2020 (16kHz) and other modes (8kHz)

tmiw:
When sound card 2's sample rate is less than sound card 1's, FreeDV terminates transmit too early when using the voice keyer even if the .wav file is in the mode's expected sample rate (e.g. 8 or 16KHz). This change ensures that the samples from the voice keyer match the sample rate that has been configured by the user prior to being passed to Codec2.

This resolves #8 and potentially also #144.


tmiw:
@drowe67, see below for how to reproduce without this PR:

Scenario 1:

1. Set mic/speaker sound device to 16KHz (e.g. Bluetooth devices on macOS, but device doesn't have to be BT).
2. Set radio sound device to something higher (e.g. 44100 or 48K).
3. Set voice keyer to an 8K .wav file and push the Voice Keyer button after pushing Start. FreeDV transmits for ~1 second or less and then stops, regardless of the length of the file.

Scenario 2:

1. Set voice keyer to a 48K .wav file.
2. Set all four devices to the same sample rate (e.g. 44100 or 48K).
3. Push the Voice Keyer button after pushing Start. Voice keyer takes significantly longer to finish transmitting and is garbled on the other end.

tmiw:

drowe67:
Hi @tmiw I'm trying to test this one but can't find a sound card that runs at 16 kHz!

I like the idea of having a resampler for the mic signal in file to address #8 :+1:

src/main.cpp
```diff
@@ -2617,7 +2625,9 @@ void txRxProcessing()
 
         int nsam_in_48 = g_soundCard2SampleRate * freedvInterface.getTxNumSpeechSamples()/freedvInterface.getTxSpeechSampleRate();
         assert(nsam_in_48 < 10*N48);
-        while((unsigned)codec2_fifo_free(cbData->outfifo1) >= nsam_one_modem_frame) {
+        
+        int bytesNeededForWrite = nsam_one_modem_frame * (g_soundCard1SampleRate/g_soundCard2SampleRate);
```
drowe67, line 2629: I'm not sure about this - think it should still be:
```
while((unsigned)codec2_fifo_free(cbData->outfifo1) >= nsam_one_modem_frame) {
```

Outfifo1 is containing frames of modulated modem signals at the g_soundCard2SampleRate sample rate, so I figure we should run around this loop while there is room for one more output modem frame at that rate. 

For example if the modem sample rate is 8000 Hz, the sound card 48000 Hz, and there is 1288 samples/modem frame at 8000 Hz, then `nsam_one_modem_frame` = 48000*1288/8000, the number of samples in one modem frame, at the output sample rate. 

Sorry of I've missed something here ... this code is pretty tricky :thinking:

src/main.cpp
```diff
@@ -2906,7 +2922,7 @@ int MainFrame::txCallback(
 
     assert(framesPerBuffer < MAX_FPB);
 
-    if (rptr && !endingTx) {
+    if (rptr && !endingTx && !(g_playFileToMicIn && (g_sfPlayFile != NULL))) {
```
drowe67, line 2925: Can we pls use a design that avoids extra conditionals?  I'm feel this (and the extra fifo write step) is more complicated than the original design of the mic-in system.

The original design was:

1. Read infifo to buf
1. Optionally replace buf samples with mic-in samples from file
1. (Add new resampler here)
1. Speex noise processor

Is there any reason why the original design wouldn't work?

src/main.cpp
tmiw: I played with it some more and I think you're right. However, `nsam_one_modem_frame` should use `g_soundCard1SampleRate`--not `g_soundCard2SampleRate`--based on the comments already in the code. In fact, testing it this way seemed to work (16K sample rate for the microphone, 48K for the radio device). I'll update accordingly.

src/main.cpp
tmiw: I think this was something I was needing to do to avoid injecting the mic audio in place of the voice keyer audio. Moving some code around in txRxProcessing() avoids the need for this, though, so I'll remove.

tmiw:
> Hi @tmiw I'm trying to test this one but can't find a sound card that runs at 16 kHz!
> 
> I like the idea of having a resampler for the mic signal in file to address #8 👍

Any sound card works to duplicate the issue, actually. It just needs to be set to a lower sample rate in the audio settings and it will behave as though it were a e.g. 16K sound card.

src/main.cpp
drowe67: Good catch :+1:

src/main.cpp
```diff
@@ -2636,22 +2645,28 @@ void txRxProcessing()
...
                     if (g_loopPlayFileToMicIn)
                         sf_seek(g_sfPlayFile, 0, SEEK_SET);
                     else {
-                        // call stop/start play menu item, should be thread safe
+                        printf("playFileFromRadio finished, issuing event!\n");
                         g_parent->CallAfter(&MainFrame::StopPlayFileToMicIn);
                     }
                 }
+                
+                codec2_fifo_write(cbData->infifo2, insound_card, nout);
```
drowe67, line 2665: Should we be doing a `codec2_fifo_write` here?

src/main.cpp
```diff
@@ -2636,22 +2645,28 @@ void txRxProcessing()
...
             }
-
+            
+            nout = resample(cbData->insrc2, infreedv, insound_card, freedvInterface.getTxSpeechSampleRate(), g_soundCard2SampleRate, 10*N48, nsam_in_48);
+
```
drowe67, line 2669: Guess this line could have stayed above the mic file processing (in the original position), and we could have resampled the mic file signal to `freedvInterface.getTxSpeechSampleRate()`.  But I don't think it makes much difference - this is fine :slightly_smiling_face:

drowe67:
Looks much cleaner now @tmiw - thanks :+1:

src/main.cpp
tmiw: Nope, will remove.

drowe67:
Hi @tmiw, so I connected two machines on the bench over an audio cable and tested Scenario 1 and Scenario 2 using 700C, both worked well :+1: I tried the Vocie Keyer, and the `Play File from Mic In` features.

So if you're happy, I figure this one is ready for merge. Seeing you have a test build out in the wild, lets give it a few days to see if any feedback pops up from end users, then we can merge this and the other two PRs.

tmiw:
@drowe67, agreed. I saw another crash in wxWidgets today but I'm going to guess there's something in PortAudio (possibly combined with macOS/Bluetooth) causing it and not FreeDV or this PR, so that may need more investigation later.

drowe67:
I have found some sound cards "just won't work" - so it may be a one off.  A few years ago I went to some length to look into a problem with a sound card an end user was reporting and it turned out to be buggy hardware - the $5 USB sound card was actually dropping samples.

tmiw:

drowe67:
@tmiw I've just noticed a conflict in this PR and https://github.com/drowe67/freedv-gui/pull/142, not sure if they appeared after I merged https://github.com/drowe67/freedv-gui/pull/146

tmiw:
@drowe67, I just resolved the conflict. Once the build checks finish we should be good.

PR #146, opened 2021-07-13
Reset scatter plot state when multi-RX switches modes

tmiw:
As reported on the digitalvoice list, 700C scatter plots have odd behavior when multi-RX is on:

Screenshot from 2021-07-12 19-18-39

On investigation, it was discovered that state needed by PlotScatter has to be reset whenever FreeDV starts receiving a new mode. This PR implements that functionality.

(This is not part of ms-optimization due to the significant code changes this requires.)

### DIFF
--- a/src/freedv_interface.h
+++ b/src/freedv_interface.h
@@ -86,6 +86,9 @@ public:
     
     void transmit(short mod_out[], short speech_in[]);
     void complexTransmit(short mod_out[], short speech_in[], float txOffset, int nfreedv);
+    
+    struct MODEM_STATS* getCurrentRxModemStats() { return &modemStatsList_[modemStatsIndex_]; }
+    
 private:
     int txMode_;
     int rxMode_;
@@ -96,6 +99,8 @@ private:
     std::deque<SRC_STATE*> rateConvObjs_;
     COMP txFreqOffsetPhaseRectObj_;
     std::deque<COMP*> rxFreqOffsetPhaseRectObjs_;
+    struct MODEM_STATS* modemStatsList_;
+    int modemStatsIndex_;
     
     struct freedv* currentTxMode_;
     struct freedv* currentRxMode_; 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -50,7 +50,6 @@ int                 g_Nc;
 int                 g_mode;
 
 FreeDVInterface     freedvInterface;
-struct MODEM_STATS  g_stats;
 float               g_pwr_scale;
 int                 g_clip;
 int                 g_freedv_verbose;
@@ -861,7 +860,7 @@ void MainFrame::OnIdle(wxIdleEvent &evt) {
 // keeps CPU load reasonable
 //----------------------------------------------------------------
 void MainFrame::OnTimer(wxTimerEvent &evt)
-{
+{    
     if (evt.GetTimer().GetId() == ID_TIMER_PSKREPORTER)
     {
         // PSK Reporter timer fired; send in-progress packet.
@@ -885,6 +884,13 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
 
     if (freedvInterface.isRunning()) {
         int currentMode = freedvInterface.getCurrentMode();
+        if (currentMode != wxGetApp().m_prevMode)
+        {
+            // The receive mode changed, so the previous samples are no longer valid.
+            m_panelScatter->clearCurrentSamples();
+        }
+        wxGetApp().m_prevMode = currentMode;
+        
         if ((currentMode == FREEDV_MODE_800XA) || (currentMode == FREEDV_MODE_2400B) ) {
 
             /* FSK Mode - eye diagram ---------------------------------------------------------*/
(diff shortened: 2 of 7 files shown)


Tyrbiter:
I have now built this and with the program with freedv-gui hash e110b19 the behaviour is much improved.

I will report further after more testing if anything else seems amiss.

tmiw:
BTW @drowe67, the screenshot occurs is when playing ve9qrp_700c.wav with multi-RX on. With multi-RX off, the behavior is significantly different and more in line with what would be expected.

tmiw:
> I have now built this and with the program with freedv-gui hash [e110b19](https://github.com/drowe67/freedv-gui/commit/e110b1941c4c85396364e342bee81eb6fcceca9b) the behaviour is much improved.
> 
> I will report further after more testing if anything else seems amiss.

Thanks @Tyrbiter! Let us know for sure if anything else comes up. 👍

src/main.cpp
```diff
@@ -885,33 +884,65 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
...
+                case FREEDV_MODE_700E:
+                    g_Nc = 17; 
+                    m_panelScatter->setNc(g_Nc);
+                    break;
+                case FREEDV_MODE_2020:
+                    g_Nc = 31;
+                    m_panelScatter->setNc(g_Nc);
+                    break;
+            }
+
```
drowe67, line 933: So previously we were relying on g_Nc being set by the current Tx mode?  I can see how that would mess things up, as we would be using a different g_Nc to that required for the newly chosen Rx mode.

src/freedv_interface.h
```diff
@@ -96,6 +99,8 @@ class FreeDVInterface
     std::deque<SRC_STATE*> rateConvObjs_;
     COMP txFreqOffsetPhaseRectObj_;
     std::deque<COMP*> rxFreqOffsetPhaseRectObjs_;
+    struct MODEM_STATS* modemStatsList_;
+    int modemStatsIndex_;
```
drowe67, line 104: Ok so a separate set of modem stats for each mode?  That's a good idea, as a single global will be overwritten by the last demod your tried, which may not be the one that gets sync :+1:

drowe67:
Nice work @tmiw - using master I tried a few times to play the 700C demo file in multi-rx mode.  I couldn't repeat the problem OP reported regularly (just once), however I was experiencing a tiny scatter diagram that took some time to expand to normal size.

When I switched to the code in this branch the 700C scatter diagram would come up at the correct size immediately :+1

I also tried the demo files of other modes and they are plotting scatter diagrams well :+1: 

Thanks for your help in testing @Tyrbiter :slightly_smiling_face:

Tyrbiter:
I'm glad that I can help, so far I have not yet used freedv on air so finding and curing these problems seemed a good idea first.

One further thing I have noticed is with the ve9qrp_700d.wav file, when played the scatter plot starts with two dots at the upper left and lower left corners of the plot before the regular centred QPSK 4 dot constellation appears after about 8 seconds. I can only get this to be repeatable by starting the GUI, starting the modem with radio input muted, then play the ve9qrp_700d.wav file. Once it happens then replaying the file again also shows the same effect. It doesn't always happen when starting and stopping different codec mode recorded files in sequence. Not sure if it's very important, but it looks like something to do with the initialisation of the values used in the plotting code perhaps.

HTH

Brian

tmiw:
@drowe67, question: what issue is the below filter fixing other than preventing divide by zero?

```
        // smooth it out and set a lower limit to prevent divide by 0 issues

        m_filter_max_xy = BETA*m_filter_max_xy + (1 - BETA)*2.5*max_xy;
        if (m_filter_max_xy < 0.001)
            m_filter_max_xy = 0.001;
```

If I just use max_xy*2.5 directly it seems to solve the 700D issue @Tyrbiter mentioned as well as prevent the spectrum display going tiny. Would it be safe to just replace it with something like:

```
        // Prevent divide by zero.
        max_xy = fmax(0.001, 2.5*max_xy);
```

drowe67:
@tmiw that's a filter to slow down the AGC response of the scatter gain scaling.  Without it, the scaling will jump around very quickly on real world signals, in response to occasional noisy symbols, and probably result in a smaller scatter diagram.

The ideal 4 dots we see with the 700D test file won't happen in real world use - there is always noise.  Try the same test with 10dB of noise added using the Tools/Options/Simulation.

tmiw:
> @tmiw that's a filter to slow down the AGC response of the scatter gain scaling. Without it, the scaling will jump around very quickly on real world signals, in response to occasional noisy symbols, and probably result in a smaller scatter diagram.
> 
> The ideal 4 dots we see with the 700D test file won't happen in real world use - there is always noise. Try the same test with 10dB of noise added using the Tools/Options/Simulation.

Even better -- I tried sample 36 from the previous testing (without the proposed change) and it looked more like a cloud of dots. Additionally, western US->Australia was doing pretty well tonight so I recorded a few more samples (just me calling CQ). The those seemed to gravitate towards an X shape.

Anyway, what does 2.5 do? It should probably be commented/placed inside a #define or const variable for future maintainers.

drowe67:
> Even better -- I tried sample 36 from the previous testing (without the proposed change) and it looked more like a cloud of dots. Additionally, western US->Australia was doing pretty well tonight so I recorded a few more samples (just me calling CQ). The those seemed to gravitate towards an X shape.

Yes that's a classic scatter diagram for HF - some carriers are attenuated more than others by frequency selective fading, so the symbols are closer to the origin as their amplitude is less.

> Anyway, what does 2.5 do? It should probably be commented/placed inside a #define or const variable for future maintainers.

Good point - I've broken it out with comments in https://github.com/drowe67/freedv-gui/pull/146/commits/17dcb6b2a238481779b7064567132b8dc1113c15.  I may have also found the source of the 700D scatter issue @Tyrbiter [pointed out](https://github.com/drowe67/freedv-gui/pull/146#issuecomment-879914086) a few days ago (losing the RHS of the 2 points) - see what you think ...

tmiw:
> > Even better -- I tried sample 36 from the previous testing (without the proposed change) and it looked more like a cloud of dots. Additionally, western US->Australia was doing pretty well tonight so I recorded a few more samples (just me calling CQ). The those seemed to gravitate towards an X shape.
> 
> Yes that's a classic scatter diagram for HF - some carriers are attenuated more than others by frequency selective fading, so the symbols are closer to the origin as their amplitude is less.
> 
> > Anyway, what does 2.5 do? It should probably be commented/placed inside a #define or const variable for future maintainers.
> 
> Good point - I've broken it out with comments in [17dcb6b](https://github.com/drowe67/freedv-gui/commit/17dcb6b2a238481779b7064567132b8dc1113c15). I may have also found the source of the 700D scatter issue @Tyrbiter [pointed out](https://github.com/drowe67/freedv-gui/pull/146#issuecomment-879914086) a few days ago (losing the RHS of the 2 points) - see what you think ...

I just tested it with the sample files and it looks okay. I did notice that some of the plots start out with the points nearly all the way in the corners for a few seconds before moving closer to the center but as mentioned previously, signals with SNRs as strong as the sample files are unlikely in practice.

drowe67:
Yes the idea is to use as much of the area as possible, so symbols will a small variance will take up more of the screen.

Let me know when you're ready to merge this one.

tmiw:
Works for me unless other issues come up. 👍

Tyrbiter:
It certainly seems to be a fair improvement to me, so if @tmiw feels it's ready I wouldn't object to a merge.

I did see on the ML someone commenting on the BPSK carriers missing on 1600 mode if all modes are decoded simultaneously. Is that by design or not?

tmiw:
@Tyrbiter, I saw that email too and I believe the lack of BPSK carriers should be resolved once this merges into master. This is what I get when I try 1600 mode with the sample 1600 .wav file:

Screen Shot 2021-07-17 at 11 14 05 AM

Tyrbiter:
@tmiw Well that all sounds good then, it's one mode that I hadn't tried for some reason so I withdraw my comment.

It's much better for me here so I think a merge is a good move :)

tmiw:

PR #147, opened 2021-07-13
Upgrade wxWidgets to 3.1.5.

tmiw:
3.1.5 was released back in April, so might as well upgrade the static/macOS build to use it. 👍


tmiw:

drowe67:
Thanks @tmiw - so looking ag the cmake changes does this affect OSX only?  Just trying to figure out if there is there any way for me to test this.

tmiw:
@drowe67, that cmake file seems to support Linux as well so it should just be a matter of using -DBOOTSTRAP_WXWIDGETS=1 when building freedv-gui. It might be a good idea to test that, too, even though most people probably use their distro's version of the library.

drowe67:
OK on Linux I tried:
```
cmake -DLPCNET_BUILD_DIR=~/LPCNet/build_linux -DBOOTSTRAP_WXWIDGETS=1 ..
cmake
```
However it choked after a little while due to some GTK dependencies.  As you suggest, it doesn't really matter for Linux as we have a distro package available.

PR #151, opened 2021-07-25
Various last-minute bugfixes

tmiw:
Found a few issues:

1. The SNR calculation in the multi-RX code is significantly off from how it's displayed in the UI, so it's possible for it to have a lot harder time getting sync on some modes (e.g. 700C). This PR updates how that's calculated so that the SNR it uses matches what the UI shows.
2. Scatter plot settings are being improperly adjusted when TX mode changes. The PR updates that logic so that the settings are only modified on OnTimer in response to RX mode changes.
3. Various memory leaks causing eventual crashes.


tmiw:
BTW, issue 1 manifests itself most easily if you comment out the logic that locks RX to the mode that was previously in sync (lines 627-636 of freedv_interface.cpp on master) and play ve9qrp_700c.wav.

Tyrbiter:
What is the status of the ms-end-tx-delay branch, I see it's ahead of master and I noticed earlier a lot of stations chopping off their last word or so of a transmission.

tmiw:
@drowe67, that branch was abandoned in favor of https://github.com/drowe67/freedv-gui/pull/123 I thought?

Tyrbiter:
Well that looks a lot more comprehensively tested and developed, my observation could of course be due to people that just let go of the ptt before they stop speaking.

tmiw:
> Well that looks a lot more comprehensively tested and developed, my observation could of course be due to people that just let go of the ptt before they stop speaking.

They may also be running older versions of the software without that fix, too.

Tyrbiter:
Yes, they could. It's not always easy to know the detail of that on air :)

Tyrbiter:
I built this version earlier today and it seems to be working correctly, the modem stats appear to correspond with the signal levels and quality I am hearing.

Tyrbiter:
I have built and installed this newest version on Fedora 34, seems to be working well, start up seemed faster again.

I will see what it's like on air later today.

tmiw:
@drowe67, I think this is ready for merge.

src/eq.cpp
```diff
@@ -50,9 +50,25 @@ void *MainFrame::designAnEQFilter(const char filterType[], float freqHz, float g
 
 void  MainFrame::designEQFilters(paCallBackData *cb, int rxSampleRate, int txSampleRate)
 {
+    bool deleteFilters = false;
+    
+    if (m_newMicInFilter && (cb->sbqMicInBass || cb->sbqMicInTreble || cb->sbqMicInMid))
+    {
```
drowe67, line 56: Won't cb->sbqMicInBass, cb->sbqMicInTreble and cb->sbqMicInMid, always be non-null?  Is it possible to have a condition where just one of them is null?

src/eq.cpp
```diff
@@ -62,25 +78,60 @@ void  MainFrame::designEQFilters(paCallBackData *cb, int rxSampleRate, int txSam
...
 void  MainFrame::deleteEQFilters(paCallBackData *cb)
 {
-    if (m_newMicInFilter) {
-        sox_biquad_destroy(cb->sbqMicInBass);
-        sox_biquad_destroy(cb->sbqMicInTreble);
-        sox_biquad_destroy(cb->sbqMicInMid);
+    if (m_newMicInFilter) 
+    {
+        if (cb->sbqMicInBass) 
+        {
```
drowe67, line 98: As above - is it possible to have just one of these filters null, and the others non-null?

src/eq.cpp
tmiw: Thinking about it more, I don't think there is. I'll assert, though, as that should probably be fixed if someone does run into such a case.

drowe67: I agree - an assert is a good idea :+1:  Although this code has been pretty stable for many years, so I'd be surprising if it has any problems.  Still, multi-rx might introduce new conditions.

Does this mean these changes to `eq.cpp` are not necessary and can be reverted?

src/eq.cpp
drowe67: Can all these changes can be reverted too, simplifying the code?

src/eq.cpp
tmiw: Yep, will do.

src/eq.cpp
tmiw: Yep, will do.

tmiw:
Hmm, pushing the Stop button crashes now:

```
Assertion failed: (cb->sbqSpkOutBass == nullptr && cb->sbqSpkOutTreble == nullptr && cb->sbqSpkOutMid == nullptr), function deleteEQFilters, file /Users/mooneer/devel/freedv-gui/src/eq.cpp, line 116.
```

I'll have to revert that commit.

drowe67:
> Hmm, pushing the Stop button crashes now:
> 
> ```
> Assertion failed: (cb->sbqSpkOutBass == nullptr && cb->sbqSpkOutTreble == nullptr && cb->sbqSpkOutMid == nullptr), function deleteEQFilters, file /Users/mooneer/devel/freedv-gui/src/eq.cpp, line 116.
> ```
> 
> I'll have to revert that commit.

Hmm, something no quite right about that.  Might be best to get to the root cause.  It would be good to understand how a condition can exists where the filters aren't initialised together.

tmiw:
@drowe67, yeah, it turns out the assertion condition was wrong; it should be != nullptr and not == nullptr :) The PR has been updated accordingly.

src/eq.cpp
drowe67: So can we also revert the changes in `designEQFilters()` between line 55 and 75 :point_up: :point_down: ie is there a specific need for all the if then clauses checking each filter?

tmiw: The latest commit only checks sbqMicInBass and sbqSpkOutBass != nullptr and asserts on the others. For example:

```
    if (m_newMicInFilter && cb->sbqMicInBass != nullptr)
    {
        // All three should be valid if one is, so we assert that here.
        assert(cb->sbqMicInBass != nullptr && cb->sbqMicInTreble != nullptr && cb->sbqMicInMid != nullptr);
        
        deleteFilters = true;
    }
```

drowe67: Gotcha. So why is it necessary to check `cb->sbqMicInBass != nullptr` ? Sorry if I'm  missing something - I'm unsure what problem the lines 73-75 actually fix?

tmiw: deleteEQFilters() still asserts on the filters being not null prior to calling sox_biquad_delete(). Even if it didn't, sox_biquad_delete() dereferences the pointers passed into it to free the filters' internal data prior to setting them to nullptr. The setting to nullptr is to ensure that we don't inadvertently touch already-freed memory (in other words, fail much more quickly if it somehow did touch the filters when they were invalid).

tmiw: Plus, failing immediately on attempted dereference means that it should be easier to figure out what happened vs. having potentially inconsistent behavior by touching memory it shouldn't.

drowe67: > deleteEQFilters() still asserts on the filters being not null prior to calling sox_biquad_delete(). 

OK, good catch.  So the core issue is `deleteEQFilters() ` and `designEQFilters` are being called out of order? I agree with the asserts and assignment of pointers to null to highlight the problem.  But I feel we should track down the root cause for asserts firing.  I'm happy to look into this one myself if you can tell me how to reproduce the problem.

tmiw: I'm not sure there's actually a problem at the moment now. The main issue was the memory leaks around the filters, and I think that was solved by the change in sox_biquad.c lines 95 and 104 + setting m_newMicInFilter and m_newSpkOutFilter to true prior to deleteEQFilters() in stopRxStream().

tmiw:
@drowe67, I was able to revert some of the designEQFilters changes since it turns out they aren't actually needed to stop the memory leak I was seeing. Unless there's anything else, it seems like this is ready for mege.

PR #153, opened 2021-08-07
Use UTF8 encoding for PortAudio device names.

tmiw:
PortAudio uses UTF8 for device names (see [this](https://github.com/PortAudio/portaudio/search?q=WideCharToMultiByte) for conversions it does in Windows and [this](https://github.com/PortAudio/portaudio/blob/2bd663ca75f424c658804f970d3045166a6d09d3/src/hostapi/coreaudio/pa_mac_core.c) for conversions on macOS). FreeDV was assuming ASCII before, thus causing display and usage issues on non-US computers (especially if the device names don't use ASCII, e.g. if they have Japanese characters per report on the digitalvoice list). This PR resolves that issue.


tmiw:
I'll generate a test build once I hear back from Peter as his problem may be similar.

tmiw:
@drowe67, I think this PR is good now. We'll see what happens with that Rec/Play 2s issue.

Should a 1.6.1 release be made after this merges or should we wait? I'm not sure how common non-English characters are going to be in sound devices.

drowe67:
Hmm, good question.  Maybe a test release for those who are interested in this fix?  But I'm OK either way....

PR #156, opened 2021-08-14
PSK Reporter encoding/decoding improvements

tmiw:
Alternate implementation of PSK Reporter improvements:

* Uses codec2's [reliable_text](https://github.com/drowe67/codec2/pull/209) API (which uses LDPC(112,56) with a max of 8 characters)
* Supports manual frequency entry for those without Hamlib enabled.

This obsoletes PR #98. Note that this PR will break compatibility for PSK Reporter with previous versions.


Tyrbiter:
There are some places with 7 character callsigns, VK#Fxyz are Australian foundation licences I believe.

tmiw:
@Tyrbiter, thanks for that. I bumped up the limit to 8 characters, which should still be good if we go LDPC.

drowe67:
Hi @tmiw - a good way to develop this protocol is to break it out as a separate module, so it can be tested on the command line, just like the recent stm32 700E tests.   It's very easy for subtle issues to slip in that will be difficult to debug or test using a GUI application and over the air testing.

Design wise, libcodec2 might be a better place for it.

One interesting "unit test" for the protocol would be to see how it stands up to typical simulated channels.  For example it should have a low packet error rate at SNRs where the voice quality is fair.

Gut feel is a block code like Golay should be sufficient, LDPC isn't that much better.

drowe67:
Just thinking about requirements:
1. How many bits (bytes) would you like in a payload?
1. How often do we need an update? (e.g. once every 10s, once an over)?
1. What PER is acceptable at a given SNR for AWGN and fading channels?

Also wondering if there are other ways to send the info that a few bits/frame - like a RSID type frame we inject at the start or end of an over, or periodically into the speech data stream.

Tyrbiter:
Tried running the PSK reporter improvements and I get a crash after some amount of time. Sometimes nearly instant, sometimes takes a little while.

gdb backtrace attached.

[gdb_output_freedv.txt](https://github.com/drowe67/freedv-gui/files/6987908/gdb_output_freedv.txt)

tmiw:
@Tyrbiter, I checked in a change that might help. Give that a shot?

@drowe67, yeah, I think moving the bulk of this logic to codec2 would be a good idea. To answer your questions:

> How many bits (bytes) would you like in a payload?

I think enough to hold about 95% of callsigns (+ ideally any prosigns) would be good. That seems to be about 7-8 bytes at the moment but could change.

> How often do we need an update? (e.g. once every 10s, once an over)?

I'd say once per transmission is adequate. However, this also means that someone listening in might miss the report that happened just before changing frequency; I'm not sure how important getting that report would be.

> What PER is acceptable at a given SNR for AWGN and fading channels?

Speaking in terms of SNR, I'd say few if any errors would be acceptable above 1-2 dB over the lowest possible SNR for a given mode (e.g. 0 dB for 700D).

Tyrbiter:
I was listening to some stations using 700D and 700E on 80m earlier today, one in particular was rarely audible using freedv, analogue SSB was equally poor, occasional syllables could be heard, 700D was struggling for sync. That is the case where some well-protected callsign data would be really useful, knowing distance and direction about the station is important.

It was noticeable how much better 700E was under the fast-fading conditions, it seemed to be worth anything up to 3-4dB SNR. I wonder if the callsign data could be modulated differently to the normal mode used for the speech?

Tyrbiter:
The code change has allowed freedv to keep running for quite a long period, seems good.

tmiw:
Good to hear, @Tyrbiter!

Anyway, I'm working on moving the main logic for the text encoding to codec2: https://github.com/drowe67/codec2/pull/209

Tyrbiter:
OK, good, will monitor progress there and test as necessary.

tmiw:
This PR now requires https://github.com/drowe67/codec2/pull/209 to compile. Based on minimal testing, it seems to work well on a live AWGN channel (to a KiwiSDR a few miles/km away from me). Some testing here would be good, then maybe we can do a test build and send that out to a wider audience.

drowe67:
Fine Business @tmiw - looking forward to seeing how it works in the wild :slightly_smiling_face:

tmiw:
@drowe67, are you able to access portaudio.com by any chance? It's been down for me since last night so I'm currently unable to do any Windows builds (as that builds static PortAudio).

Tyrbiter:
Down for me too Mooner, a shame as I'm very impressed with your pskreporter changes.A question; do both ends of a contact need to be running the new code for this to work? Will it make any difference if pskreporter is disabled?-- Brian. G8SEZSent from my Galaxy
-------- Original message --------From: Mooneer Salem ***@***.***> Date: 24/08/2021  19:09  (GMT+00:00) To: drowe67/freedv-gui ***@***.***> Cc: Brian ***@***.***>, Mention ***@***.***> Subject: Re: [drowe67/freedv-gui] PSK Reporter encoding/decoding improvements (#156) 
@drowe67, are you able to access portaudio.com by any chance? It's been down for me since last night so I'm currently unable to do any Windows builds (as that builds static PortAudio).

tmiw:
Yeah, I switched freedv-gui to the PA GitHub repository to keep from holding this PR up. Currently building some Windows builds so we'll see how that goes.

tmiw:
https://www.dropbox.com/sh/044azug3aj6zsnt/AAAh573m0tv-MGweRa0mt-1ba?dl=0

drowe67:
> @drowe67, are you able to access portaudio.com by any chance? It's been down for me since last night so I'm currently unable to do any Windows builds (as that builds static PortAudio).

Down for me too ..

tmiw:
So far, so good:

Screen Shot 2021-08-27 at 4 52 10 PM

tmiw:
So yeah, it looks like this + the corresponding codec2 changes are working well. I think we can start reviewing both PRs and merging them.

build_linux.sh
```diff
@@ -9,7 +9,7 @@ export CODEC2DIR=$FREEDVGUIDIR/codec2
 export LPCNETDIR=$FREEDVGUIDIR/LPCNet
 
 # change this when working on combined codec2/freedv-gui changes
-CODEC2_BRANCH=master
+CODEC2_BRANCH=ms-reliable-text
```
drowe67, line 13: @tmiw should we change this back to `master` prior to merge?

build_windows.sh
```diff
@@ -21,7 +21,7 @@ export LPCNETDIR=$FREEDVGUIDIR/LPCNet
 # First build and install vanilla codec2 as we need -lcodec2 to build LPCNet
 cd $FREEDVGUIDIR
 git clone https://github.com/drowe67/codec2.git
-cd codec2 && git checkout master && git pull
+cd codec2 && git checkout ms-reliable-text && git pull
 mkdir -p $BUILD_DIR && cd $BUILD_DIR && rm -Rf *
```
drowe67, line 25: and here I guess ...

drowe67:
Looks great @tmiw :slightly_smiling_face: Pls feel free to press merge when you've tweaked those branches back to master.

tmiw:
@drowe67, can #158 be merged in first? I merged that PR into this one at some point so that both features would be available for test builds. I'll work on switching things back to master in the meantime.

build_windows.sh
tmiw: Done.

build_linux.sh
tmiw: Done.

PR #158, opened 2021-08-15
Update mode display in status bar on PTT start/stop.

tmiw:
Per request on digitalvoice mailing list, this performs an update of the mode display whenever the user pushes the PTT button. This allows earlier notification of incorrect mode usage (vs. waiting for stop/start or a PSK Reporter report).

### DIFF
--- a/src/hamlib.cpp
+++ b/src/hamlib.cpp
@@ -259,6 +259,7 @@ void Hamlib::enable_mode_detection(wxStaticText* statusBox, bool vhfUhfMode)
     {
         m_modeBox->SetLabel(wxT("unk"));
         m_modeBox->Enable(false);
+        m_modeBox->SetForegroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT));
         m_modeBox = NULL;
     }
     else
@@ -292,6 +293,7 @@ void Hamlib::disable_mode_detection()
         // Disable control.
         m_modeBox->SetLabel(wxT("unk"));
         m_modeBox->Enable(false);
+        m_modeBox->SetForegroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT));
         m_modeBox = NULL;
     }
 }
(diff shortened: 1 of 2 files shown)


tmiw:
@Tyrbiter, I noticed a bit more latency when pushing the PTT button but I'm not sure how much it'd affect people in practice (I use a SDR with CAT control over TCP/IP).

Tyrbiter:
I have just tried it, noting that the reason for the change improves the update rate and immediately signals the radio mode and brings up the red error colour if the wrong sideband is selected on the radio.

Using my TS-890 I don't see any apparent latency change going from Rx->Tx it's essentially instant, the Tx->Rx case might have a little latency but I think that's probably due to emptying the Tx buffers before dropping PTT which is a fairly recent change. The USB CAT is running at 115200 so the time taken to read the mode (and frequency?) is the minimum possible. Maybe if someone uses much slower CAT it would be noticeable?

tmiw:
Another thought was to try to delay the update until after TX starts. I'm not sure offhand if there are any radios that don't support responding to non-PTT stop commands during TX, though.

Anyway, this might be worth putting in as part of a test build to make sure there's no obvious issues.

Tyrbiter:
My thought right now is to wait for a few weeks until hamlib-4.3 lands, there are a lot of changes being made to deal with the complexity of the main/sub VFOA/VFOB and split problems that some people are suffering with. It's all very confusing and adding this change into the mix is not sensible right now.

I believe that hamlib-4.3 is currently supposed to release by 1st September, but again, previous releases have been delayed ;-)

tmiw:
FWIW, the macOS version uses whatever Hamlib is in GitHub master at the time the build's made and I haven't had any problems. FreeDV doesn't rely on much from the library, though (just PTT and VFO A frequency/mode).

Tyrbiter:
OK, I did a little more testing, trying to trick it by setting the wrong sideband on the unselected VFO but it reliably identifies the selected VFO every time. When the wrong mode is selected, on clicking stop the 'unk' is still red, not sure if that is logically correct but it makes the warning status more obvious.

I have been using the released hamlib-4.2 rather than github master, it seems OK but then freedv doesn't use it extensively which I think is when the bugs bite.

Tyrbiter:
The fix makes the 'unk' label colour correct again. Thanks.

PR #166, opened 2021-09-08
Fix off by one error preventing UTF8 encoded callsign from being null terminated.

tmiw:
Looks like an off by one error in freedv-gui was causing my callsign to be reported incorrectly by others on PSK Reporter. This PR fixes that issue.

\+ @drowe67

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1606,8 +1606,9 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
         }
         else
         {
-            char temp[1024];
-            strncpy(temp, wxGetApp().m_psk_callsign.ToUTF8(), wxGetApp().m_psk_callsign.Length() + 1);
+            char temp[9];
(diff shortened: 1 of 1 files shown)


tmiw:
So https://github.com/drowe67/codec2/pull/227 seems to have no impact on the fix but it's probably good to merge anyway.

src/main.cpp
```diff
@@ -1606,8 +1606,9 @@ void MainFrame::OnTogBtnOnOff(wxCommandEvent& event)
         }
         else
         {
-            char temp[1024];
-            strncpy(temp, wxGetApp().m_psk_callsign.ToUTF8(), wxGetApp().m_psk_callsign.Length());
+            char temp[9];
+            memset(temp, 0, 10);
```
drowe67, line 1610: @tmiw I think this is writing beyond the end of the array

tmiw: Yep, thanks! Updated.

PR #169, opened 2021-09-12
Resolve heap overflow during txCallback triggered by BT usage.

tmiw:
Possible fix for #144. I noticed the following AddressSanitizer output while trying to use my Bluetooth headset with FreeDV:

```
rig.c(2173):rig_get_freq return(0)
    #0 0x105eff12d in MainFrame::txCallback(void const*, void*, unsigned long, PaStreamCallbackTimeInfo const*, unsigned long, void*) main.cpp:3047
    #1 0x1073cf8dc in AdaptingOutputOnlyProcess+0x22c (libportaudio.2.dylib:x86_64+0x88dc)
    #2 0x1073ceb35 in PaUtil_EndBufferProcessing+0x1a7 (libportaudio.2.dylib:x86_64+0x7b35)
    #3 0x1073d38b3 in AudioIOProc+0x532 (libportaudio.2.dylib:x86_64+0xc8b3)
    #4 0x11b231a7d  (CoreAudio:x86_64+0xea7d)
    #5 0x11b257993  (CoreAudio:x86_64+0x34993)
    #6 0x11b234c86  (CoreAudio:x86_64+0x11c86)
    #7 0x7fff21e59a8b in invocation function for block in HALC_ProxyIOContext::HALC_ProxyIOContext(unsigned int, unsigned int)+0x1985 (CoreAudio:x86_64+0x17ea8b)
    #8 0x7fff21ff4b03 in HALB_IOThread::Entry(void*)+0x47 (CoreAudio:x86_64+0x319b03)
    #9 0x7fff204de8fb in _pthread_start+0xdf (libsystem_pthread.dylib:x86_64+0x68fb)
    #10 0x7fff204da442 in thread_start+0xe (libsystem_pthread.dylib:x86_64+0x2442)

0x615000179d00 is located 0 bytes to the right of 512-byte region [0x615000179b00,0x615000179d00)
allocated by thread T0 here:
    #0 0x1086d6460 in wrap_malloc+0xa0 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x48460)
    #1 0x1073ce40a in PaUtil_InitializeBufferProcessor+0x3d8 (libportaudio.2.dylib:x86_64+0x740a)
    #2 0x1073d1bbe in OpenStream+0x6b5 (libportaudio.2.dylib:x86_64+0xabbe)
    #3 0x1073cd944 in Pa_OpenStream+0x104 (libportaudio.2.dylib:x86_64+0x6944)
[...]

PR #168, opened 2021-09-12
Final tidying up before official 1.6.1 release.

tmiw:
Updates:

* Ensure that memory used by PSK Reporter is freed on window close.
* Documentation/CMakeLists updates to reflect new release.

PR #173, opened 2021-09-15
Suppress unnecessary rig_init in hamlib.

tmiw:
Per #172, this only calls rig_init() if the user changes the radio type or on first start.

@drowe67, @Tyrbiter, let me know if this looks okay.

### DIFF
--- a/src/hamlib.cpp
+++ b/src/hamlib.cpp
@@ -39,6 +39,7 @@ static int build_list(const struct rig_caps *rig, rig_ptr_t);
 
 Hamlib::Hamlib() : 
     m_rig(NULL),
+    m_rig_model(0),
     m_modeBox(NULL),
     m_freqBox(NULL),
     m_currFreq(0),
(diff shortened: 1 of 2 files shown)


drowe67:
Looks good to me @tmiw :+1: It would be good to test this change with a wider audience as part of the next beta, PTT has been a tricky area in the past

tmiw:
I'll generate a new test build tonight containing this and #169. Then if good, we can merge all three pending PRs and generate a final 1.6.1 build.

tmiw:
Just to confirm, this is with https://github.com/Hamlib/Hamlib/issues/797 applied too, right? If I understood the discussion thread correctly, this PR + that patch should have done the trick (although the VFO will still switch to VFO A on first start, just suppressing it on future starts while the app is open).

Anyway, I'd rather not have FreeDV always keep Hamlib connected as that's a more significant design change.

tmiw:
BTW, I just emailed out [the RC build](https://www.dropbox.com/sh/gfrszhss1htpzux/AABqezaHdni9yYJjrRHMkOeDa?dl=0) (based on #168 which contains all current open PRs).

Tyrbiter:
No, I was using the released hamlib-4.3.1, so I will correct that mistake and report back.

Tyrbiter:
I'm with @drowe67 regarding the trickiness of this, so I don't want to hold up the new release and risk breaking this area of the program.

Thinking cap back on...

tmiw:
I haven't heard of any issues with the test build, so we can probably merge this now.

PR #175, opened 2021-10-08
Make multi-RX in single threaded mode the default.

tmiw:
Per discussion with @drowe67, this makes single threaded mode the default for multi-RX.

(Additionally, this PR also makes multi-RX enabled by default. I don't think it'll hurt to do so but I can keep multi-RX itself disabled by default if desired.)


tmiw:
Yeah, this changes the default for new installations, not necessarily for existing ones. If multi threaded mode does end up causing widespread issues on the existing install base, it shouldn't be a problem recommending that setting be changed.

PR #183, opened 2021-11-06
Build: Better error handling and let git ignore temporary build directories.

aknrdureegaesr:
The `build_linux.sh` script says what it is doing and stops when encountering an error.

Directories that are constructed temporarily during the build
are now ignored by git, as we do not want to version those.

Ah, one more thing: `bash` now officially lives in `/usr/bin`, the `/bin` directory is deprecated.

Vy 73, Andreas, DJ3EI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+/build_linux/
+/build_osx/
+/codec2/
+/hamlib/
+/hamlib-code/
+/LPCNet/
+/macdylibbundler/
+
+a.out
+*.o
+*.make
+*.log
+*.marks
--- a/build_linux.sh
+++ b/build_linux.sh
@@ -1,8 +1,11 @@
-#!/bin/bash
+#!/usr/bin/bash
 # build_ubuntu.sh
 #
 # Build script for Ubuntu and Fedora Linux, git pulls codec2 and
 # lpcnet repos so they are available for parallel development.
+
+# Echo what you are doing, and fail if any of the steps fail:
+set -x -e
 
 export FREEDVGUIDIR=${PWD}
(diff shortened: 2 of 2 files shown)


tmiw:
The general C/C++ .gitignore template should be included in .gitignore too for those who want to build freedv-gui manually and not via the script. Also, it would be good to have the following for build_osx.sh:

```
build_osx
hamlib-code
hamlib
macdylibbundler
```

drowe67:
Thanks @aknrdureegaesr :slightly_smiling_face: 

@tmiw are your [comments above](https://github.com/drowe67/freedv-gui/pull/183#issuecomment-962473385) something you would like to see in this PR, or more general issues for another PR?

tmiw:
> @tmiw are your [comments above](https://github.com/drowe67/freedv-gui/pull/183#issuecomment-962473385) something you would like to see in this PR, or more general issues for another PR?

I figured it'd be good to include here since we're already touching .gitignore. However, it can go in another PR if preferred.

tmiw:
Looks good here 👍

aknrdureegaesr:
At your service...

Adding the directory was easy. I'm not entirely sure what a good C/C++ - template would be. So I checked which kind of files are actually created below `build_linux`, took some care that those are not also present in already checked-in material, and this is how I came up with the list you see now.

Assuming you do not want to have several commits in the history, I replaced my previous commit with the new suggestion.

tmiw:
@aknrdureegaesr, actually, several commits is no problem. Something for future reference.

@drowe67, any additional comments or should I go ahead and merge?

drowe67:
@tmiw I'm happy if you are :slightly_smiling_face:

PR #184, opened 2021-11-06
Use CMake to generate macOS app bundle.

tmiw:
Some cleanups in the CMakefiles to allow CMake to generate the app bundle for macOS instead of running commands manually. This also uses a forked copy of macdylibbundler that fixes a build issue discovered while building on macOS Monterey.


drowe67:
@tmiw - I can't really comment on this one as it's OSX related.  So if you're happy I'm happy :slightly_smiling_face:

Issue #185, opened 2021-11-10
Feature request - Please provide simplified audio setup for recieve only systems.


g0hww:
It is not possible to quickly configure a receive only system to allow me to rapidly listen to freedv transmissions without being forced to go through the process of configuring transmit audio paths.

I would like to be able to disable the transmit capability, so that I can simply configure the receive audio paths.

My HF kit is capable of transmitting, and is configured to allow all routing of Tx and Rx audio for digimode progs such as JS8Call, QSSTV, ARDOP or FLDigi through Pulse Audio.  I don't tend to keep a headset connected to that Linux box, so generally don't have a mic available. More often than not, I just want to listen to some FreeDV audio, with analog voice coming from the laptop's speakers.

Going a bit off topic now ...

Ideally, all input and output streams could be pointed to the "pulse" device, with actual configuration left to tinkering with PAVU Control.  I have some hopes that a resolution of #16 would improve the situation. However, the 2 second test is not sufficient to switch back and forth between PAVU Control and the FreeDV GUI in order to configure pulse audio externally. It would be preferable to have an infinite test duration.  Also,  the UI does not let me configure Pulse as the device for all 4 audio inputs and outputs.  I feel that it should allow this.  Obviously it would be down to the user to ensure that things are sanely configured.

drowe67:
> It is not possible to quickly configure a receive only system to allow me to rapidly listen to freedv transmissions without being forced to go through the process of configuring transmit audio paths.

This is covered at the start of the FreeDV manual.

Re off topic and other detailed discussion - pls open a topic in [Discussions](https://github.com/drowe67/freedv-gui/discussions).  We'd like to use Issues only for demonstrable bugs.

g0hww:
Oh, wow. OK. I have dabbled with FreeDV on and off for a number of years and never even considered looking for a manual before. Thanks.

tmiw:
BTW, there's also some info [here](https://groups.google.com/g/digitalvoice/c/robpK8BOjfI/m/Pa-K4IEaBAAJ) on building a version of PortAudio that has PulseAudio enabled and subsequently including it in FreeDV. That will at least allow Pulse devices to show up instead of needing to choose the generic "pulse" device. Once that PR finally gets into PortAudio and new distro packages made, FreeDV should be able to handle PulseAudio with little to no modification.

drowe67:
[...]

PR #187, opened 2021-11-12
Install user manual along with executable

tmiw:
Installs the user manual alongside the executable per comments in https://github.com/drowe67/freedv-gui/issues/185.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -560,11 +560,11 @@ if(WIN32)
     set(CPACK_NSIS_MODIFY_PATH OFF)
     set(CPACK_NSIS_MENU_LINKS
         "http://freedv.org" "FreeDV Homepage")
-    set(CPACK_NSIS_EXTRA_UNINSTALL_COMMANDS "reg delete HKEY_CURRENT_USER\\SOFTWARE\\CODEC2-Project\\FreeDV /f")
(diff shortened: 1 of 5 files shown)


tmiw:
@drowe67, not sure if there's a cross platform way of opening web pages so not sure about option 3 in the original comment yet. Suggestions appreciated for sure.

drowe67:
Hi @tmiw - I'm sorry I thought I had commented on this one, but perhaps I didn't press the Comment button - Ooops!

This WxWidgets control might be useful for hyperlinks: https://docs.wxwidgets.org/trunk/classwx_hyperlink_ctrl.html.

However I think the key issue here is to alert first time users that a manual exists.  After that it's pretty easy to find (e.g. Google, freedv.org), even if we don't have it hyperlinked.  The natural place to look is the Help Menu.  So perhaps we could:
1. Add some text to the Help-About dialog "see on line FreeDV Manual for extensive help".
1. Add a Help-Manual option.
1. We could add a short URL link e.g. freedv.org#manual that's easy to type in even if it's not a hyperlink.

tmiw:
Turns out there's a built-in wxWidgets function for opening the browser. Added "User Manual" to the Help menu.

drowe67:
Thanks @tmiw that's great :smile: For a new user it's quite natural to look at the Help menu, and the manual option is now very obvious.  Feel free to merge and close #185 if you are happy.

PR #82, opened 2020-12-30
Autodetect sound device config changes in FreeDV and notify user

tmiw:
Moving PortAudio changes originally in #77 to a separate PR for further debugging/discussion. Currently, Virtual Audio Cable causes problems with Windows due to how devices are named, so this is a WIP.

PR #189, opened 2021-11-28
Create "Easy Setup" dialog to simplify initial setup.

tmiw:
This PR creates an "Easy Setup" dialog under the Tools menu that aims to provide a much simplified initial FreeDV setup. This dialog also comes up on initial FreeDV startup after installation (instead of the previous "this may be an initial install" dialog implemented in #82).

Current functionality:

* Audio configuration:
    * Avoids interfering with any existing configuration unless specifically modified by the user.
    * For Flex radios, intelligently combines "DAX Audio RX" devices with the single "DAX Audio TX" device, avoiding the need to use the advanced sound setup.
    * Defaults to the built-in speaker/mic for the "analog" device; selecting another analog device requires advanced setup.
    * On Windows machines, merges the two separate mic and speaker devices into a single selectable device.
    * Advanced button to go to the existing audio configuration dialog.
* PTT/CAT configuration
    * Hamlib support
    * Serial port PTT support
    * Advanced button to go to the existing PTT configuration dialog.
    * "Test" button activates PTT as well as emits a sine wave on the radio's audio device to verify power out.
* Reporting configuration (callsign/grid square for PSK Reporter).


tmiw:
Merged this and #186 into v1.6.2-test to make test build generation easier. Should have something out tonight for initial testing hopefully.

tmiw:
So we can facilitate #194, let's go ahead and review/merge this now. The serial PTT support can be added in a future PR.

drowe67:
@tmiw this is an interesting experiment in the ease of set up meme.  Thank you for working on it. However it is a significant amount of extra code that will need to be maintained and supported into the future.  It's essentially a parallel path for the existing setup system. So there needs to be a good reason for merging it.

Do we have any feedback from end users to determine if this PR makes set up significantly easier? I appreciate this may be difficult to gauge, as people are only novice users once.

Can we use this to replace the current set up system, rather than providing two ways of doing the same thing?

tmiw:
> @tmiw this is an interesting experiment in the ease of set up meme. Thank you for working on it. However it is a significant amount of extra code that will need to be maintained and supported into the future. It's essentially a parallel path for the existing setup system. So there needs to be a good reason for merging it.
> 
> Do we have any feedback from end users to determine if this PR makes set up significantly easier? I appreciate this may be difficult to gauge, as people are only novice users once.
> 
> Can we use this to replace the current set up system, rather than providing two ways of doing the same thing?

There's some feedback [here](https://groups.google.com/g/digitalvoice/c/xLzf9o2KNiA). The main ask seems to be serial PTT, which I haven't fully figured out how to integrate here yet.

As for whether we can replace the existing setup with this, unfortunately not 100%. For instance, some setups (such as for SDRs) use different devices for both receive and transmit. I've also seen a report of someone using two different devices on the analog side as well.

drowe67:
Hi @tmiw - the comments seem to focus on the need for additional various options required for different radios, rather that the core goal of ease of set up for people new to the software.

This once again suggests that set does have some irreducible minimum complexity.  I guess there is no getting away from the need for four sound devices and PTT control across a variety of use cases.  I have gone around this "thought loop" myself a few times!

Perhaps the solution to helping people set up FreeDV might not be in additional software, but other media such as YouTube guides (that's what I use when I want a new skill).

Anyhoo, it seems like this PR is a work in progress - thanks for experimenting with this. Is it OK if we defer merging it is feature complete and demonstrated real gains in the ease of setup area?

cybork42:
Making things easier with the audio path configuration really is important. Even after some time of using, I find the two screens for transmit and receive to be difficult to use. The first time user will get through it somehow, but it stays difficult even afert that.

I agree that there is need to configure each of the four paths individually in many scenarios. The easy setup is a TRX on one USB port and the headset on another system device (internal or USB sound card or... what have you).

So in a sense having an image with a box in the middle (FreeDV), a headset on the right side, a TRX icon on the left and a checkbox "joint channel" or so on each side (TRX and headset), could be an easy setup. 
Pulldown for that joint channel on TRX side, second pull-down on headset side, with channels as shown on the long lists in current software.

If one of those "joint channel" checkboxes gets unticked, a second pulldown is no longer greyed out on the corresponding side and allows to make a separate pull-down choice for "in" and "out" signal paths. The former "joint channel selector" turns into "in" or "out", the no longer greyed out one accordingly. 

For example in cases where you use a sudio mic and a headset, connected to different system audio devices, or with SDR radios.

How about that?

tmiw:
@cybork42, have you had an opportunity to try the code in this PR? (I have Windows and Mac builds [here](https://www.dropbox.com/sh/w5ubs3dvod907zd/AAD5NYdldHoV2Y-A1WuK55WMa?dl=0) as well.) Does that get much closer to "ideal" for ease of setup? Perhaps this PR can be a starting point for additional changes if needed.

cybork42:
Not yet. I will give it a try as soon as possible. Probably in a day or two.
Will provide an update 😀. 

Am 25. Dezember 2021 03:25:57 MEZ schrieb Mooneer Salem ***@***.***>:
***@***.***, have you had an opportunity to try the code in this PR? (I have Windows and Mac builds [here](https://www.dropbox.com/sh/w5ubs3dvod907zd/AAD5NYdldHoV2Y-A1WuK55WMa?dl=0) as well.) Does that get much closer to "ideal" for ease of setup? Perhaps this PR can be a starting point for additional changes if needed.

cybork42:
hello @tmiw, Merry Christmas!

This is an update after testing FreeDV 1.6.2-devel-20211128-779e8f0 in 64bit version on Win10;
preconditions: IC7300 hooked up, USB sound card for headset hooked up; earlier install of FreeDV on same computer available.
observations:
1. chose easy setup
2. Step1: select sound device shows "(multiple)", but just have IC7300; no other choice in pull-down besides "(multiple)"
3. "decided audio ..." and "transmitted audio..." chose my USB sound card
4. Step2 CAT Control shows my com port and some settings for IC7300
5. click on "advanced" leads nowhere, 
6. click on "Test" activates TRX to start TX (why not say "test TX switch to transmit" or something?)
6. click on "advanced" still leads nowhere
7. "enable reporting" chosen active with right call sign (probably from Win registry?)
8. click "Ok".
9. on main screen click "PTT" -> works.

Surprising: never have had any selection for TRX USB interface anywhere for Audio path, display was "multiple", but works in spite of that.

cybork42:
My Idea in simple ASCII graphics was more general, maybe something like this:
´´´
                                                                                                 ^
                                                                                                 |
             Simple                              Simple                                          |ANT
              +-+ +---------------++              +-+ +---------------++                         |
              |X| | USB #3 (USBSo ||              |X| | USB #6 (Codec ||                         |
              +-+ |               ||              +-+ |               ||                         |
                  +---------------v+                  +---------------v+                         |
                                                                                                 |
+--------------+                    +---------------+                    +-------------+         |
|              |                    |               |                    |             |         |
|   Input      | -----------------> |               | -----------------> |             |         |
|              |                    |     FreeDV    |                    |     TRX     +---------+
|    Output    | <----------------- |               | <----------------- |             |
|              |                    |               |                    |             |
+--------------+                    +---------------+                    +-------------+

                  +---------------++                  +---------------++
                  | USB #3 (USBSo ||                  | USB #6 (Codec ||
                  |               ||                  |               ||
                  +---------------v+                  +---------------v+
´´´

tmiw:
> This is an update after testing FreeDV 1.6.2-devel-20211128-779e8f0 in 64bit version on Win10; preconditions: IC7300 hooked up, USB sound card for headset hooked up; earlier install of FreeDV on same computer available. observations:
> 
> 1. chose easy setup
> 2. Step1: select sound device shows "(multiple)", but just have IC7300; no other choice in pull-down besides "(multiple)"

Interesting, I may have missed something with how the system matches speaker and mic devices together. What does the IC-7300 look like in the current Audio Options dialog? 

> 3. "decided audio ..." and "transmitted audio..." already selected my USB sound card, ok, well done.

Yep, it will either show your already-configured options (as well as for the drop-down just above) or choose your system's default audio devices here.

> 5. click on "advanced" leads nowhere,

Oops, fixed as of 571c1b2.

> 6. click on "Test" activates TRX to start TX (why not say "test TX switch to transmit" or something?)

IIRC this used to have a longer label but it was getting truncated on Windows. I tried a few things to resolve this but didn't have luck in that regard. Help here would be appreciated for sure.

> My Idea in simple ASCII graphics was more general, maybe somehow like in the chart below.
> 
> 1. right from the start, "simple" is active, and you OP can chose one device for input/output from the pull-down menu, one for TRX, other pull-down; good suggestions for the default display are certainly a very good idea!
> 2. the lower half selections are greyed out and show the same as the upper half device selections; when changed, they change accordingly
> 3. when OP deactivates "Simple" the lower half allows other selections than the ones from above, starting with the last active selection, but now the lower pull-downs work on their own, choices are possible different to the upper half choices.
> 
> I think this is much easier. Integrating this with the Ham-Lib interface certainly is a very useful idea, too!

This might be worth coding up a prototype. Feel free to do so if you'd like and we'll take a look. 👍

cybork42:
I can confirm build conflicts, however, mine are different. I get a fail about the declaration of
 `
`  analogDevicePlaybackDeviceId_`  while compiling source `dlg_easy_setup.cpp`

tmiw:
> I can confirm build conflicts, however, mine are different. I get a fail about the declaration of ` ` analogDevicePlaybackDeviceId_`` while compiling source `` dlg_easy_setup.cpp ``
> 
> and lots of other unexpected issues (see attached file).[ [build.txt](https://github.com/drowe67/freedv-gui/files/10222323/build.txt) ](url)

Still working on it. :) Currently adapting this branch to support pipewire/PulseAudio as well as e.g. support for different sample rates for mic and speaker.

tmiw:
I was able to get it to compile and be usable on macOS tonight. Unfortunately my testing is currently limited (my primary development machine is with Apple to get repaired, so any VMs I could use to test on other platforms would be ARM only at the moment) but it should be close to back how it was when development ground to a halt.

Probably the next thing I'll do is think about serial PTT some more. Seems like it's used often enough that it can't just be relegated to the Advanced window. Some questions:

1. Which pin is most commonly used for serial PTT (RTS, DTR)?
2. Is there a commonly used polarity or does that need to be selectable as well?

tmiw:
I went ahead and just copied the entire serial PTT UI from the current PTT window. The UI looks like this now:

*No PTT/CAT Control:*

Screenshot 2022-12-15 at 12 34 32 AM

*Hamlib:*

Screenshot 2022-12-15 at 12 34 34 AM

*Serial PTT:*

Screenshot 2022-12-15 at 12 34 37 AM

I think it's okay now, especially since the non-selected options are hidden. Would appreciate thoughts on that part of the window, though.

PR #192, opened 2021-12-11
Only use device names when doing startup checks

tmiw:
Per additional debug output in #190, it looks like PortAudio outputs two error codes for the problematic sound device:

* -9997 (paInvalidSampleRate)
* -9993 (paBadIODeviceCombination)

As it's unnecessary to check all possible sample rates just to determine the device ID, this PR  ~modifies the startup check to only check against the configured sample rate~ removes the sample rate check. This should keep PortAudio from entering a state that prevents use of the configured sample rate.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3077,7 +3077,7 @@ int MainFrame::txCallback(
     return paContinue;
 }
 
-int MainFrame::getSoundCardIDFromName(wxString& name, bool input, int sampleRate)
+int MainFrame::getSoundCardIDFromName(wxString& name, bool input)
 {
     int result = -1;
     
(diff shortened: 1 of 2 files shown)


Tyrbiter:
I hacked this into the v1.6.2-test code and it seems to be working normally on my Fedora 35 system, I don't know if there is something specific I should be looking for but it starts, finds devices and works in the same way as the v1.6.2-test branch.

tmiw:
> I hacked this into the v1.6.2-test code and it seems to be working normally on my Fedora 35 system, I don't know if there is something specific I should be looking for but it starts, finds devices and works in the same way as the v1.6.2-test branch.

Thanks for testing. Yeah, it's a bit weird that `Pa_IsFormatSupported()` won't work on the device that spawned the issue/PR but opening the device and transmitting through it seems fine. 

Anyway, @drowe67, this is ready for review.

drowe67:
Well done on tracking this one down @tmiw :slightly_smiling_face: Pls feel free to merge

PR #207, opened 2022-01-16
Remove requirement to restart FreeDV after resetting configuration to defaults.

### LINKED ISSUE #202: Restore Defaults not restoring defaults on Ubuntu 20
Tested  5ecb957a on Ubuntu 20

1. Configure something (like Tools-PTT check Hamlib).
2. Tools - Restore Defaults (but keep freedv-gui running).
3. ls shows `~/.freedv` is deleted.
4. close the freedv-gui app.
5. `~/.freedv` is restored.

I think it's something to do with the [wxWidgets DeleleAll() function](https://docs.wxwidgets.org/3.0/classwx_config_base.html#af2dd5ec6b56f89decf0e96b057c4846f). This deletes the disk file, but then something in the shut down code must be reconstructing it.

tmiw:
Resolves #202 by resetting configuration options immediately upon selecting Tools->Restore Defaults. This means that the user can continue using the application after configuration reset instead of needing to restart.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -283,12 +283,6 @@ void MainFrame::loadConfiguration_()
     Fit();
     wxSize size = GetBestSize();
 
-    // Add to the "best" width and height to prevent hiding of the waterfall
-    // and the controls just below it. The values here were determined by 
-    // experimentation.
-    size.SetWidth(size.GetWidth() + 375);
-    size.SetHeight(size.GetHeight() + 100);
-
     if (w < size.GetWidth()) w = size.GetWidth();
     if (h < size.GetHeight()) h = size.GetHeight();
     SetClientSize(w, h);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -456,6 +456,9 @@ void MainFrame::loadConfiguration_()
 
     m_ckboxSNR->SetValue(wxGetApp().m_snrSlow);
     setsnrBeta(wxGetApp().m_snrSlow);
+    
+    // Show/hide frequency box based on PSK Reporter enablement
+    m_freqBox->Show(wxGetApp().m_psk_enable);
 }
 
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1093,7 +1093,7 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
         strncpy(callsign, (const char*) wxGetApp().m_callSign.mb_str(wxConvUTF8), MAX_CALLSIGN - 2);
         if (strlen(callsign) < MAX_CALLSIGN - 1)
         {
-            strncat(callsign, "\r", 1);
+            strcat(callsign, "\r");
         }     
      
         // buffer 1 txt message to ensure tx data fifo doesn't "run dry"
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -254,41 +254,11 @@ int MainApp::OnExit()
     return 0;
 }
 
-//-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
-// Class MainFrame(wxFrame* pa->ent) : TopFrame(parent)
-//-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
-MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent)
+//-------------------------------------------------------------------------
+// loadConfiguration_(): Loads or sets default configuration options.
+//-------------------------------------------------------------------------
+void MainFrame::loadConfiguration_()
 {
-    m_zoom              = 1.;
-
-    #ifdef __WXMSW__
-    g_logfile = stderr;
-    #else
-    g_logfile = stderr;
-    #endif
-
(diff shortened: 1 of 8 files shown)


src/main.cpp
```diff
@@ -581,6 +474,126 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent)
...
+    tools->AppendSeparator();
+    wxMenuItem* m_menuItemToolsConfigDelete;
+    m_menuItemToolsConfigDelete = new wxMenuItem(tools, wxID_ANY, wxString(_("&Restore defaults")) , wxT("Delete config file/keys and restore defaults"), wxITEM_NORMAL);
+    this->Connect(m_menuItemToolsConfigDelete->GetId(), wxEVT_COMMAND_MENU_SELECTED, wxCommandEventHandler(MainFrame::OnDeleteConfig));
+
+    tools->Append(m_menuItemToolsConfigDelete);
+
+    loadConfiguration_();
+    
+    if(wxGetApp().m_show_wf)
```
tmiw, line 516: @drowe67, we don't currently have these as editable within the application. Should these options just go away, or is there value in still letting people manually edit their config files to disable these tabs?

drowe67: Good point - yes it's fine to remove these options.  I think they are there for historical reasons - GUI code that was written by another author very early in the project.

tmiw: Sounds good. Removed. 👍

tmiw:
@drowe67, all good to merge?

drowe67:
Hi @tmiw - thanks for working on this, nice to clean up all that old code.

Testing on my 1366 x 789 laptop .... when I Tools-Restore defaults, the screen suddenly changes to a size far larger than my available screen resolution.  Curiously the square "maximise" icon is missing from the freedv-gui window controls in the upper RHS.  Closing and restarting the program restores the size to the maximum allowable on my display.

tmiw:
> Hi @tmiw - thanks for working on this, nice to clean up all that old code.
> 
> Testing on my 1366 x 789 laptop .... when I Tools-Restore defaults, the screen suddenly changes to a size far larger than my available screen resolution. Curiously the square "maximise" icon is missing from the freedv-gui window controls in the upper RHS. Closing and restarting the program restores the size to the maximum allowable on my display, which would perhaps be a better "default".

I pushed a minor update to save/restore the main window's position and size during the reset process. While that does mean that we effectively don't reset that, that might be okay.

drowe67:
> I pushed a minor update to save/restore the main window's position and size during the reset process. While that does mean that we effectively don't reset that, that might be okay.

Thanks, I just tried the latest. On my laptop, the window size after a restore defaults is still larger than the available display size, and square maximise icon doesn't appear to be there, until the app is restarted.

BTW is there a reason for the Report Frequency field to be red?  My eye is drawn to it.  That might be distracting for a new user, as I understand it's an optional feature.

drowe67:
> > Hi @tmiw - thanks for working on this, nice to clean up all that old code.
> > Testing on my 1366 x 789 laptop .... when I Tools-Restore defaults, the screen suddenly changes to a size far larger than my available screen resolution. Curiously the square "maximise" icon is missing from the freedv-gui window controls in the upper RHS. Closing and restarting the program restores the size to the maximum allowable on my display, which would perhaps be a better "default".
> 
> I pushed a minor update to save/restore the main window's position and size during the reset process. While that does mean that we effectively don't reset that, that might be okay.

Isn't that working against the idea of "restore defaults"?  Perhaps the window size should be set to the default size (ie the behavior after a fresh install of freedv-gui)?

tmiw:
> Thanks, I just tried the latest. On my laptop, the window size after a restore defaults is still larger than the available display size, and square maximise icon doesn't appear to be there, until the app is restarted.

Hmm, we were adding a bit extra height and width on startup, and that seemed to be interfering with the sizing on config reset. I removed that code and at least on Linux, it seems to work fine now.

> BTW is there a reason for the Report Frequency field to be red? My eye is drawn to it. That might be distracting for a new user, as I understand it's an optional feature.

Red is supposed to indicate an "invalid" value. This is important as CAT control is no longer required for PSK Reporter functionality and the latter requires a valid frequency to properly report.

drowe67:
> Hmm, we were adding a bit extra height and width on startup, and that seemed to be interfering with the sizing on config reset. I removed that code and at least on Linux, it seems to work fine now.

Thanks, that's looking much better on my laptop now - screen defaults to a usable size.
 
> > BTW is there a reason for the Report Frequency field to be red? My eye is drawn to it. That might be distracting for a new user, as I understand it's an optional feature.
> 
> Red is supposed to indicate an "invalid" value. This is important as CAT control is no longer required for PSK Reporter functionality and the latter requires a valid frequency to properly report.

Cool, yes I thought it might be something like that.  So my concern is that by highlighting that field in a default config - a new user is being drawn to/asked to understand an advanced feature (PSK reporter).  This works against our simplicity/ease of use meme. 

Would it be possible for that field to be the default color (or even not displayed), until PSK reporter has been configured and enabled?

tmiw:
> Would it be possible for that field to be the default color (or even not displayed), until PSK reporter has been configured and enabled?

Done. 👍

drowe67:
> > Would it be possible for that field to be the default color (or even not displayed), until PSK reporter has been configured and enabled?
> 
> Done. +1

Nice!  Works well for me, and makes the Main screen a little more cleaner at start up :+1:

tmiw:
> > > Would it be possible for that field to be the default color (or even not displayed), until PSK reporter has been configured and enabled?
> > 
> > 
> > Done. +1
> 
> Nice! Works well for me, and makes the Main screen a little more cleaner at start up 👍

Excellent. Anything additional needed for this PR?

drowe67:
It's great @tmiw - pls feel free to press the big green button :slightly_smiling_face:

tmiw:
> It's great @tmiw - pls feel free to press the big green button 🙂

Done 👍

PR #208, opened 2022-01-21
Add volume controls for mic in/speaker out

tmiw:
Adds volume equalizer control for mic in/speaker out in the Filter dialog per #198.

TBD:

* Desired volume gain range (currently the same as for the others, -20 to 20 dB)
* Layout/positioning of new filter control
* Possible additional cleanup of the Filter dialog

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,7 +41,9 @@ set(FREEDV_SOURCES
     sox/sox.h
     sox/sox_i.h
     sox/soxomp.h
+    sox/util.c
     sox/util.h
+    sox/vol.c
     sox/xmalloc.h
     sox/xmalloc.c
     topFrame.h
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,6 +36,9 @@ set(FREEDV_SOURCES
     sox/effects.c
     sox/effects.h
     sox/effects_i.c
+    sox/effects_i_dsp.c
+    sox/fft4g.c
+    sox/formats.c
     sox/formats_i.c
     sox/libsox.c
     sox/sox.h
--- a/src/dlg_filter.cpp
+++ b/src/dlg_filter.cpp
@@ -199,11 +199,8 @@ FilterDlg::FilterDlg(wxWindow* parent, bool running, bool *newMicInFilter, bool 
 
     wxBoxSizer* bSizer31 = new wxBoxSizer(wxHORIZONTAL);
 
-    m_sdbSizer5OK = new wxButton(this, wxID_OK);
+    m_sdbSizer5OK = new wxButton(this, wxID_CLOSE);
     bSizer31->Add(m_sdbSizer5OK, 0, wxALL, 2);
-
-    m_sdbSizer5Cancel = new wxButton(this, wxID_CANCEL);
-    bSizer31->Add(m_sdbSizer5Cancel, 0, wxALL, 2);
 
     bSizer30->Add(bSizer31, 0, wxALL | wxALIGN_CENTER, 3);
 
@@ -258,7 +255,6 @@ FilterDlg::FilterDlg(wxWindow* parent, bool running, bool *newMicInFilter, bool 
     m_SpkOutEnable->Connect(wxEVT_COMMAND_CHECKBOX_CLICKED, wxScrollEventHandler(FilterDlg::OnSpkOutEnable), NULL, this);
     m_SpkOutDefault->Connect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnSpkOutDefault), NULL, this);
 
-    m_sdbSizer5Cancel->Connect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnCancel), NULL, this);
     m_sdbSizer5OK->Connect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnOK), NULL, this);
 
 }
@@ -316,7 +312,6 @@ FilterDlg::~FilterDlg()
     m_SpkOutEnable->Disconnect(wxEVT_COMMAND_CHECKBOX_CLICKED, wxScrollEventHandler(FilterDlg::OnSpkOutEnable), NULL, this);
     m_SpkOutDefault->Disconnect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnSpkOutDefault), NULL, this);
 
-    m_sdbSizer5Cancel->Disconnect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnCancel), NULL, this);
     m_sdbSizer5OK->Disconnect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnOK), NULL, this);
 }
 
@@ -389,7 +384,7 @@ EQ FilterDlg::newEQ(wxWindow* parent, wxSizer *bs, wxString eqName, float maxFre
 //-------------------------------------------------------------------------
 // ExchangeData()
 //-------------------------------------------------------------------------
-void FilterDlg::ExchangeData(int inout, bool storePersistent)
+void FilterDlg::ExchangeData(int inout)
 {
     if(inout == EXCHANGE_DATA_IN)
     {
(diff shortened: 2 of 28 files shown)


tmiw:
@cybork42, go ahead and give this a shot? Definitely not complete but should be good as a basis for further refinement.

cybork42:
@tmiw well done! Compiled and tested quickly. Works as expected but needs more testing an da closer look at the changes; I will take about two days to finish (--> skiing weekend) and report back asap.

tmiw:
@drowe67, @cybork42, I started cleaning up the UI in the Filter dialog a bit and ended up with the following as of the latest commit:

Screen Shot 2022-01-23 at 3 09 41 PM

Questions:

1. In the original discussion thread, there was mention of potentially combining OK/Cancel into a single Close button as the adjustments happen in real time. Do we still want to do that?
2. Are the Bass/Treble/Mid controls sufficient or do we want to go more towards something like this (of course, removing the "Rock" drop down):

best-equalizer-settings-rock

tmiw:
Actually, an additional question: if we do switch to per-bin control like the second screenshot above, is there any value in also having the plot on the bottom? Or would it make sense to remove it?

tmiw:
One additional effect that may be worth including from sox is called "earwax". It's supposed to improve headphone listening per the following from the man page:

> **earwax**
>
> Makes audio easier to listen to on headphones. Adds 'cues' to 44.1kHz stereo (i.e. audio CD format) audio so that when listened to on headphones the stereo image is moved from inside your head (standard for headphones) to outside and in front of the listener (standard for speakers).

and [this page](https://web.archive.org/web/20040923193653/http://www.geocities.com/beinges/works.htm).

OTOH, given that FreeDV is inherently mono (not stereo) and encodes speech (and not music), I'm not sure this will improve things much. Comments would definitely be appreciated.

drowe67:
> @drowe67, @cybork42, I started cleaning up the UI in the Filter dialog a bit and ended up with the following as of the latest commit:

That screen shot looks quite nice on your machine @tmiw.  Yes in general I think the vertical layout is better.

However it doesn't seem to render as well on my Ubuntu machine.  For some reason the fonts/style on the EQ section appears different to the rest of the Dailog.  When the window is enlarged there is a lot of "white space".

Screenshot from 2022-01-25 06-38-34

>     1. In the original discussion thread, there was mention of potentially combining OK/Cancel into a single Close button as the adjustments happen in real time. Do we still want to do that?

Yes good idea, assuming all adjustments (not just the EQ) work in real time.
 
>     2. Are the Bass/Treble/Mid controls sufficient or do we want to go more towards something like this (of course, removing the "Rock" drop down smile ):

I don't have a strong opinion either way.  What benefit would such a change have? To move from a parametric to graphic equaliser would require a rewrite of all the DSP.  Is this effort warranted?

drowe67:
> One additional effect that may be worth including from sox is called "earwax". It's supposed to improve headphone listening per the following from the man page:

We need to listen to the effect to determine if it has any value. This could be achieved using the command line codec2/freedv and sox tool son a few samples.  We did evaluate some effects from the Argentinian FreeDV community a  few months ago, but I think we struggled to show  improvements.

To improve the speech quality I'd be inclined to put more work into the codec, for example https://github.com/drowe67/codec2/discussions/283

tmiw:
> That screen shot looks quite nice on your machine @tmiw. Yes in general I think the vertical layout is better.
> 
> However it doesn't seem to render as well on my Ubuntu machine. For some reason the fonts/style on the EQ section appears different to the rest of the Dailog. When the window is enlarged there is a lot of "white space".

I checked in some tweaks that will make the font match the rest of the window. I also made some minor tweaks to spacing I ran out of time. I'll look into the rest of the concerns tonight.

tmiw:
@drowe67, I believe I've resolved all the spacing concerns now:

Screen Shot 2022-01-26 at 12 47 09 AM

Let me know if anything else needs to be adjusted.

tmiw:
> > One additional effect that may be worth including from sox is called "earwax". It's supposed to improve headphone listening per the following from the man page:
> 
> We need to listen to the effect to determine if it has any value. This could be achieved using the command line codec2/freedv and sox tools on a few samples. We did evaluate some effects from the Argentinian FreeDV community a few months ago, but I think we struggled to show improvements.
> 
> To improve the speech quality I'd be inclined to put more work into the codec, for example [drowe67/codec2#283](https://github.com/drowe67/codec2/discussions/283)

I haven't listened with headphones yet but I do notice a difference between earwax and not-earwax. The commands I used to generate the files are the following:

```
mooneer@ubuntu:~/freedv-gui/codec2/build_linux$ src/freedv_tx 700D ../raw/ve9qrp.raw tx.raw --clip 1 --txbpf 1
mooneer@ubuntu:~/freedv-gui/codec2/build_linux$ src/freedv_rx 700D tx.raw rx.raw
frames decoded: 702  output speech samples: 898560
mooneer@ubuntu:~/freedv-gui/codec2/build_linux$ sox -b 16 -r 8000 -c 1 -t raw -e signed-integer rx.raw ve9qrp_earwax.wav rate 44100 channels 2 earwax
mooneer@ubuntu:~/freedv-gui/codec2/build_linux$ sox -b 16 -r 8000 -c 1 -t raw -e signed-integer rx.raw ve9qrp.wav rate 44100 channels 2
mooneer@ubuntu:~/freedv-gui/codec2/build_linux$ 
```

Unfortunately I'm having a hard time attaching the output here, so I'll email to you.

drowe67:
Thanks of those sample @tmiw.  Yes it's a pity we can't paste `.wav` files into GitHub.  For comparing two processing methods it's best to use short segments, e.g. 3-10s.  I listened to the first 10s of each for comparison. The earwax version seemed a lot quieter, and had less high freq content, it didn't think it was better than the non-earwax version.  No major difference in quality between them.  What do you think?

cybork42:
@tmiw and @drowe67: lots of things happened on this thread and I failed to keep up, sorry about that. Other topics pulling attention. 

- **equalizer**: I believe most users are hams with some experience about filters and work well specifying "Q" for a filter. This is also well supported by the responsive filter curve where you see what's happening. A graphic equalizer is more well-known, but from experience you end up trying weird stuff with the sliders there, and the filter curves fail to follow the displayed settings. I think the "tech style" with frequency, level and "Q" are Ok. In another way they are also more portable from one computer to another (numeric settings are a fine reference!).

- **earwax**: I didn't try and I don't like. Why? It blurs and covers the bare-bone function of FreeDV's audio path. I mean it's Ok to adjust to a speaker's modulation so the algorithms work well. But the output should be as much as purce from decoding as possible. It's another story when you can switch it on/off completely, as is with the mic and speaker equalizer. I don't use the speaker equalizer tab either (different when listening to radio streams on the PC).

- the left microphone in Equalizer control should probably better be named "Common" rather than "Vol"? I mean, technically a microphone doesn't have a "volume"... It's great you can see the consequence immediately with the filter curve shape!

- **microphone gain settings**: I still like the dialog a lot; maybe some minor optimization possible: pull the "Enable" checkbuttons out to the left so they receive more visual attention  (="click here first before anywhere else", esp. because the "child" controls get greyed out after disabling). 
Maybe indent the "Speex..." and "700c/700d" to the right? The texts are longs anyway, maybe with an indented vertical sizer rather than a horizontal one? And once they align vertically, maybe just the same for the controls of the "1600 LPS Post Filter", so "Enable" aligns horizontally with a vertical sizer that includes (0-1Khz, Beta and Gamma)?

- the new modality of the window with "Close" Button only is great: you can have it open (and the apply button is not nedeed, which was a worry in my earlier post), that's wonderful!

In all: very good work!

drowe67:
> @drowe67, I believe I've resolved all the spacing concerns now:
> Let me know if anything else needs to be adjusted.
Screenshot from 2022-01-27 06-32-23

Yes that's much nicer :+1:   One minor question - is the font size on the tabs at the bottom (e.g. "Microphone in Equaliser") seems small for me.  But this must be just "how it is" for wxWidgets  - the font size on the main window tabs is also a bit smaller than the other text.

tmiw:
> Thanks of those sample @tmiw. Yes it's a pity we can't paste `.wav` files into GitHub. For comparing two processing methods it's best to use short segments, e.g. 3-10s. I listened to the first 10s of each for comparison. The earwax version seemed a lot quieter, and had less high freq content, it didn't think it was better than the non-earwax version. No major difference in quality between them. What do you think?

Without headphones, I think I can definitely see why people think earwax makes the sound seems like it's farther back. It's much harder to tell with headphones (albeit I was using a Bluetooth headset). I'm not sure it adds anything to intelligibility in a quiet environment, though it may be different outside or in a generally noisier environment.

> Yes that's much nicer 👍 One minor question - is the font size on the tabs at the bottom (e.g. "Microphone in Equaliser") seems small for me. But this must be just "how it is" for wxWidgets - the font size on the main window tabs is also a bit smaller than the other text.

It's actually something that was in the previous code. I removed it so the font should look more like the default now.

> * **earwax**: I didn't try and I don't like. Why? It blurs and covers the bare-bone function of FreeDV's audio path. I mean it's Ok to adjust to a speaker's modulation so the algorithms work well. But the output should be as much as purce from decoding as possible. It's another story when you can switch it on/off completely, as is with the mic and speaker equalizer. I don't use the speaker equalizer tab either (different when listening to radio streams on the PC).

It would definitely be something that can be enabled or disabled if it were to be adopted. In fact, it *wouldn't* be a good idea to always have it on as it's intended for use with headphones.

> * the left microphone in Equalizer control should probably better be named "**Common**" rather than "Vol"? I mean, technically a microphone doesn't have a "volume"... It's great you can see the consequence immediately with the filter curve shape!

Maybe "level"?

> * **microphone gain settings**: I still like the dialog a lot; maybe some minor optimization possible: pull the "Enable" checkbuttons out to the left so they receive more visual attention  (="click here first before anywhere else", esp. because the "child" controls get greyed out after disabling).

It seems like the current position is still okay because people would go from top to bottom. @drowe67, thoughts?

drowe67:
> > * the left microphone in Equalizer control should probably better be named "**Common**" rather than "Vol"? I mean, technically a microphone doesn't have a "volume"... It's great you can see the consequence immediately with the filter curve shape!
> 
> Maybe "level"?

I quite like Vol, as that will be understood quickly by most people, despite not being technically correct.  Gain is another possibility. Yes I agree it's function is also very clear by what happens to the plot :heavy_check_mark: 

> > * **microphone gain settings**: I still like the dialog a lot; maybe some minor optimization possible: pull the "Enable" checkbuttons out to the left so they receive more visual attention  (="click here first before anywhere else", esp. because the "child" controls get greyed out after disabling).
> 
> It seems like the current position is still okay because people would go from top to bottom. @drowe67, thoughts?

I have a minor preference for having the "Enable" checkbuttons out to the left.  But happy to leave it up to you guys :slightly_smiling_face:

tmiw:
The Enable checkboxes and Default buttons should now be on the left. 👍

cybork42:
I tend to agree (strongly :-)).

drowe67:
Thanks for your work on this one @tmiw - yes I'm happy for you to merge.

PR #211, opened 2022-02-08
[Experimental] 2020A/B support

tmiw:
This is a temporary PR to support testing of 2020A and B modes. See https://github.com/drowe67/codec2/pull/274 for more details.


src/freedv_interface.cpp
```diff
@@ -93,9 +121,21 @@ void FreeDVInterface::start(int txMode, int fifoSizeMs, bool singleRxThread, boo
     for (auto& mode : enabledModes_)
     {
         struct freedv* dv = nullptr;
-        if ((mode == FREEDV_MODE_700D) || (mode == FREEDV_MODE_700E) || (mode == FREEDV_MODE_2020)) {
+        if ((mode == FREEDV_MODE_700D) || (mode == FREEDV_MODE_700E) || (mode == FREEDV_MODE_2020) ||
+            (mode == FREEDV_MODE_2020A) || (mode == FREEDV_MODE_2020B)) {
             // 700 has some init time stuff so treat it special
```
drowe67, line 126: I don't think comment applies any more...

src/freedv_interface.cpp
```diff
@@ -93,9 +121,21 @@ void FreeDVInterface::start(int txMode, int fifoSizeMs, bool singleRxThread, boo
...
+            {
+                // Set index optimized VQ to 1 (required for 2020A/B).
+                adv.lpcnet_vq_type = 1;
+            }
+            else if (mode == FREEDV_MODE_2020)
+            {
+                adv.lpcnet_vq_type = 0;
+            }
+            
             dv = freedv_open_advanced(mode, &adv);
```
drowe67, line 139: Suggest we just use `freedv_open()` for all modes - that will set up index optimisation correctly.  FYI using **0** is a predictive quantiser, where we quantise the differences from the last frame.  Efficient, but bit errors tend to propagate so it degrades badly in the sort of error prone channels we use for HF digital voice. Vanilla 2020 currently uses **1**, but no need to expose all this at the `freedv-gui` level.

src/freedv_interface.cpp
tmiw: Deleted if block per below.

src/freedv_interface.cpp
tmiw: Updated accordingly. 👍

drowe67:
Hi @tmiw - builds well for me using `build_linux.sh`, thanks.  

Now I need to figure out how to test it :thinking:   Some ideas:

1. For 2020, check effect of clipping, make sure clipping checkbox is doing something.  Scatter diagram should get noisier in loopback.  OTA it should improve Rx SNR.
1. For 2020A and B, perhaps test with canned test files created with `freedv_tx`.  
1. Perhaps try canned test files with channel modelling to make sure it's hanging on Ok, e.g. 2020B should cope with `ch ... --mpd`, but 2020 and 2020A should do pretty badly.

Clipping should be on for 2020A/B as well, to get the full benefit.

drowe67:
Ok I've pushed a few new sample files.  They work pretty well, so I'm pretty sure the Rx side of the new modes is working as advertised.

drowe67:
OK I've changed the 2020A Unique Word to make 2020 and 2020A less likely to sync on each other.  You'll need to `git pull && make` on codec2.

For some reason the Sync  green light isn't going out though, I might need to tweak the UW thresholds in `libcodec2`.

Would it be OK to disable the default multirx-on option in this branch please?  I was accidentally testing with it (and it didn't seem to select the correct 2020 version).

tmiw:
> OK I've changed the 2020A Unique Word to make 2020 and 2020A less likely to sync on each other. You'll need to `git pull && make` on codec2.
> 
> For some reason the Sync green light isn't going out though, I might need to tweak the UW thresholds in `libcodec2`.
> 
> Would it be OK to disable the default multirx-on option in this branch please? I was accidentally testing with it (and it didn't seem to select the correct 2020 version).

2020A and B shouldn't be part of multi-RX now regardless of how it's set, similar to 800XA and 2400B. As long as 2020A or B are selected before pushing Start, it won't decode any others. If that's not actually the case, we should fix that for sure.

Tyrbiter:
Tried to build this branch using build_linux.sh, for some reason the correct codec2 and LPCNet branches are not being pulled.

Errors are:

+ git checkout dr-2020A
error: pathspec 'dr-2020A' did not match any file(s) known to git


+ git checkout dr-vq-index-opt
error: pathspec 'dr-vq-index-opt' did not match any file(s) known to git

I fixed this by running a git pull --all in each sub-directory after which the branches do show up.

Just thought it worth mentioning.

drowe67:
> Tried to build this branch using build_linux.sh, for some reason the correct codec2 and LPCNet branches are not being pulled.

Hmm, that's curious.  I just tried a fresh checkout and it worked just fine, as did the original build.

drowe67:
@tmiw OK I tried this with the ALSA loopback module, and could see the scatter diagram change when the clipping checkbox was toggled.  I've bumped the version number to make it easy to differentiate from 1.7.0

So If you are OK with this PR I'm happy for this to be released for wider use by Alpha testers.  When convenient for you, could you please generate some Windows installer files? That will give us about a week before the Activity day to see if there are any major bugs. I haven't tried to run it on Windows myself, just Linux, so a first step will be to see if it actually works on Windows.

I'd prefer not to merge until we are sure the new modes are useful.

tmiw:
@drowe67, I just generated the Windows and macOS builds and emailed out to the two mailing lists. 👍

drowe67:
> @drowe67, I just generated the Windows and macOS builds and emailed out to the two mailing lists. +1

Thanks @tmiw :grin:

PR #214, opened 2022-02-25
PSK Reporting enhancements

tmiw:
Updates PSK Reporter feature as follows:

1. Encodes callsign regardless of whether the internet is working.
2. Sends report upon pushing Stop (vs. simply clearing the report list).
3. Performs reporting in background instead of hanging the caller of the PskReporter class.
4. Suppress reporting if we're playing back a radio file (to avoid false reports).


tmiw:
@drowe67, looks like this was already merged into #211 but another test build including this won't hurt.

tmiw:
I haven't heard any specific complaints about the PSK Reporter related changes, so this is probably ready to merge. @drowe67, agree?

drowe67:
Sure pls merge @tmiw

PR #215, opened 2022-02-26
Fix spelling errors

dforsi:
This PR fixes some small typos in a string, in the documentation and in comments.
The only change in the code is in the string is "locahost" corrected as "localhost".

Maybe you could add codespell to Travis?
This command would return a non-zero exit code if there is a new error or a new word to ignore:
`codespell --ignore-words-list=inout,nin,ontop,parm,tthe,ue`

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -291,11 +291,12 @@ hardware, what you have tried, and the exact nature of the problem.
 
 ## FreeDV 2020 mode is greyed out
 
-You must have a modern CPU with AVX support to run FreeDV 2020.  If
-you do not have AVX the FreeDV 2020 mode button will be greyed out.
-A Microsoft utility called [coreinfo](https://docs.microsoft.com/en-us/sysinternals/downloads/coreinfo)
-can be also used to determine if your CPU supports AVX.  A * means
-you have AVX, a - means no AVX:
+In order to use FreeDV 2020 mode, you must have one of the following:
+
+1. An Intel based CPU with AVX support. A Microsoft utility called [coreinfo](https://docs.microsoft.com/en-us/sysinternals/downloads/coreinfo)
+can be used to determine if your CPU supports AVX.  A * means you have 
+AVX, a - means no AVX:
+
 ```
 AES             -       Supports AES extensions
 AVX             *       Supports AVX instruction extensions
@@ -308,6 +309,10 @@ or the output of the `lscpu` command:
 lscpu | grep -o "avx[^ ]*"
 ```
 will display `avx` (or `avx2`) if your CPU supports the instructions.
+
+2. A Mac with an ARM processor (e.g. 2020 Mac Mini or later).
+
+If your system does not meet either (1) or (2), the 2020 option will be grayed out.
 
 ## FreeDV 2020 mode is slow on ARM Macs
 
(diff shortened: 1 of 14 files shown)


USER_MANUAL.md
```diff
@@ -293,12 +293,12 @@ hardware, what you have tried, and the exact nature of the problem.
 
 You must have a modern CPU with AVX support to run FreeDV 2020.  If
```
tmiw, line 294: Hmm, now that I'm looking at this again, perhaps we should update this paragraph to indicate that the M1 Macs can also run 2020?

USER_MANUAL.html
```diff
@@ -349,8 +349,8 @@ <h2 id="tools---filter"><span class="header-section-number">9.1</span> Tools - F
 </thead>
 <tbody>
 <tr class="odd">
-<td>Noise Supression</td>
-<td>Enable noise supression, dereverberation, AGC of mic signal using the Speex pre-processor</td>
+<td>Noise Suppression</td>
+<td>Enable noise suppression, dereverberation, AGC of mic signal using the Speex pre-processor</td>
 </tr>
 <tr class="even">
 <td>700C/700D Auto EQ</td>
```
drowe67, line 356: The HTML version of the user manual is automatically generated from the Markdown version so best not to modify it directly.

drowe67:
Thanks @dforsi :slightly_smiling_face: 

> Maybe you could add codespell to Travis? This command would return a non-zero exit code if there is a new error or a new word to ignore: `codespell --ignore-words-list=inout,nin,ontop,parm,tthe,ue`

That's an interesting idea.  So what would happen if spelling errors where found?  We could also run `codespell` as a ctest - that would let us check spelling before pushing.  It's often more convenient to catch issues before a push.

dforsi:
The `build_linux.sh` script doesn't run the tests and you can't be sure that contributors run them, so you probably need both `ctest` for convenience and CI for completeness.

If we could be sure that a spelling error is in the GUI or in the documentation I would say to fail the build. Since this isn't possible but the errors are quick to fix you may still fail the build, however having CI errors when committing to master can be confusing for third parties, it suits better when using a develop branch or topic branches.

USER_MANUAL.md
tmiw: Updated this section. Feel free to change if needed.

USER_MANUAL.html
tmiw: I just pushed an update to USER_MANUAL.md so that should correct itself hopefully.

tmiw:
> The `build_linux.sh` script doesn't run the tests and you can't be sure that contributors run them, so you probably need both `ctest` for convenience and CI for completeness.
> 
> If we could be sure that a spelling error is in the GUI or in the documentation I would say to fail the build. Since this isn't possible but the errors are quick to fix you may still fail the build, however having CI errors when committing to master can be confusing for third parties, it suits better when using a develop branch or topic branches.

I feel like only updating GitHub Actions is sufficient as long as we don't allow merging unless all checks pass. It doesn't necessarily matter that something in a fork isn't spelled correctly, just that there's no obvious issues in the official repo.

drowe67:
@dforsi @tmiw - yes running the spell check in the GitHub actions sounds like a fine approach for `freedv-gui`.  I might add a `ctest` for spelling to `codec2`.

tmiw:
> @dforsi @tmiw - yes running the spell check in the GitHub actions sounds like a fine approach for `freedv-gui`. I might add a `ctest` for spelling to `codec2`.

I created #216 to add the codespell run. It's failing right now since this PR hasn't been merged yet, of course. Results [here](https://github.com/drowe67/freedv-gui/runs/5372206934?check_suite_focus=true).

drowe67:
@tmiw I'm ok for you to merge this when you are ready

PR #220, opened 2022-03-25
Build Windows builds for PRs

tmiw:
Just an experiment for now. We should be able to build Windows ~and macOS~ binaries when creating PRs to ensure there's no OS-specific compiler issues. (In the past, the GitHub actions have occasionally passed while MinGW also spit out errors.)

EDIT: macOS builds currently take too long to be practical for PRs, so this PR is just to add Windows builds.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -1,4 +1,4 @@
-name: Build FreeDV
+name: Build FreeDV (Linux)
 
 on:
   push:
--- a/.github/workflows/cmake-windows.yml
+++ b/.github/workflows/cmake-windows.yml
@@ -1,4 +1,4 @@
-name: Build FreeDV (Linux)
+name: Build FreeDV (Windows)
 
 on:
   push:
--- a/.github/workflows/cmake-windows.yml
+++ b/.github/workflows/cmake-windows.yml
@@ -1,4 +1,4 @@
-name: Build FreeDV (macOS)
+name: Build FreeDV (Linux)
 
 on:
   push:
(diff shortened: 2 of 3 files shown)


tmiw:
So I'm not sure I want to keep the macOS one here given how long it normally takes to build on my laptop. We'll see if it passes without needing to do anything regardless. Alternatively, we can maybe set it up to only get executed when merging PRs to master?

drowe67:
> So I'm not sure I want to keep the macOS one here given how long it normally takes to build on my laptop. We'll see if it passes without needing to do anything regardless. Alternatively, we can maybe set it up to only get executed when merging PRs to master?

Yes I think that's possible with GitHub actions.

tmiw:
> > So I'm not sure I want to keep the macOS one here given how long it normally takes to build on my laptop. We'll see if it passes without needing to do anything regardless. Alternatively, we can maybe set it up to only get executed when merging PRs to master?
> 
> Yes I think that's possible with GitHub actions.

Hmm, Homebrew doesn't seem to support universal binaries anyway. Combined with the GitHub macOS runners being x86 only and "self-hosted" runners [being a security nightmare](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security), I'm fine just doing Windows and Linux builds. In my experience, if the Linux build fails, the macOS one will too (unless it's something PulseAudio-related).

drowe67:
> Hmm, Homebrew doesn't seem to support universal binaries anyway. Combined with the GitHub macOS runners being x86 only and "self-hosted" runners [being a security nightmare](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security), I'm fine just doing Windows and Linux builds. In my experience, if the Linux build fails, the macOS one will too (unless it's something PulseAudio-related).

Agreed - we seem to be doing the cross platform thing pretty well :+1:

tmiw:
> > Hmm, Homebrew doesn't seem to support universal binaries anyway. Combined with the GitHub macOS runners being x86 only and "self-hosted" runners [being a security nightmare](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security), I'm fine just doing Windows and Linux builds. In my experience, if the Linux build fails, the macOS one will too (unless it's something PulseAudio-related).
> 
> Agreed - we seem to be doing the cross platform thing pretty well 👍

👍 

Anyway, I can merge up if there aren't any issues here?

drowe67:
Sure, all good to merge @tmiw :+1:

Issue #232, opened 2022-04-30
when cloning a certain branch...


cybork42:
... github will provide files from right taht branch, but all build scripts will still default to use the default repositories, which defaults to `master`.
Maybe it makes sense to enhance the intelligence and awareness of build scripts to look into the `.git` directory to detect the cloned branch and use that branch for consecutive cloning?

tmiw:
Copying previous text from the just-closed codec2 issue:

> Maybe this is no real issue, but rather an imrpovement suggestion, or even a nonsense idea. When you execute
>
> `git clone --branch ms-2020-experiment https://github.com/drowe67/freedv-gui.git`
>
> you get what you expect, all source files from that named branch. However and naturally, the build scripts know nothing about being in a branch and still say
>
> `git clone https://github.com/drowe67/freedv-gui.git`
>
> An improvement suggestion might be to have the build scripts look at the .git file and identify which branch was requested and rather use files from that branch rather than master when cloning anew?

tmiw:
@cybork42, I took a look at the build scripts in the current master (`build_linux.sh`, `build_osx.sh` and `build_windows.sh`) as well as the Docker scripts and I don't see anything that's specifically cloning freedv-gui again. Well, other than [this Docker script](https://github.com/drowe67/freedv-gui/blob/master/docker/fdv_win_fedora/entrypoint.sh), but that takes the desired freedv-gui branch directly from `freedv_build_windows.sh` (as provided by the user using the `--branch` command line argument, or master if none provided).

Just so I'm sure we're on the same page, were you thinking that `freedv_build_windows.sh` would be modified to always default `--branch` to the current branch? Or did I miss some other git clone reference in the current scripts?

cybork42:
@tmiw: you are right, in fact the build script for docker was the one that made me stumble about this issue, and it's the only one script that uses some "branch" arguments.
When one performs a `docker-compose -f docker-compose-win.yml build`, you would expect that all later builds will really use the branch cloned to local storage without mentioning anything special. The additional arguments to `./freedv_build_windows.sh` are not really obvious, are they?
My suggestion would be to do the following change:

```
FDV_GIT_REPO=https://github.com/drowe67/freedv-gui.git
# FDV_GIT_BRANCH=master
FDV_GIT_BRANCH=`git branch | cut -c 3-`
echo "building branch $FDV_GIT_BRANCH"
FDV_BOOTSTRAP_WX=0
```

tmiw:
[...]

PR #233, opened 2022-05-02
Default repo and branch to the current one when using the Docker script.

tmiw:
Updates the Docker `freedv_build_windows.sh` script to retrieve the current branch and repo URL and use those for defaults instead of master/the main repo. Should be useful for those who are forking the repo to make testing a bit easier.


tmiw:
@cybork42, re: #232--does this work for you? It's a minor change but it should still hopefully make things a bit easier when making Windows builds.

cybork42 (approved):
looks even better than my proposal, well done!

PR #234, opened 2022-05-02
Fix typos in code that saves sample rates.

tmiw:
Thanks Alan for discovering this bug during testing on Fedora. Turns out it was a missed copy-paste edit, hence the sample rates being set to -1 in certain cases.


tmiw:
@drowe67, looks okay?

PR #235, opened 2022-05-02
Add missing required gcc-g++ package to Fedora build instructions.

tmiw:
Discovered that `gcc-g++` isn't installed on Fedora 36 when using `sudo yum groupinstall 'Development Tools'`. To be safe, I added a specific reference to that package in `README.md`.


tmiw:
@Tyrbiter, I thought that `gcc-g++` was already in that Yum group but I guess not. The current instructions work for me now but I'd appreciate a confirmation if you get a chance. 👍

Tyrbiter:
I have just had a look at this. I know what I have here, I hope that my setup is not broken although I don't think so as all the builds of freedv and related code (and indeed everything else) work reliably.

yum is quite old and has been superseded by dnf for quite a few Fedora releases, so the command should be:

'sudo dnf install' with the group and specific packages as shown previously but with one exception. There is no gcc-g++ package any more, if I try to install it I get:

Package gcc-c++-12.0.1-0.16.fc36.x86_64 is already installed.

So you need to change the gcc-g++ to gcc-c++ to be sure it will work now and in future.

I hope that is all that's needed.

tmiw:
> I have just had a look at this. I know what I have here, I hope that my setup is not broken although I don't think so as all the builds of freedv and related code (and indeed everything else) work reliably.
> 
> yum is quite old and has been superseded by dnf for quite a few Fedora releases, so the command should be:
> 
> 'sudo dnf install' with the group and specific packages as shown previously but with one exception. There is no gcc-g++ package any more, if I try to install it I get:
> 
> Package gcc-c++-12.0.1-0.16.fc36.x86_64 is already installed.
> 
> So you need to change the gcc-g++ to gcc-c++ to be sure it will work now and in future.
> 
> I hope that is all that's needed.

Thanks! I wasn't aware that `yum` is old/deprecated (I haven't used Fedora for much in quite a while, I typically use Debian-based distros). I just corrected `gcc-c++` in the README.

Anyway, I'll leave this for a day or two in case there's anything else glaringly missing before merging. 👍

Issue #149, opened 2021-07-15
Error in compiling FreeDV on ARM64 on Raspberry PI


UgoPoddine:
Hi everyone,

I'm trying to build, using the usual script build_linux.sh, FreeDV-GUI over an ARM64 image on RaspberryPI3+
FreeDV GUI last commit (937e8be).
Linux 5.10.17-v8+ #1421 SMP PREEMPT Thu May 27 14:01:37 BST 2021 aarch64 GNU/Linux
The building process abends as follow.

Can you help ? thank-you very much
Best regards
Ugo

[...]

[  3%] Building CXX object src/CMakeFiles/freedv.dir/dlg_audiooptions.cpp.o
cd /home/pi/FreeDV/freedv-gui/build_linux/src && /usr/bin/c++  -DGIT_HASH=\"937e8be\" -DWXUSINGDLL -D_FILE_OFFSET_BITS=64 -D__WXGTK__ -I/home/pi/FreeDV/freedv-gui/src -I/home/pi/FreeDV/freedv-gui/build_linux/src -I/home/pi/FreeDV/freedv-gui/build_linux -isystem /usr/lib/arm-linux-gnueabihf/wx/include/gtk2-unicode-3.0 -isystem /usr/include/wx-3.0 -isystem /home/pi/FreeDV/freedv-gui/codec2/src -isystem /home/pi/FreeDV/freedv-gui/codec2/build_linux -isystem /home/pi/FreeDV/freedv-gui/LPCNet/src  -Wall -pthread -g   -o CMakeFiles/freedv.dir/dlg_audiooptions.cpp.o -c /home/pi/FreeDV/freedv-gui/src/dlg_audiooptions.cpp
In file included from /home/pi/FreeDV/freedv-gui/src/main.h:33,
                 from /home/pi/FreeDV/freedv-gui/src/dlg_audiooptions.cpp:22:
/usr/include/wx-3.0/wx/filename.h: In static member function \u2018static wxUniChar wxFileName::GetPathSeparator(wxPathFormat)\u2019:
/usr/include/wx-3.0/wx/filename.h:473:43: error: ambiguous overload for \u2018operator[]\u2019 (operand types are \u2018wxString\u2019 and \u2018unsigned int\u2019)
         { return GetPathSeparators(format)[0u]; }
                                           ^
/usr/include/wx-3.0/wx/filename.h:473:43: note: candidate: \u2018operator[](const wchar_t*, long int)\u2019 <built-in>
[...]

PR #236, opened 2022-05-08
Rearrange shutdown actions to resolve hangs on slower machines

tmiw:
On some slower systems (for example, the Raspberry Pi 3), enabling multiple receive and/or using 700C results in the application hanging after pressing Stop. This is because the TX and RX threads can queue up significant numbers of audio signals while in the process of shutting down, preventing the threads from ever exiting. To resolve this, the sound devices are closed before the TX/RX threads, ensuring that the latter doesn't continue to have audio data queued up.

Thanks to @UgoPoddine for the assistance in diagnosing the issue (see comments beginning [here](https://github.com/drowe67/freedv-gui/issues/149#issuecomment-1113428373) for context).


tmiw:
@drowe67, as a note, 700C seems to need around 6-7 seconds to decode a 10 second file that doesn't contain valid 700C audio (on the Raspberry Pi 3). This was tested using `time ./freedv_rx 700C ve9qrp_10s.raw rx.raw`. In short, still a bit faster than real time but not by much. I'm not sure how much more it can really be optimized without major changes.

drowe67:
> @drowe67, as a note, 700C seems to need around 6-7 seconds to decode a 10 second file that doesn't contain valid 700C audio (on the Raspberry Pi 3). This was tested using `time ./freedv_rx 700C ve9qrp_10s.raw rx.raw`. In short, still a bit faster than real time but not by much. I'm not sure how much more it can really be optimized without major changes.

Yes acquisition is pretty heavy in 700C. Is it vectorised for the Pi?  I'm not sure any serious optimisation effort is warranted - Pi 3 is a pretty old legacy platform with very few users.

tmiw:
> > @drowe67, as a note, 700C seems to need around 6-7 seconds to decode a 10 second file that doesn't contain valid 700C audio (on the Raspberry Pi 3). This was tested using `time ./freedv_rx 700C ve9qrp_10s.raw rx.raw`. In short, still a bit faster than real time but not by much. I'm not sure how much more it can really be optimized without major changes.
> 
> Yes acquisition is pretty heavy in 700C. Is it vectorised for the Pi? I'm not sure any serious optimisation effort is warranted - Pi 3 is a pretty old legacy platform with very few users.

When I was debugging on @UgoPoddine's Pi, he was running the 64 bit version of Raspberry Pi OS, which enabled NEON in Codec2 without needing to do anything extra. I imagine it wouldn't be able to run 700C at all without it.

UgoPoddine:
Hello mates,

In any case thank-you for your efforts, really appreciated.
Please consider that :

A) on 32bits Raspberry Pi3 (the same machine), history has been troubled but successful. In the first years, the GUI itself was a bit too slow for the pi and we were forced to use a small home made codec2 wrapper to use, mainly in 700C mode. A couple of years ago for a certain period everything was ok with GUI and 700C, after this we have had a period in which 700C was unable to stop like now. New release solved the issue. In the last month issue was again there on 32bit on 700C
B) on the new 64bits OS we have had what described: first versions of OS show problems in compiling the GUI due to wxwidget mismatch, solved with new versions of the OS.
C) regarding hardware age, please consider that the PI3 is a 5years aged 4CPU machine, compatible with 64bits os, NEON ready, well supporting the most common digital modes tools (Fldigi, wsjt-x, JS8call, FreeDATA..) and that a lot of hams use it extensively in their stations, as it was the first well performing "pocket" PC.
Also considering the low rotation of hardware on hams' shacks and the increasing price of PI4, not sure that it can be really considered in end of life" status, but of course it's up to you 

Thank-you again
73, Ugo


Il giorno 8 mag 2022, 07:58, alle ore 07:58, Mooneer Salem ***@***.***> ha scritto:

tmiw:
So yeah, what we're officially supporting is probably more of a long-term discussion so we'll table that for now.

Anyway, I'm going to generate a test build to email out but how well is this fix working for you, @UgoPoddine? Just want to confirm before I close out this PR.

drowe67:
> So yeah, what we're officially supporting is probably more of a long-term discussion so we'll table that for now.

@tmiw I was thinking along similar lines - it would be good to document what we platforms support (same for codec2).  Maybe with some numbers (e.g. CPU load metrics).

tmiw:
> > So yeah, what we're officially supporting is probably more of a long-term discussion so we'll table that for now.
> 
> @tmiw I was thinking along similar lines - it would be good to document what platforms we support (same for codec2). Maybe with some numbers (e.g. CPU load metrics).

Based on my experience with @UgoPoddine's rPi3, I suspect that device will be the ultimate minimum in terms of system requirements for freedv-gui. However, as most users use x86 machines, it would be good to get a baseline minimum on that platform too. (FWIW, FreeDV seems to run fine on my 2011 MacBook Air with Windows 10 as well as a Microsoft Surface Go 2 tablet, but IIRC both are 64-bit devices. Anything that doesn't support AMD64 may be a different story.)

OTOH, Codec2 is going to be much more lenient. After all, it can run a subset of the available modes on the STM32 no problem. 

(Speaking of the STM32, I did find out recently that the ESP32-S3 can't handle the BPF right now; 700D TX with BPF enabled takes slightly more than 160ms to run, guaranteeing dropouts. It is close enough that optimizations there and/or elsewhere in the TX chain could make it work. Without BPF, the TX time is more like 125ms per output block, more than enough to avoid dropouts.)

drowe67:
@tmiw - there might be other, more efficient, ways to implement the Tx BPF, for example an IIR filter.  Alternative implementations would need to be PER tested.

When I first released 700D, some of the testers were running crusty old XP machines.  Hams tend to keep old computers, part of the demographic I guess.  I'm not sure that means we have to support them.  It's hard to set a numerical requirement, as the mileage will vary from machine to machine.

I would agree that the typical use case is a 10 year old x86 desktop.  Using Pi's as Desktops is probably less common, they are more often used headless.

tmiw:
> @tmiw - there might be other, more efficient, ways to implement the Tx BPF, for example an IIR filter. Alternative implementations would need to be PER tested.

Honestly, I was going to offload the BPF to the TLV320 audio codec chip as it has built-in filtering functionality:

Screen Shot 2022-05-16 at 12 32 32 PM

Low priority, though :)

> I would agree that the typical use case is a 10 year old x86 desktop. Using Pi's as Desktops is probably less common, they are more often used headless.

👍

drowe67:
Interesting idea.  The problem with offloading from DSP to hardware is it can't be (c)tested.  I have some mild concerns about the amplitude and phase distortion of tight filters, e.g. carriers at the edges could be attenuated more than those in the centre.  If the phase distortion is gradual, it should be OK, as we have phase estimation for each carrier.

tmiw:
I haven't heard of any issues due to this PR, so I'm going to go ahead and merge.

PR #241, opened 2022-06-12
Add sox as a required package for Fedora.

tmiw:
While testing with Fedora Rawhide, I found that the compile would fail with the provided README instructions due to `sox` not being on the system. This PR adds it as a required dependency to avoid compiler errors.


tmiw:
@Tyrbiter, I *think* the instructions are fine now but confirmation would be good.

BTW should I remove the PortAudio instructions for Fedora given that it's basically pipewire only now?

Tyrbiter:
OK, I have looked at these changes and I can't see any problems with the overall instructions and necessary packages.

I build freedv using separate rpm builds for codec2 and lpcnet as set up by Richard Shaw KF5OIM, I have now added a Requires: entry for sox into the freedv.spec file I use. I will add a request to make this change in the Redhat bugzilla.

tmiw:
> OK, I have looked at these changes and I can't see any problems with the overall instructions and necessary packages.
> 
> I build freedv using separate rpm builds for codec2 and lpcnet as set up by Richard Shaw KF5OIM, I have now added a Requires: entry for sox into the freedv.spec file I use. I will add a request to make this change in the Redhat bugzilla.

Sounds good, thanks!

PR #243, opened 2022-06-14
60 meters should show as USB instead of LSB.

tmiw:
By request in #240, this change makes 60 meters show up as red in the status bar unless using USB.


Tyrbiter:
Built this change and have tested it, it is working as expected.

Thanks!

tmiw:
> Built this change and have tested it, it is working as expected.
> 
> Thanks!

No worries!

PR #256, opened 2022-07-09
Save rig names instead of rig IDs.

### LINKED ISSUE #253: Identity of connected radio missing
Could freedv-gui please add the identity or model of the connected radio in the main window please.

Hamlib seems to change the list and this is not picked up by freedv so I end up wondering why the radio gets multiple selections, changing VFOs and similar because of the mismatch of the PTT setup with the real radio.

As an example if Kenwood TS-890S is selected then I would hope to see just TS-890S in the corner of the window.

tmiw:
This PR saves rig names instead of rig IDs. By doing this, we can be more immune to changes in the Hamlib rig list. Resolves #253.


tmiw:
I'm going to merge this to master so it can get a bit more runtime as part of 1.8.2.

Issue #255, opened 2022-07-08
Release 1.8.0 depends on unreleased FREEDV_MODE_2020B from codec2


bwarden:
The release 1.8.0 source code will not compile against the latest release of codec2 (1.0.3). That release does not yet include the implementation of FREEDV_MODE_2020B (https://github.com/drowe67/codec2/pull/274).

https://github.com/drowe67/freedv-gui/blob/a612cdc79a155e0bb4614f616f6b2fb54687b63e/src/freedv_interface.cpp#L45-L46
https://github.com/drowe67/freedv-gui/blob/a612cdc79a155e0bb4614f616f6b2fb54687b63e/src/main.cpp#L922

tmiw:
As a workaround, use codec2 master. I'll see if we can get a new codec2 release out.

tmiw:
See https://github.com/drowe67/codec2/releases/tag/v1.0.4 for version 1.0.4.

tmiw:
Sorry, but it seems that v1.0.4 was premature. Continue to use codec2 master as a workaround. We're figuring out a way forward and will update with the solution.

tmiw:
Try https://github.com/drowe67/freedv-gui/releases/tag/v1.8.1. Feel free to reopen if you continue to have build issues.

PR #257, opened 2022-07-10
Disable 2020B unless Codec2 provides it.

tmiw:
Due to build issues with v1.8.0 as reported in #255 and in-progress Codec2 API changes in master, 2020B must be disabled in the source code by default. This PR will allow 2020B to be enabled as an option if Codec2/LPCNet master are used for the build.


build_linux.sh
```diff
@@ -21,23 +21,23 @@ export CODEC2DIR=$FREEDVGUIDIR/codec2
 export LPCNETDIR=$FREEDVGUIDIR/LPCNet
 
 # change this when working on combined codec2/freedv-gui changes
-CODEC2_BRANCH=master
-LPCNET_BRANCH=master
+CODEC2_BRANCH=v1.0.3
+LPCNET_BRANCH=v0.2
```
drowe67, line 25: So if we are checking out codec2 v1.0.3 ... how do we get 2020B?

tmiw: The user would need to edit this script to use master instead (or link against a Codec2 built manually from it).

tmiw:
Hmm, slight problem. It appears `lpcnet_get_hash()` isn't in v0.2. That particular function was added in [this commit](https://github.com/drowe67/LPCNet/commit/653d86b713163bf111a1a5bcf736bf43cee3e9de). Should a v0.3 be released that's tagged at that commit? Is there another commit that should be used for v0.3 instead?

drowe67:
Hmm, that's curious - so we have been using that function for two years, and I guess several freedv-gui releases.  How have we been building and packaging freedv-gui?  Perhaps we have just been using LPCNet/master?

tmiw:
> Hmm, that's curious - so we have been using that function for two years, and I guess several freedv-gui releases. How have we been building and packaging freedv-gui? Perhaps we have just been using LPCNet/master?

Yeah, we've just been using master. I'd say to just continue using master except that there are likely API changes vs. 0.2, right?

drowe67:
Yes good point about the [LPCNet API changing](https://github.com/drowe67/LPCNet/pull/42/files#diff-17917e96232aa697f318461704737980fa1974317a71781fc3f0b351f6ca31e6L17).  That in turn implies the need for codec2/master, or, as you suggest, and earlier version of LPCNet, if we stick with codec2 1.0.3

Yikes, this is all getting messy :thinking: 

How about this:
1. I'll look at codec2/master and see if I can make the minimum changes to make `freedv_api.h` compatible with existing applications - it was really just that one variable/structure.
1. I wont worry about the 2020C work for now, will do that later.
1. This implies we'll release 2020B into the wild now, which I'm not 100% happy with (as I'd like to compare it with 2020C first). Meh.
1. We can then have a another try at a codec2 1.0.4 release, and do a 0.3 LPCNet release off LPCNet/master.

Does that plan work for you @tmiw ?

tmiw:
> Does that plan work for you @tmiw ?

That works. Once (4) is done, I'll update the build scripts again, confirm that all platforms build, then merge and release 1.8.1. 👍

PR #261, opened 2022-07-18
Build Git version of Hamlib ourselves for Win32

### LINKED ISSUE #260: Upgrade Hamlib for Windows to 4.4
Per [COPR](https://copr.fedorainfracloud.org/coprs/hobbes1069/mingw/packages/), the Docker container uses Hamlib version 3.3, which is significantly behind the macOS build (and what [Fedora 36](https://packages.fedoraproject.org/pkgs/hamlib/hamlib/) provides). As there have been changes in radio support since the packages have been generated, it would be good to update the COPR packages as well or possibly just build Hamlib from git like what the macOS build does.

(ping @hobbes1069)

tmiw:
Builds latest Hamlib whenever Docker container is used. This is so that the Windows builds use the same version of Hamlib as macOS. Resolves #260.


tmiw:
@hobbes1069, I think it would still be good to have v4.4 mingw{32,64}-hamlib packages as this makes the build take significantly longer. In any case, I'll post up a test build using this PR to make sure there are no major issues with the Hamlib that got built.

CC @drowe67 in case he has feedback too.

tmiw:
Merging as I haven't heard any objections. 👍

PR #262, opened 2022-07-26
Implement automatic stats reset after sync loss

tmiw:
Feature request to automatically reset stats after sync loss. See https://github.com/drowe67/freedv-gui/discussions/244 for more info.

TODO:

* Prototype automatic sync reset and ensure that it works properly with hardcoded timeouts.
* Add configuration UI to enable/disable feature.


Tyrbiter:
I also noticed this https://github.com/drowe67/codec2/issues/182 which makes me wonder if this effort would better wait until a conclusion is reached within codec2.

tmiw:
> I also noticed this [drowe67/codec2#182](https://github.com/drowe67/codec2/issues/182) which makes me wonder if this effort would better wait until a conclusion is reached within codec2.

Based on [previous discussion](https://github.com/drowe67/freedv-gui/discussions/242) it seems like we could proceed now if we wanted, unless there's something I'm missing?

Tyrbiter:
> > I also noticed this [drowe67/codec2#182](https://github.com/drowe67/codec2/issues/182) which makes me wonder if this effort would better wait until a conclusion is reached within codec2.
> 
> Based on [previous discussion](https://github.com/drowe67/freedv-gui/discussions/242) it seems like we could proceed now if we wanted, unless there's something I'm missing?

Ah, OK, yes I looked at the previous PR but didn't realise that there were a couple of changes pulled in as a result of it. I withdraw my comment given this has already been done.

drowe67:
> > > I also noticed this [drowe67/codec2#182](https://github.com/drowe67/codec2/issues/182) which makes me wonder if this effort would better wait until a conclusion is reached within codec2.
> > 
> > 
> > Based on [previous discussion](https://github.com/drowe67/freedv-gui/discussions/242) it seems like we could proceed now if we wanted, unless there's something I'm missing?
> 
> Ah, OK, yes I looked at the previous PR but didn't realise that there were a couple of changes pulled in as a result of it. I withdraw my comment given this has already been done. I notice that PR182 is still open @drowe67 is that intentional?

Yes.

Tyrbiter:
Right. It will be interesting to see what other changes come along.

Tyrbiter:
Hi @tmiw 

So far the 10 second timeout seems to be fairly good, but there are still a few cases where it is a little long.  If 2 stations are in a QSO and pass it back quickly then the BER can be confused by this.

How much trouble would it be to give a choice of 2, 5 and 10 seconds?

tmiw:
> Hi @tmiw
> 
> So far the 10 second timeout seems to be fairly good, but there are still a few cases where it is a little long. If 2 stations are in a QSO and pass it back quickly then the BER can be confused by this. It's also because on some channels the false sync rate is higher than on others.
> 
> How much trouble would it be to give a choice of 2, 5 and 10 seconds?

Shouldn't be too much trouble. Would this setting need to be one that can be changed while the session's running, or would it be okay if the user had to stop first (like with most other settings in the application)?

Tyrbiter:
I think it would be fine to make it only change when the modem is stopped, and possibly have a default of 5 seconds.

tmiw:
> I think it would be fine to make it only change when the modem is stopped, and possibly have a default of 5 seconds.

Added. It still defaults to 10 seconds but you can go to Tools->Options->Modem to bump it up or down.

Tyrbiter:
Have now built and installed this, seems to work but I will need to try it with on air signals to assess it.

One small thing is that with the time set to 10 seconds the modem tab box for the number is a bit too narrow, so if this is the right line of code:

sizerModem->Add(sbSizer_modemstats,0, wxALL|wxEXPAND, 3);

then I assume making it slightly wider needs a tweak to one of the parameters.

tmiw:
> Have now built and installed this, seems to work but I will need to try it with on air signals to assess it.
> 
> One small thing is that with the time set to 10 seconds the modem tab box for the number is a bit too narrow, so if this is the right line of code:
> 
> sizerModem->Add(sbSizer_modemstats,0, wxALL|wxEXPAND, 3);
> 
> then I assume making it slightly wider needs a tweak to one of the parameters.

Yep, I just widened it a bit.

Tyrbiter:
> Yep, I just widened it a bit.

Thanks, also notice something similar on the debug tab with the box for the Fifo Size.

tmiw:
> > Yep, I just widened it a bit.
> 
> Thanks, also notice something similar on the debug tab with the box for the Fifo Size.

Just made that wider too. I'll merge the PR in a few days if there are no other issues.

Tyrbiter:
That has fixed it, just looked through the tabs, the same change is needed for the UDP port number.

Tyrbiter:
I'm currently listening to the UK FreeDV net on 60m and I can confirm that using a 5 second stats reset time is much better than 10 seconds, very few stations manage to resume carrier in such a short period so the stats are behaving exactly as they should.

tmiw:
> That has fixed it, just looked through the tabs, the same change is needed for the UDP port number.

Done. 👍

PR #267, opened 2022-08-14
avoid unreproducible date time embedding

maitbot:
avoid unreproducible date time embedding
by not using __DATE__ and __TIME__ compiler macros.

This will help to support reproducible builds.

73 de aa4hs,
-Maitland


tmiw:
Google seems to say that the formats of `_DATE_` and `_TIME_` are explicitly mentioned in the ANSI C standards. Which compiler(s) aren't doing that, out of curiosity?

tmiw:
If we're going to be removing build time from sox, should it be removed from CMakeLists.txt as well? For example, `DATE_RESULT`, which seems to only be used for generating package names.

tmiw:
Sounds good, will merge. Thanks!

PR #268, opened 2022-08-14
update cmake

maitbot:
Some minor refactoring of Codec2 build to better use CMake
features to improve readability ans maintainability.

- Follow the convention of cmake_minimum_required() being the first CMake command.
(Moved CMAKE_OSX_DEPLOYMENT_TARGET definition down next to another OSX setting.)

- Update cmake_minimum_required to 3.13.
Following codec2.
(POLICY CMP0075 was introduced in CMake 3.12 - since we want to follow NEW policy
we can skip the cmake_policy handling.)

- Use (new since CMake 3.0) project() features.
This is the best single-point definition of project version information.
(This handles the FreeDV_VERSION* variables for us, so the other definitions
can be removed.) This patch includes changes necessary to make version.h

- Better DATE_RESULT setting.
Using CMake's string(TIMESTAMP... function does this in 1 line rather than 5.
Also, CMake honors the SOURCE_DATE_EPOCH environment variable needed to support
reproducible builds.
https://reproducible-builds.org/specs/source-date-epoch/

- Enable CMAKE_EXPORT_COMPILE_COMMANDS
Does not change the resulting build of freedv, but allows enabling IDE features for
developers editing source files.

73 de aa4hs,
-Maitland


CMakeLists.txt
```diff
@@ -24,41 +27,19 @@ endif("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
...
 message(STATUS "Compilation date = XX${DATE_RESULT}XX")
 
 #
-# Set FreeDV version and generate src/version.h
+# generate src/version.h
 #
-set(FREEDV_VERSION_MAJOR 1)
-set(FREEDV_VERSION_MINOR 8)
-set(FREEDV_VERSION_PATCH 2)
-set(FREEDV_VERSION_SUFFIX "")
```
tmiw, line 53: How would suffix be handled after the changes? This tends to be used for test builds to make it easy to differentiate from official ones.

maitbot: CMake VERSION* handles 4 parts, the last being TWEAK.
My commit has FREEDV_VERSION_SUFFIX set to FreeDV_VERSION_TWEAK in version.h

tmiw: Hmm, reading more, `*_VERSION_TWEAK` only accepts integer values, so changing the top of the file to "1.8.2-devel" or "1.8.2.devel" won't exactly work. I pushed a commit that brings that back as a separate CMake variable.

PR #273, opened 2022-08-18
Enable 2020C in the UI.

tmiw:
Adds UI to access 2020C mode (from newest codec2).


tmiw:
Not sure I like how many items are in the list now:

Screen Shot 2022-08-18 at 7 21 13 AM

If we want to keep support for all of these, we'll need to find some other way of letting people choose.

tmiw:
@drowe67 and others, once we're sure this works I'll generate a test build for wider use?

Tyrbiter:
It's not so bad with a slightly different window arrangement
freedv_2020C

tmiw:
It's also possible that one or both of the new 2020 modes could disappear depending on user input, so perhaps we don't need to immediately consider how to clean up the mode selection?

tmiw:
1.8.4-devel prerelease build with 2020C: https://github.com/drowe67/freedv-gui/releases/tag/v1.8.4-20220819

tmiw:
Per Project Leadership Team discussion yesterday on April 14, 2023, we're not going to include 2020C at this time. This PR will be closed for now.

PR #274, opened 2022-08-21
Update build to reflect LPCNet decoupling from Codec2

tmiw:
Since we no longer need to build non-LPCNet Codec2 before building LPCNet, we can go ahead and simplify the FreeDV package building process a bit. We also need to fix some Windows compile errors that were introduced as a result.

Please see https://github.com/drowe67/codec2/pull/350 and https://github.com/drowe67/LPCNet/pull/43 for additional info.


tmiw:
@drowe67, this is pending the linked PR from Codec2 plus one more I'm going to submit for LPCNet shortly before it can be merged.

PR #287, opened 2022-10-30
Resolve issues with modulated TX audio stuttering

tmiw:
Originally noticed this using SmartSDR on macOS. Basically, if the sample rate on both devices in the Transmit tab is the same (e.g. 48KHz), the transmitted signal is perfectly fine. However, if one of them is a different sample rate (say, 16K for the analog device, which isn't uncommon for Bluetooth headsets), the TX waterfall looks a lot dirtier. The resulting signal also is nearly impossible to decode using FreeDV on another radio.

Further debugging and investigation resulted in several changes:

1. Cleanup of the TX step's implementation to align it more closely with how RX is implemented.
2. Cleanup of how number of required samples are calculated (to increase maintainability).
3. Fixed a typo during TX pipeline creation that resulted in unnecessary sample conversions _after_ the TX signal was generated.


tmiw:
I'd like to put out v1.8.4 before Activity Day so I'm planning on merging and generating new builds within a day or two. Let me know if this somehow breaks something and we can investigate further.

PR #312, opened 2022-12-10
Add 2020B to multi-RX feature

tmiw:
Per testing, 2020B does in fact work when added to multi-RX (tested with sample .wav files) and does not seem to interfere with decoding of 2020 or other modes. This PR enables users to switch to 2020B during transmit and receive without restarting the session.


Tyrbiter:
I now have an installed "kitchen sink" version with this added, it seems to work on Fedora 37 but I won't get much on air testing until other people have builds with these modes available.

tmiw:
> I now have an installed "kitchen sink" version with this added, it seems to work on Fedora 37 but I won't get much on air testing until other people have builds with these modes available.

The `ms-2020c` branch has this change but for 2020C too if interested. 👍

PR #313, opened 2022-12-10
Hide modes not used on the SM1000.

### LINKED ISSUE #311: Please label experimental modes on the GUI
Good day.

There are now 8 modes available for operators, according to the user manual at least two of them (2020 and 2020B) are experimental as of today, and does not mention whether the remaining 6 modes are stable or experimental and subjected to change.

Please label the experimental status on the GUI captions - thank you.

tmiw:
Updates the main window UI to hide modes not used on SM1000/ezDV (e.g. 2020) behind an "Others" button. Modes to always show are TBD.  Resolves #311.


tmiw:
UI on startup:

Screenshot 2022-12-10 at 9 02 40 AM

UI after expanding the additional modes:

Screenshot 2022-12-10 at 9 02 44 AM

Tyrbiter:
I get a core dump when I click Start, short output is this:

[bdm@deangelis ~]$ freedv
double free or corruption (out)
Aborted (core dumped)

Can run gdb if you need it.

tmiw:
> I get a core dump when I click Start, short output is this:
> 
> [bdm@deangelis ~]$ freedv double free or corruption (out) Aborted (core dumped)
> 
> Can run gdb if you need it. Note that I have 2020C available, not sure if this could be the problem.

No need, I was able to duplicate on macOS too :) Try again?

Tyrbiter:
OK, can now start the modem without the crash.

2 things, main.cpp:1735 looks like it should have the same assert before it as main.cpp:1727

The selected mode is now in both the SM-1000 and Other lists, I assume we should have only 1 mode selected for Tx?

tmiw:
> OK, can now start the modem without the crash.
> 
> 2 things, main.cpp:1735 looks like it should have the same assert before it as main.cpp:1727
> 
> The selected mode is now in both the SM-1000 and Other lists, I assume we should have only 1 mode selected for Tx?
> 
> That last might be due to not having implemented remembering the state of the Others mode list, currently it reverts to being closed on restart. The logic here has just got more complex, so I wasn't expecting it to be perfect straight away.

Hmm, seems to work fine on macOS. Will need to investigate further; it might be something Linux-specific causing it.

tmiw:
I added a few hidden radio buttons as apparently [GTK requires at least one selected at all times](https://github.com/wxWidgets/wxWidgets/issues/12344). Hopefully that works better for you?

Tyrbiter:
Yes, I can now only see a single radio button selected.

It does mean that I can close the Others list with, say, 2020 mode selected and see nothing selected. Is that the sort of behaviour we want? I am not sure how this should be handled, it may not actually matter much but it could confuse users.

Tyrbiter:
Gets my vote, will this be persistent across restarts of freedv or just the modem?

tmiw:
> Gets my vote, will this be persistent across restarts of freedv or just the modem?
> 
> Also, should Others: become Expert: ?

The previous mode selection is already persistent between startups so the logic to expand the other modes doesn't need to do anything specific to store its own state.

As for the labeling, let's see what others think.

Issue #317, opened 2022-12-16
Option -> Modem "Clipping" and "Band Pass Filter" applies to all OFDM modes?


cybork42:
The Options -> Modem dialog suggests that "Clipping" and "700D TX Band Pass Filter" are "FreeDV 700 Options".
During our last FreeDV technical meeting we all seemed to perceive that clipping is active at least for all OFDM modes, isn't it?

From studying the code it seems that regardless what the default settings in CODEC2 are, they are overridden by run-time settings as configured in FreeDV, like here:
freedvInterface.setRunTimeOptions( (int)wxGetApp().m_FreeDV700txClip, (int)wxGetApp().m_FreeDV700txBPF);
(code from main.cpp, lines 1430ff)

Our suggestion would be to adjust the Options -> Modem dialog, provided our observation and the code study are not wrong?

tmiw:
Clipping and BPF [appear to default to true](https://github.com/drowe67/freedv-gui/blob/master/src/main.cpp#L519) in the configuration file, which should currently match the defaults on the Codec2 side IIRC. 

Speaking of Codec2, clipping is valid for all OFDM modes as you state but BPF is only available for 700D and 700E:

```
void freedv_set_clip(struct freedv *f, int val) {
    f->clip_en = val;
    if (is_ofdm_mode(f)) {
      f->ofdm->clip_en = val;
      /* really should have BPF if we clip */
      if (val)
          ofdm_set_tx_bpf(f->ofdm, true);
    }
}

/* Band Pass Filter to cleanup OFDM tx waveform, only supported by some modes */

void freedv_set_tx_bpf(struct freedv *f, int val) {
    if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700E, f->mode) 
        || FDV_MODE_ACTIVE( FREEDV_MODE_DATAC0, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_DATAC3, f->mode)) {
        ofdm_set_tx_bpf(f->ofdm, val);
    }
}
```

Thus my suggestions:

1. "700D TX Band Pass Filter" becomes "700D/E TX Band Pass Filter" in the GUI.
2. "FreeDV 700 Options" becomes "Modem Options".

Does that seem reasonable?

cybork42:
Sounds very much like a great plan to me!
(Apparently I didn't understand the BPF Code well enough...) 

[...]

PR #332, opened 2022-12-27
Remove "force sync" option from Tools->Options

tmiw:
Per comments in #317, it's desired to remove the force sync option from Tools->Options as it's no longer needed.


tmiw:
@drowe67, what do you think about the ReSync button on the main window? Still a use for it despite removing the 700D force sync option?

drowe67:
> @drowe67, what do you think about the ReSync button on the main window? Still a use for it despite removing the 700D force sync option?

Good question :thinking: It probably isn't needed, but it's sometimes fun to press and see what happens.  Might be a good question for end users.  I guess there is an argument for simplifying the GUI as much as possible - as people do get over loaded by options.

Tyrbiter:
I did think that my discussion about 2020 modes and strange lack of decoding on successive transmissions might make this resync button useful, I don't recall using it on any other occasion.

tmiw:
> I did think that my discussion about 2020 modes and strange lack of decoding on successive transmissions might make this resync button useful, I don't recall using it on any other occasion.
> 
> I'd say leave it for now.

On that note, I did notice that `all_2020B_mpp.wav` was being detected as 2020C the other day and thus decoded incorrectly. This button could allow it to force resync to another working mode, so yeah, we'll keep it for now.

PR #339, opened 2023-01-15
Add units for SNR gauge.

### LINKED ISSUE #337: SNR vertical gauge has no units shown
Since the squelch setting is labeled in dB, the SNR gauge should also have this label.

tmiw:
Fixes #337 by adding units for the SNR gauge to match squelch.


Tyrbiter:
Yes, that's good, working as intended.

PR #347, opened 2023-02-21
Always have the Mode box take up the maximum size.

### LINKED ISSUE #346: Mode box in GUI changes columns each time Others is expanded/contracted
A minor niggle, but this also causes the Start and PTT buttons to jump about which would be nice to avoid.

After starting the program, the Mode box is in the left column to the right of the graph area.

Expand the Others modes and the Mode box jumps to the top of the right column, because it would otherwise extend below the Main window lower edge.

Now click the arrow button to contract the Modes list, still at the top right, but with only the 3 modes shown. On a 3rd click the Modes box moves back to the left column but is now expanded with the last 2 or 3 entries of the Others modes missing, the list is truncated by the main window edge. Also the Control box is too narrow now, Report Frequency is truncated at both left and right edges.

Naturally this depends on how the windows are sized when starting.
freedv_Mode

tmiw:
Resolves #346 by preventing the Mode box from resizing.


Tyrbiter:
The Others modes don't show up with the expanded selection, I'm guessing that the last edit at topFrame.cpp:673 needs to be non-empty :)

tmiw:
> The Others modes don't show up with the expanded selection, I'm guessing that the last edit at topFrame.cpp:673 needs to be non-empty :)

Yep. Weirdly, it worked fine on my Mac without that, just not on Linux.

Tyrbiter:
There was something like that in a different PR fairly recently, not sure why but the MacOS wxWidgets code is just a bit different. 

I currently have 2020C built into my local freedv-gui, with that entry there the value needed in topFrame.cpp:374 is 280 to prevent the box changing size when the list is expanded. At some point when the list size stabilises I suppose this should be optimised.

tmiw:
> There was something like that in a different PR fairly recently, not sure why but the MacOS wxWidgets code is just a bit different.
> 
> I currently have 2020C built into my local freedv-gui, with that entry there the value needed in topFrame.cpp:374 is 280 to prevent the box changing size when the list is expanded. At some point when the list size stabilises I suppose this should be optimised.

Yeah, I'll have to update this in `ms-2020c` for sure. 

Anyway, thanks! Will merge since there don't seem to be any other issues.

PR #356, opened 2023-03-26
Use official release versions of hamlib instead of latest Git

tmiw:
Due to bugs in 1.8.8 and 1.8.9-devel resulting from the use of Git hamlib, this PR downgrades hamlib to the latest official release for Windows and macOS binaries.


tmiw:
Oops, need to have the changes apply to master, not `ms-2020c`. Recreating PR.

PR #362, opened 2023-04-10
Save received callsigns and display in combo box.

tmiw:
Adds requested feature to provide a history of previously received callsigns.


tmiw:
> Do you think having a timestamp for each entry would be worthwhile?

I think so but I'm not sure about how it's currently displayed with the commit I just pushed. Suggestions are appreciated. 👍

Tyrbiter:
I managed to see 2 callsigns in the list, with timestamps.

I would suggest maybe having the timestamp 2nd and perhaps right justified? It's a bit cluttered without a space between the 2.

Tyrbiter:
> > I would suggest maybe having the timestamp 2nd and perhaps right justified? It's a bit cluttered without a space between the 2.
> 
> How about the change I just pushed? It should split those into two separate columns, or it does on macOS anyway.

Yes, and it does on Linux too, or at least on mine :)

The only thing is that after selecting an empty list and closing it the field is highlighted, but this could be just because I haven't got a callsign in the new version yet. It's only a small thing though.

tmiw:
> The only thing is that after selecting an empty list and closing it the field is highlighted, but this could be just because I haven't got a callsign in the new version yet. It's only a small thing though.

I think I'm noticing that behavior too but I'm not fully sure if it can be prevented. In any case, I pushed out a new test build with the contents of this PR thus far (https://github.com/drowe67/freedv-gui/releases/tag/v1.8.9-20230414) so we'll see how it goes with wider testing.

Tyrbiter:
A further thought, it would be useful to allow choosing between local and UTC for the timestamps. Something in the options?

tmiw:
> A further thought, it would be useful to allow choosing between local and UTC for the timestamps. Something in the options?

Maybe. How common is UTC display in other ham radio applications, out of curiosity?

Tyrbiter:
> > A further thought, it would be useful to allow choosing between local and UTC for the timestamps. Something in the options?
> 
> Maybe. How common is UTC display in other ham radio applications, out of curiosity?

It's quite common, because some countries expect log keeping to be done in UTC (when it was required in the UK we had to use GMT in the log) hence my comment. Having the universal reference may be useful for comparison or even for arriving on time for meetings :-p

Something else I have noticed, if a station is heard multiple times without another being heard in between, the callsign list has the first instance recorded but the time recorded never changes.

Tyrbiter:
Here's a nice list of stations heard with the latest multiple callsigns code:

Screenshot from 2023-04-15 23-11-09

tmiw:
> > Maybe. How common is UTC display in other ham radio applications, out of curiosity?
> 
> It's quite common, because some countries expect log keeping to be done in UTC (when it was required in the UK we had to use GMT in the log) hence my comment. Having the universal reference may be useful for comparison or even for arriving on time for meetings :-p

Sounds good. I just added an option in Tools->Options->Reporting to enable this.

> Something else I have noticed, if a station is heard multiple times without another being heard in between, the callsign list has the first instance recorded but the time recorded never changes.

That's intentional as it's supposed to show the time that the callsign was first decoded. Should it show the time of the last decode instead?

Tyrbiter:
I have built the UTC option, will report if it doesn't do what it's meant to.

I hadn't thought of the first/last decode question. How difficult would it be to indicate the number of discrete decodes for a repeated entry?

Tyrbiter:
The UTC option appears to function correctly. Thanks for adding it.

tmiw:
> The UTC option appears to function correctly. Thanks for adding it.

No worries! I think this is okay to merge now, but feel free to open an issue if anything else comes up related to this.

PR #369, opened 2023-04-19
Add Record button on the main page.

tmiw:
Per request mentioned in #368, this PR adds a record button on the main window to make it faster to kick off a recording from the radio.


Tyrbiter:
OK, built and tested, appears to work in that I have some .wav files in my home directory.

I will report further once I have managed to record some actual freedv signals :)

tmiw:
> One suggestion, when this is revisited would it be sensible to move the record button nearer to the Control buttons? Or maybe add a keyboard accelerator?

Maybe. I'll probably release a test build soon with this feature and see if there are any suggestions on placement.

Tyrbiter:
I have built and am testing these latest changes.

Had an idea: the callsign panel is blank once a signal has disappeared, so add a small indicator that the callsign list contains entries.

tmiw:
> Had an idea: the callsign panel is blank once a signal has disappeared, so add a small indicator that the callsign list contains entries.

Any ideas on what that should look like?

tmiw:
> I'm not familiar with what wxWidgets offers, but perhaps just a numeric count to the right of the drop down arrow, blanked when it's 0.

I think disabling the combo box may provide sufficient indication. See https://github.com/drowe67/freedv-gui/pull/378.

Tyrbiter:
That's a far more elegant suggestion, award yourself a gold star

Tyrbiter:
Record filename string is truncated in the main window, not sure why as window is wide enough for it:

Screenshot from 2023-04-25 18-40-41

tmiw:
> Record filename string is truncated in the main window, not sure why as window is wide enough for it:

Thanks! This has been fixed in https://github.com/drowe67/freedv-gui/pull/379.

Tyrbiter:
One thing I have noticed, when the recording is ended the audio sent to the playback device is stopped until the OK button is clicked in the dialog.

tmiw:
> One thing I have noticed, when the recording is ended the audio sent to the playback device is stopped until the OK button is clicked in the dialog.

The waterfall definitely stops until you push OK but I don't recall the audio stopping as well (though I could be remembering wrong). The latter isn't ideal for sure.

Anyway, now that the status bar is working properly, perhaps the dialog boxes aren't actually necessary? People can still get feedback by either looking there or checking the Tools menu.

PR #378, opened 2023-04-25
Disable callsign drop-down unless there are entries in the list.

tmiw:
This PR disables the callsign drop-down list when the list has no entries. This allows users to better determine when they've received signals.


Tyrbiter:
This seems to work OK, it's not completely obvious when the drop down arrow pulls up the list but the field background colour only appears when the list is active which is quite a nice cue.

I'd say this is fine now.

Tyrbiter:
What I see since this change is that the callsign display field does not clear after a certain time even if another signal appears without the pskreporter callsign/reporting enabled.

tmiw:
> What I see since this change is that the callsign display field does not clear after a certain time even if another signal appears without the pskreporter callsign/reporting enabled. A second signal appearing may clear it, I haven't managed to get a good handle on this yet.

I don't think it's related to this change specifically but rather stuff that I did in #352. On one hand, I did get feedback back then preferring that the callsigns stay displayed until another one is decoded, so that behavior wasn't changed at that time. On the other, I'm not sure that's necessary anymore given the drop-down list. I'll leave it for now unless I hear further feedback.

Tyrbiter:
OK, what I saw today is that you get the previous callsign showing during a transmission from the current station who is using the free text sending setting. Maybe there is an interaction with that. Will try to pin it down.

PR #379, opened 2023-04-26
Prevent status bar text from truncating unless required.

tmiw:
Per report [here](https://github.com/drowe67/freedv-gui/pull/369#issuecomment-1522174476), the status bar is improperly truncating the text within it. This is due to the status bar being split into three segments, of which only one is really being used.

This PR sets the status bar such that it only has one segment. Additionally, the status bar text (used for the various Tools->Record/Play options) has been standardized and cleaned up.


Tyrbiter:
The status bar is now behaving itself correctly after this change.

PR #389, opened 2023-05-09
Add last received SNR to callsign list.

tmiw:
Per request from https://github.com/drowe67/freedv-gui/discussions/388, this PR adds a new column showing the last received SNR for the contact.


Tyrbiter:
I changed this:

snrAsString.Printf(wxT("%d"), pendingSnr);

to

snrAsString.Printf(wxT("%.1f"), pendingSnr);

and it seems to work, although the only SNR I've managed to capture was 0.0 dB

I thought a little more resolution would be useful.

tmiw:
In the current code, pendingSnr is an int and not float, so that change won't have any effect. Is there a need to have more precision? I'm not sure there's much (if any) difference between, say, 9 and 9.5 dB.

Tyrbiter:
I should have spotted that.

Personally I'd prefer to have 0.1dB resolution unless there is a reason not to do it, the SNR in the main window displays 0.1dB resolution so the value is available I think.

tmiw:
> I should have spotted that.
> 
> Personally I'd prefer to have 0.1dB resolution unless there is a reason not to do it, the SNR in the main window displays 0.1dB resolution so the value is available I think. Just noticed that this is in the pskreporter code where you do an (int)( g_snr + 0.5 ) so maybe pskreporter can only cope with integer values.

Yeah, PSK Reporter only does integers but it shouldn't be a huge deal to have the callsign list show with higher precision: https://github.com/drowe67/freedv-gui/pull/391

PR #390, opened 2023-05-10
Add support for potential future FreeDV mapping tool

tmiw:
This PR makes changes to the freedv-gui codebase to support a potential future FreeDV mapping tool (similar to PSK Reporter but with FreeDV-specific enhancements). In order to support such a tool, the following has to be done:

1. ~Remove dead end UDP code and associated system options.~
2. ~Move PSKReporter class into a new `fdv_reporting` component.~
3. ~Create IReporter base interface for PSKReporter and future map support class.~
4. ~Update reporting logic to iterate over list of reporters (instead of just using PSKReporter).~
5. ~Create map support class and associated protocol.~
    a. Using socket.io with hardcoded URL for testing.
6. ~Create mapping tool and link to application.~
    a. https://freedv-reporter.k6aq.net/ (source code https://bitbucket.org/tmiw/freedv-reporter/)
7. ~Update Tools->Options to make reporting options more agnostic.~


Tyrbiter:
Should this be able to use https:// as pretty much everything is going that way?

tmiw:
> Should this be able to use https:// as pretty much everything is going that way?

Possibly. The socket.io C++ bindings can do it, even, just OpenSSL needs to be linked in as well. And I'm not really sure how to do that for Windows builds just yet. 

Even if the FreeDV app itself doesn't use https, that doesn't preclude the web UI from doing so (and in fact, is probably mandatory at this point for e.g. Google Maps).

Tyrbiter:
OK, wasn't sure about how this all works, but I thought I'd ask in case there was something that needed to change significantly. Looks interesting even if there isn't yet a site where all this can be linked together.

Tyrbiter:
https://github.com/drowe67/freedv-gui/pull/390/commits/7cc1ed724e099d1be03d7fd106baca54821dfad9

This change is causing my freedv to frequently report nothing of any use, with SNRs down to -5 and no callsign. We never got a fix for the sync on a tone around 700Hz, so many analogue SSB and other noise-like signals seem to cause sync to happen.

tmiw:
Yeah, doesn't look like we can just report when we have sync for >X seconds when sync can happen anytime other than an actual FreeDV signal. I reverted that commit.

tmiw:
I accidentally committed these changes when I committed the hamlib PR. They do seem to be running okay, though, so it's probably safe to merge regardless. If we run into any other problems, we can open a new issue/PR.

PR #393, opened 2023-05-17
Defer sound device checking until Start is pushed.

tmiw:
Some users plug in their sound devices after starting FreeDV but before pushing the Start button. The existing implementation in `master` causes the sound device(s) to be inappropriately wiped. This PR removes the startup check and checks only on pushing the Start button.


w4jkl:
This is a great idea. I've been annoyed by the setup clearing several times.
73,
John W4JKL

PR #394, opened 2023-05-17
Add ability for Hamlib to use RTS/DTR instead of CAT for PTT.

tmiw:
Per request from Discord due to some interfaces (such as from MicroHam) only supporting CAT control for frequency/mode retrieval.


mturpin-pmc:
This allows proper PTT operation with my Microham USB Interface III.  
Works as expected in operation after pressing START.   
The application can query the radio via Hamlib for the frequency.  
The PTT button works as expected.

The "Test PTT" button in the "PTT Config" menu seems to ignore the "PTT uses:" setting however..

tmiw:
> This allows proper PTT operation with my Microham USB Interface III. Works as expected in operation after pressing START. The application can query the radio via Hamlib for the frequency. The PTT button works as expected.
> 
> The "Test PTT" button in the "PTT Config" menu seems to ignore the "PTT uses:" setting however..

Good to hear! I think I forgot to copy some setup stuff to the PTT Config, hence why it didn't work there. I'll copy that over in a bit.

PR #395, opened 2023-05-20
Link in dynamic libraries for hamlib

tmiw:
WIP so that people can change out their Hamlib library to diagnose CAT control/PTT issues (especially on Windows and macOS).


tmiw:
Just confirmed that this runs fine on both Windows and macOS. I'll merge now.

@Tyrbiter BTW, the libcodec2 and liblpcnetfreedv.so paths in `ldd` pointing to /home and not /usr are likely because both have had their versions incremented recently. I don't think this is actually a problem.

PR #396, opened 2023-05-20
Reduce spacing between lines on Easy Setup dialog

tmiw:
Adding the PTT source dropdown caused the Easy Setup dialog to be cut off on 720p displays again. This PR cuts down the spacing between rows to attempt to help with that.


Tyrbiter:
This seems to work for me, the window is compressed vertically and the buttons and drop downs are still visible and operate correctly. I'm not using a 720p display, but there is plenty of vertical space on 1080p.

PR #400, opened 2023-05-26
Use get_vfo instead of retrieving VFO B frequency directly on start.

tmiw:
Just a minor change for the Hamlib logic to hopefully avoid an issue where the radio swaps over to VFO B before going back to VFO A.

Also, using different VFOs for PTT seems to cause problems on some radios. This PR causes RIG_VFO_CURR to always be used for PTT.


tmiw:
Turns out WSJT-X does have some special Hamlib logic. The current multi-VFO check on startup is influenced by this code.

Additionally, it seems to just use `RIG_VFO_CURR` for PTT; it doesn't bother getting the current VFO beforehand in that case. Perhaps going back to that (along with this change) will be sufficient to fix any remaining issues.

Tyrbiter:
I am going to integrate this PR into my local build over this weekend, I hope I can test it without the crash problem intruding.

Tyrbiter:
A quick test on the TS-890 shows that this code works as expected on both VFO A and B, A is set to 20m and B is on 40m, Tx occurs on the correct VFO and band as selected on the radio after I swap bands manually. I also tested in Memory mode and Tx correctly occurred on the Memory frequency so I don't think this has broken anything here.

I will continue to build with this PR included.

PR #402, opened 2023-05-28
Update user manual to reflect existence of multiple reporting services.

tmiw:
Per email feedback, this PR updates the user manual to reflect the support of multiple reporting services (PSK Reporter and FreeDV Reporter).


Tyrbiter:
I think this is a good set of changes, I don't think it needs anything else added now, perhaps in the light of user questions, I hope they will come up on the chat page on the reporter web site.

Issue #403, opened 2023-05-29
FreeDV Reporter - should not show receiving or transmitting freedv when analog is selected


Tyrbiter:
I use freedv-gui when operating in analogue modes, it's a useful interface with my USB microphone, I don't use the front panel mic connectors on my radios.

It seems to me that the FreeDV Reporter logic should recognise this and take my callsign out of the list if I am not in digital mode.

tmiw:
See PR #404. 👍

Tyrbiter:
Looking good to me @tmiw, I hope it works as well for everyone else too.

PR #404, opened 2023-05-29
Disconnect from FreeDV Reporter if in analog mode.

tmiw:
Per request in #403, this disconnects the application from FreeDV Reporter when in analog mode.


Tyrbiter:
I have built this and started testing, so far it seems to work exactly as intended. I will check to make sure that there is no obvious interaction with the crash problem from before.

PR #405, opened 2023-05-29
Use drop-down list for frequency on main window

tmiw:
Per feedback from the digitalvoice list, this PR does the following:

1. Changes the frequency box in the lower right to a drop-down list containing all of the frequencies currently listed on [FreeDV Reporter](https://freedv-reporter.k6aq.net/) and the [QSO Finder](https://qso.freedv.org/).
2. If hamlib is enabled and the frequency is changed, the frequency is changed on the radio as well (in addition to the correct mode as per the accepted standard: LSB/DIGL for <= 10 MHz and USB/DIGU for >10MHz or 60 meters).


tmiw:
I tested with an IC-705 and a Flex 6300 and both seem to work okay with this PR. 

@Tyrbiter, does this work fine on the TS-890S as well?

tmiw:
> In the main window the entry field in the drop-down is not wide enough at 10MHz and above.

Fixed.

> The TS-890 is now being set into USB-D mode, I normally use USB and of course there is the analogue button to think about!

The Analog button should now cause the radio to go from USB-D/LSB-D to just USB or LSB (and back again once Analog is no longer pressed).

Re: normally using just USB mode, I thought that regular USB and LSB messed with the FreeDV waveform and reduced SNR at best. What are your reasons for using it?

> It would be nice to put the radio back into the mode it was in before running freedv, but I'm not sure if I think this should happen when stopping the modem or when exiting the program, maybe that could be a choice on the radio tab?

The code now makes an attempt to revert the frequency and mode when pushing Stop. I don't think waiting until the main window's actually closed is something that would be expected by most people (not to mention that it would require additional code changes to do so).

Tyrbiter:
Oh yes, the reversion thing with Stop vs exiting the whole program. I have no problem with that, it probably makes more sense than my hand-waving suggestion anyway.

Tyrbiter:
Just noticed, the version at https://github.com/drowe67/freedv-gui/pull/405/commits/e90f379e85571a80397e3b295d1f489d0b654a34 doesn't change the mode to DATA on Start, only when selecting a frequency from the drop-down list. I think that's a bug.

w4jkl:
Please leave an empty slot in the list so that we are not locked into the list (who decided those frequencies anyway?) and we can enter an arbitrary frequency.
Thank you.
73,
John W4JKL

tmiw:
> Just noticed, the version at [e90f379](https://github.com/drowe67/freedv-gui/commit/e90f379e85571a80397e3b295d1f489d0b654a34) doesn't change the mode to DATA on Start, only when selecting a frequency from the drop-down list. I think that's a bug.

Yep, it's a bug. I should have a fix for that (and the segfault) by tonight.

> Please leave an empty slot in the list so that we are not locked into the list (who decided those frequencies anyway?) and we can enter an arbitrary frequency. Thank you. 73, John W4JKL

The text in the drop-down list is able to be edited, so you're not restricted to the frequencies in the list. 

As for your other question, I know that 14.236 is an official IARU frequency for digital voice. I'm not sure about the others.

tmiw:
So yeah, check out Tools->Options->Rig Control for the setting that will force it to use USB/LSB instead of USB-D/LSB-D.

Tyrbiter:
That's a nice improvement, I don't see the crash on exit any more and I have the Rig Control tab in the Options with a checkbox that changes the selected mode when I click Start. The radio also returns to the previous frequency and mode on clicking Stop.

The only thing I notice is that if I select Analog, the radio mode remains as DATA, I would expect it to switch to USB or LSB on the current frequency of the active VFO. I suppose this could have another checkbox on the Rig Control tab, but that seems like overkill to me as it's obvious what the Analog button does.

Tyrbiter:
Good, that seems to do exactly what it should now, I have tried a few sequences of button pushes and closing the program and so far it's looking good.

Tyrbiter:
> > One of the 5MHz frequencies should be 5366.5 kHz, it's shown as .3
> 
> Oops, will fix. Looks like it's mistyped on FreeDV Reporter too.

Could the reporter web page get one more significant figure for the frequency?

Tyrbiter:
The reported frequencies, so that people don't appear to be off the notional frequencies.

Tyrbiter:
> > The reported frequencies, so that people don't appear to be off the notional frequencies. 5.3665 shows as 5.367 for instance.
> 
> Updated. You may need to clear your browser cache prior to reloading, however.

Shift-reload did it. Thanks.

tmiw:
Cool. Assuming you don't find any further problems today, I think we can merge this and officially release 1.8.10. 👍

Tyrbiter:
> Cool. Assuming you don't find any further problems today, I think I can merge this tonight and officially release 1.8.10. +1

That will be a relief for everyone I think :)

PR #408, opened 2023-06-02
Fix issue preventing FreeDV Reporter from remaining in sync with radio.

### LINKED ISSUE #407: 1.8.10 release breaks FDV reporting.
I built freedv-1.8.10-devel yesterday (31/05/2023) from git master at 0c096ed and all was fine.

Today I saw that 1.8.10 final was released and was all set to push this new release to Mageia (just in time for Mageia 9) after using and testing git master myself for a few weeks but my plan has been thwarted.

I see that rig frequency control has been added which is a massive untested change to introduce in the 24 hours before a final release!

I have just spent several hours trying to figure what is going on.

======================
After reverting to 0c096ed:
 I can switch from memory mode to VFO frequency at will and FreeDv shows the reported frequency correctly. FDV Reporter likewise follows this after a few seconds.

However if I run klog which also tracks the rig frequency all is fine unless I switch to VFO from M mode, then the rig starts to rapidly jump from VFOA to Memory channel and back endlessly. This could be a klog issue as it is supposed to be read only so ignoring this for now.

=========================================
I have now reinstalled 1.8.10 release to test again:
klog is not running.
With M mode selected and the frequency at 14.240, FDV is showing 14.240 and FDV reporter is showing the same. Switching M channel to 14.236 shows correct freq in FDV but now after a minute or so FDVR is still showing 14.240 which replicates what I was seeing earlier today.
Changing rig M freq to 7.177 likewise shows correctly in FDV but FDVR is still at 14.240 with last update showing as 9 mins ago.

Changing to VFO mode on rig now:
VFO was at 18.150 (see /OT at end while I remember). FDV shows 18.150. FDVR is still showing 14.240.

Selecting and clicking a new freq in the new selector in FDV changes the rig correctly to that frequency and hey presto FDVR changes too.

Going back to the rig M mode changes FDV to the correct freq, but FDVR is again stuck on the old VFO freq.

So for now this is looking like a reporting bug introduced by the new frequency setting feature, ignoring the fight between freedv, klog and hamlib for the time being.

/OT
While looking at this I noticed a typo in the freq listings in FDVR for 17m. It says 18.188 which I guess should have been 18.118. 

System is Mageia 9 (dev) (Linux) x86_64
Hamlib 4.5.5 using rigctld with rig type 2003 Kenwood TS450S
lib64sioclient is from git at b10474e

Barry

tmiw:
Fixes #407 by bringing back combo box event handler that was improperly removed. Additional changes were made to ensure manual frequency entry still works.


USER_MANUAL.md
```diff
@@ -854,6 +863,11 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 # Release Notes
 
+## V1.8.XX June 2023
```
tmiw, line 866: I'm undecided if we need to release a 1.8.10.1 immediately or if this can wait till 1.8.11. Thoughts?

Tyrbiter: > I'm undecided if we need to release a 1.8.10.1 immediately or if this can wait till 1.8.11. Thoughts?

I think the .1 release is needed asap, this has bitten one person but could easily affect others.

I am now testing the new code, but won't get much time in person until Friday evening UTC.

Tyrbiter:
I have done some testing and as far as I can see everything is working as it should be, from my point of view it's good to go.

tmiw:
FYI I was doing a bit of testing with my FT-817 as well earlier and apparently hamlib will not send the needed settings change commands to change "DIG" mode to "USER-L" (from "USER-U") when FreeDV requests `RIG_MODE_PKTLSB`, thus causing Hamlib errors. I don't think that's in scope for this PR, though, nor something that needs to be worked around for 1.8.10.1 (especially since users can probably just use the Tools->Options setting for regular USB/LSB instead).

Tyrbiter:
Hamlib is a work in progress, seemingly without end :/

I am using 4.6~git master as that is what we're being asked to test, 4.5.6 has been dropped as it's too difficult to back port to it.

If it doesn't work with current 4.6~git then the Devs are asking for bug reports, so unless the FT817 is unfixable it's probably worth doing that.

tmiw:
I can edit the build scripts locally (after releasing 1.8.10.1) to use Hamlib git instead and see what happens. Meanwhile, as a workaround, FreeDV should at least change the frequency regardless of whether a mode change worked (relying on the indicator turning red if it's the incorrect sideband).

mdblack98:
I added PKTLSB to the FT817 -- and a warning message to tell the user about Menu 26.
I think that's the best I can do since we can't modify USER-L USER-U for the FT817 that I know of (not documented).

mdblack98:
See the update on the Hamlib page -- we might be able to fix this based on what you users want.
We can write to the EEPROM to set PKTUSB and PKTLSB but we haven't been writing to the FT817 EEPROM to avoid burnout.  Mode change is infrequent enough we could do that without increasing the number of writes dramatically.
As a side note I thought this whole idea of upper/lower sideband was long gone as it was due to the limits of very old rigs.  I don't any digital mode that does LSB at all.  FT8 sure doesn't and I don't now why FreeDV can't just stick with PKTUSB.
https://github.com/Hamlib/Hamlib/issues/1308

Tyrbiter:
> As a side note I thought this whole idea of upper/lower sideband was long gone as it was due to the limits of very old rigs. I don't know of any digital mode that does LSB. FT8 sure doesn't and I don't now why FreeDV can't just stick with PKTUSB. [Hamlib/Hamlib#1308](https://github.com/Hamlib/Hamlib/issues/1308)

It's something that came about because as FreeDV is a voice mode it seemed sensible to stay with the LSB below 10MHz convention. I don't know if this convention will ever be changed. It would take a lot of effort to do it.

mdblack98:
FreeDV is digital voice.
So make it a choice and let people learn that LSB below 10Mhz is old school and no longer necessary and to use USB for all bands -- it simplifies life and reduces confusion.  Make USB the default.
Switching USER-L USER-U on the FT817 could probably be done even with LSB operation if I can get somebody to confirm that Hamlib is reading all the modes correctly.
The mode switch should be pretty infrequent enough we wouldn't run into the 1M to 10M write limit of the EEPROM.
One question on the FT-817 -- if you change frequencies and power-cycle the rig does the old frequency come back up?  That would mean it is saving that to EEPROM and mode changes would be a LOT less frequent than frequency changes.

tmiw:
Thanks for the update @mdblack98, I'll take a look at the related Hamlib issue after responding here.

> I don't know of any digital mode that does LSB.

IIRC HF packet uses LSB. I'm sure there are at least one or two others, but I don't know any others offhand.

> It's something that came about because as FreeDV is a voice mode it seemed sensible to stay with the LSB below 10MHz convention. I don't know if this convention will ever be changed. It would take a lot of effort to do it.

I brought up the question in the digitalvoice list, so we'll see how that goes. I think the fact that there's rig control (other than PTT) in the application now changes things a bit, though the users that use serial PTT instead of Hamlib may disagree.

> One question on the FT-817 -- if you change frequencies and power-cycle the rig does the old frequency come back up? That would mean it is saving that to EEPROM and mode changes would be a LOT less frequent than frequency changes.

I just tested on mine by doing the following:

```
$ rigctl -m 1020 -r /dev/...

Opened rig model 1020, 'FT-817'

Rig command: f
Frequency: 14235990

Rig command: F 7177000

Rig command:
```

(turn off FT-817, unplug power for ~5 seconds and reattach/turn on)

```
Rig command: f
Frequency: 7176970

Rig command:
```

(I bumped the VFO dial while powering down, hence the discrepancy.)

So yeah, it does look like frequency changes are getting stored in EEPROM.

PR #422, opened 2023-06-07
Set initial audio device count on startup

tmiw:
There's a bug where Tools->Audio Settings shows "none" for all four devices unless you push Start first. This PR resolves that particular issue by loading the number of sound cards on startup. (This is a side effect introduced by #393 as the code prior to that PR implicitly did this on startup during the audio device checks.)


tmiw:
I'm thinking we might need a 1.8.10.2. :(

tmiw:
Or actually, perhaps we can go to .11 if all open PRs get merged?

Tyrbiter:
[spinal tap]
It's always good to go to 11.
[/spinal tap]

PR #427, opened 2023-06-08
Add forgotten frequency/mode suppression on close.

tmiw:
PR #424 was incomplete and didn't suppress frequency and mode changes when shutting down. This PR adds the additional required code to make this happen.


tmiw:
> I have had this running for 24 hours with multiple Start/Stop cycles and closing and opening freedv without any apparent problem.

Cool, will merge this. The other issue @barjac saw seems like a Hamlib thing but that can be worked around in another PR if needed.

PR #442, opened 2023-06-18
Hamlib: Set mode before frequency.

### LINKED ISSUE #441: Mode should be changed before frequency
Per the digitalvoice list, apparently the FT-991A will offset the frequency a bit when using FreeDV as the mode gets changed after the frequency. Swapping the order of operations will supposedly fix this.

tmiw:
Resolves #441 by changing the order of operations that the hamlib code performs. That is, the radio's mode should be updated before the frequency so that if we're going between LSB and USB, some radios don't end up offset from where they should be.


Tyrbiter:
Seems to work for me, no obvious problems.

PR #453, opened 2023-06-27
Set background color in FreeDV Reporter to red on TX.

tmiw:
I was finally able to work around the issue I was having with the background color in FreeDV Reporter not reverting when going from TX to RX. This PR adds that feature to bring the built-in client more in parity with the Web version.


Tyrbiter:
Seems to work for me OK. Checked for own Tx and remote stations.

PR #461, opened 2023-07-08
Auto-size columns in Audio Options based on device names.

tmiw:
Based on feedback, this PR auto-sizes the columns inside the Audio Options window to make longer device names easier to read.


tmiw:
@barjac, hopefully this works okay for you. 👍

barjac:
Yes, thanks. :)
I agree with Brian that the other columns would benefit from the same treatment.

tmiw:
> Yes, thanks. :) I agree with Brian that the other columns would benefit from the same treatment.

Wait, the code should already be resizing all four columns unless I'm misunderstanding something?

barjac:
Yes exactly, sorry for confusion - I was going from memory.

Tyrbiter:
> > Yes, thanks. :) I agree with Brian that the other columns would benefit from the same treatment.
> 
> Wait, the code should already be resizing all four columns unless I'm misunderstanding something?
> 
> EDIT: oh, Brian meant preserving the size/location of the Audio Options dialog. +1

Yes, I meant that the last 3 columns are effectively hidden so to see them all the window has to be widened, but then when it's closed the same thing happens when it is reopened.

PR #462, opened 2023-07-08
Save and restore FreeDV Reporter window state

tmiw:
Per feedback, this PR adds the ability to restore FreeDV Reporter window state on next start.


tmiw:
@barjac, I did some testing here (albeit on macOS) and it looked like the main window was saving and restoring its size on startup, so I just added the ability to do the same for the FreeDV Reporter window. Let me know if you run into problems.

barjac:
Using a package from  https://github.com/drowe67/freedv-gui/tree/ms-preserve-reporter-layout head.

The main window still moves so it is centred vertically in the screen.
The Reporter is almost in the right place.
Before restarting:
Screenshot_20230708_124958
After stop, close and re-start of freedv
Screenshot_20230708_125104

The main issue now with Reporter is that the dialogue will not expand horizontally to fit the screen width. The Reporter window will but the height (which previously would expand to take more calls) and the width which with some fiddling would expand horizontally now seems solidly fixed.

Maybe I should not be testing this branch but applying the PR to master? I did try but the PR patch would not apply.
How should I be testing your PRs?

barjac:
This is what the last release can be set to, with screen layout how it would be good to see on start-up.
Screenshot_20230627_202442

barjac:
This patch will not apply to master, however testing a build from  this branch does not allow the table to be expanded to fit the full screen width and seems to have a fixed size with scroll bars.

tmiw:
> The main window however is still returning to be vertically centred in the main screen.

I did some more testing and I think it might be related to the wxWidgets version that's in use. I'm not able to duplicate this on OpenSUSE or macOS, for example, and this appears to be due to using the "wxWidgets-3_2" package on OpenSUSE (and always bootstrapping the latest on macOS). Compare to Ubuntu 22.04 (where I *am* able to duplicate the issue) where it apparently uses wxWidgets 3.0.5 if "apt search" is any indication:

```
libwxgtk3.0-gtk3-0v5/jammy,now 3.0.5.1+dfsg-4 arm64 [installed,automatic]
  wxWidgets Cross-platform C++ GUI toolkit (GTK 3 runtime)

libwxgtk3.0-gtk3-dev/jammy,now 3.0.5.1+dfsg-4 arm64 [installed]
  wxWidgets Cross-platform C++ GUI toolkit (GTK 3 development)
```

I'll update my build_linux.sh locally and see if I can get it to build its own wxWidgets instead (similar to what's done for Windows and macOS).

barjac:
We are also using wxgtku3.0-3.0.5.1
My D.E. is plasma5 which uses kwin-5.27.5
I will test this branch in different D.E.s tomorrow.

barjac:
I realized that we had wxgtk3.2-3.2.1 available in the distro and re-built ms-preserve-reporter-layout branch against it.
Success!
Only one issue remaining:
When the program is re-started and the modem started, the reporter table is too narrow, but by grabbing the reporter window re-sizing corner (bottom right) and moving it a fraction, the table immediately expands to fit the window.
Just restarted:
Screenshot_20230710_114419
After slight movement of window corner:
Screenshot_20230710_114439

barjac:
I just tested this in an IceWM session and the behaviour is identical.

Tyrbiter:
> Is the OK button in FDVR redundant?

Maybe it should be labeled 'Close'

barjac:
> > Is the OK button in FDVR redundant?
> 
> Maybe it should be labelled 'Close'

I don't see any need for it. The window has a 'X' and when FreeDV closes, Reporter closes with it. To move it off the screen it can be minimised.
Why is it needed?

Tyrbiter:
> > > Is the OK button in FDVR redundant?
> > 
> > 
> > Maybe it should be labelled 'Close'
> 
> I don't see any need for it. The window has a 'X' and when FreeDV closes, Reporter closes with it. To move it off the screen it can be minimised. Why is it needed?

Some people use WMs that don't show all or any of the conventional window controls? Odd, but they exist.

barjac:
I will take your word for that ;) 
If it must stay then I would go for "Close" or "Exit" as "OK" would normally be in answer to a question or to accept a change, neither of which apply here.

tmiw:
> When the program is re-started and the modem started, the reporter table is too narrow, but by grabbing the reporter window re-sizing corner (bottom right) and moving it a fraction, the table immediately expands to fit the window.

This should be fixed now. Additionally, "OK" was changed to "Close" and I fixed a display issue with the SNR column (the website blanks it out on frequency change but the app didn't until now). Let me know how this goes and hopefully we can merge if all goes well.

barjac:
Yes this is working - Nice!
However...
There is a problem when stopping the program with no Last TX or Last Update fields having dates in them. 
On next start the table width is inadequate to take all the fields when dates do appear later.
I think the 'unknown's in these columns should be padded out to the width of the longest possible date by default. This should then remove the need to ever expand them further and the table width will always open at that width.
'Unknown' is longer than callsigns or modes but shorter than dates.

I did see the main window restore to mid screen height on a few occasions but this cleared itself on subsequent tests. Maybe caused by the exact relative positions of the windows on saving or the phase of the moon :)

Another issue I just saw after leaving it running while using Firefox to type this, is that on returning to FreeDV I found the reporter blank except for my call and the bottom buttons moved to the left of the window just below it. Touching the window resize corner handle slightly restored it back to normal.

barjac:
Hmm.. Not really. That has padded fields that were OK not just the dates.
The Last transmit column, if all are unknown on program start, still is truncated after a transmission is made.

This needs some fine tweaking as now the 'Last update' is also truncated as the whole line is too long, probably due to globally padding 'Unknown' rather than tweaking it on a per column basis?

Screenshot_20230711_181711

barjac:
I have discovered what causes the main window to restore to being vertically centred on the screen.
It happens when the program was previously closed with the main window touching the screen top edge. Moving it down a few pixels fixes it so it restarts in the correct last position.

Tyrbiter:
Sounds like a WM or DE bug.

barjac:
> Sounds like a WM or DE bug.
Yes maybe kwin, it does not happen in IceWM

barjac:
I have been playing with the reporter and now have this, based on a patched 40b3e94 .
By removing 'unknown' and using some shorter column labels it is more compact and recovers correctly from close/re-launch.
What do you think?
Screenshot_20230711_230213

barjac:
This is the diff applied to 40b3e94 (board will not accept .diff or .patch extensions)
[unknown.txt](https://github.com/drowe67/freedv-gui/files/12021885/unknown.txt)

Tyrbiter:
QRA should be QTH, QTH locator (originally called Maidenhead) was the replacement as QRA squares only covered Europe.

As for forcing upper case, yes for callsigns, but the locator string can be extended to 8 or more characters some of which are lower case.

https://en.wikipedia.org/wiki/Maidenhead_Locator_System has the details. It does mention 5th and 6th pairs of characters, but absolute hard and fast rules seem to be missing :)

tmiw:
I don't think I agree with the proposed column headings, especially "QTH Locator" or "QRA". But that could be something that's more common to say outside the US, too (vs. e.g. "Grid Square").

Anyway, the latest in this PR changes how the column resizing is done. Instead of only when a user connects, it now takes the text lengths of each of the cells into account when determining whether column resize is needed. This allows each server update to potentially trigger a column resize without triggering any noticeable flicker. (Oh, and there was a off by one error that was preventing the "Last Update" column from resizing, which was probably contributing too.)

Tyrbiter:
> I don't think I agree with the proposed column headings, especially "QTH Locator" or "QRA". But that could be something that's more common to say outside the US, too (vs. e.g. "Grid Square").

Actually I agree, I just say "Locator" and have done even when it was the old QRA system. Grid Square makes perfect sense too, you may have seen that some people put other grid system references in the field so a neutral description is the thing to have.

barjac:
> > I don't think I agree with the proposed column headings, especially "QTH Locator" or "QRA". But that could be something that's more common to say outside the US, too (vs. e.g. "Grid Square").
> 
> Actually I agree, I just say "Locator" and have done even when it was the old QRA system. Grid Square makes perfect sense too, you may have seen that some people put other grid system references in the field so a neutral description is the thing to have.

Yes that makes sense I would prefer Locator too, although other systems is not good if there is any chance that mapping could be incorporated in the future.

barjac:
The latest update to this PR is working mostly OK for me. Maybe the 'Last TX' column could be padded very slightly to allow for dates added later than whatever set the width. See station transmitting in the screenshot who appeared last.

Screenshot_20230712_104525

This (below) happens if Firefox is opened for any length of time while FreeDV is running like now as I type this, on returning to Freedv:
  
Screenshot_20230712_105643

barjac:
It seems that the column width for 'Last TX' on start of the Reporter is not using the widest data width, unless it is simply counting characters and not pixels. Both the entries below under 'Last TX' were there at startup.

Screenshot_20230712_183234

Tyrbiter:
I think this is due to a proportional font, I see this too although it's not preventing me from reading the information.

tmiw:
I can't test right now to see if the change I just committed helps, but give it a shot. If we still need to, we might be able to manually add padding and pass the calculated size in.

barjac:
Looks good to me! Well done!
The only changes that I would now make would be to force upper case on the first two characters of the locator, and change 'Grid Square' to 'Locator' to save some wasted space in that column, that is only normally needed by the label.

Screenshot_20230712_222801

Tyrbiter:
Earlier I saw a JA station with a 10 character locator string :)

tmiw:
Looks like we're all good on this PR now assuming the Windows version (which I'm building now) works.

BTW I can't duplicate the Firefox issue on Ubuntu 22.04 so I wouldn't be surprised if it was some window manager issue:

Screenshot 2023-07-12 at 6 55 12 PM

PR #460, opened 2023-07-05
Add support for modifying the drop down frequency list.

tmiw:
Per feedback in #432, this PR adds the ability for users to modify the frequency list in the lower right hand corner of the main window (via Tools->Options->Rig Control). The new UI looks as follows:

Screenshot 2023-07-04 at 8 25 11 PM

Note: PR #457 must merge before this can as the code here depends on that PR.

PR #466, opened 2023-07-09
Saves position/size of Audio Config dialog between executions.

tmiw:
Per request from https://github.com/drowe67/freedv-gui/pull/460#issuecomment-1627505631, this PR saves and restores the position/size of the Audio Configuration dialog similarly to the FreeDV Reporter dialog (#462) and the main window.


Tyrbiter:
This is now good, I can see all the columns nicely.

PR #469, opened 2023-07-11
Add ability to suppress automatic frequency reporting on radio changes.

tmiw:
For some setups (i.e. ALE), it's not preferred to have FreeDV automatically report frequency changes to FreeDV/PSK Reporter. This PR adds a new setting in Tools->Options->Reporting ("Manual Frequency Reporting") to prevent frequency updates from the radio from being propagated to the FreeDV UI and further upward to the various reporting services.

### DIFF
--- a/src/config/ReportingConfiguration.cpp
+++ b/src/config/ReportingConfiguration.cpp
@@ -33,6 +33,8 @@ ReportingConfiguration::ReportingConfiguration()
         
     , reportingFrequency("/Reporting/Frequency", 0)
         
+    , manualFrequencyReporting("/Reporting/ManualFrequencyReporting", false)
+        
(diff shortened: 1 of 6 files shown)


Tyrbiter:
I'm not familiar with ALE operation so I suggest adding a simple explanation of what ALE does and why this is needed to the manual.

tmiw:
> I'm not familiar with ALE operation so I suggest adding a simple explanation of what ALE does and why this is needed to the manual.

I added a paragraph to the manual about it. Let me know anything that should be changed about the wording.

Tyrbiter:
I think that's adequate, unless you can think of other modes/uses that have similar frequency changes.

PR #485, opened 2023-07-26
Use separate max values for each slider type in the Filter dialog.

### LINKED ISSUE #463: Filter setting sliders too aggressive (1.18.12 pre)
Recently the level adjustments for the filters have become too course and difficult to set.
Reported to me by GW3OAJ on Windows and confirmed by me in Linux.

tmiw:
Resolves #463 by using separate ranges for each slider type on the Filter dialog. This has the effect of making sure a single press of the arrow key (when a slider has focus) adjusts a single "unit" on the selected control, hopefully making it easier to make filter selections.


tmiw:
@barjac, @Tyrbiter, let me know how this change works for you. I think that as long as you can return to a previous value by using the arrow keys, we should be fine. 👍

Tyrbiter:
It looks OK to me now, although the click and drag interface is odd on Fedora I suspect that is a wxWidgets bug of some kind. The arrow keys work, but there is nothing to cue the user to use them unless I've missed something.

tmiw:
> Adding a sentence to the manual would make sure that we can deny any problems plausibly ;-)

Done 👍

Issue #148, opened 2021-07-13
Waterfall update speed slows significantly with multiple panes visible


tmiw:
Per digitalvoice list:

> Thanks for all your work bur unfortunately the scatter plot is not visible on my windows10 box.

> Another interesting thing I noticed is that if the waterfall display is the only active pane then it takes approx 30 seconds for waterfall to scroll down ie the time scale is accurate. However with multiple panes ((spectrum at top and waterfall at bottom) it now takes 60 second for waterfall to go from top to bottom. There is a waterfall speed issue as the size of the window is resized. This is not  a serious issue.

Thoughts on what could be happening?

drowe67:
Hi @tmiw - hmm that's an interesting one.  The multiple pane stuff has always been kind of opaque, something internal to wxWIdgets.  For example we have never worked out how to save/restore the multiple pane state.

Perhaps some insight could be gained by dumping a few variables (like the X and Y window size) as it's resized, and see how that affects the algorithm that drives the update rate.

tmiw:
MainFrame::OnTimer() is what triggers the adding of data points (if needed) and the repainting of each of the panes (regardless of whether they're displayed). I think wxWidgets suppresses the Refresh() calls for the panes that aren't visible, though, or at least doesn't trigger the actual painting to the screen until they do become visible. OnTimer() seems to be set to be called every 100ms (governed by _REFRESH_TIMER_PERIOD, defined as DT*1000 where DT == 0.1), which means that one of the panes visible in OP is potentially taking significantly longer than that to redraw.

Alternatively, it's possible that this block is simply not executing:

```
    if (m_panelWaterfall->checkDT()) {
        m_panelWaterfall->setRxFreq(FDMDV_FCENTRE - g_RxFreqOffsetHz);
        m_panelWaterfall->m_newdata = true;
        m_panelWaterfall->setColor(wxGetApp().m_waterfallColor);
        m_panelWaterfall->Refresh();
    }
```

checkDT() is defined as follows:

```
bool PlotWaterfall::checkDT(void)
{
    // Check dY is > 1 pixel before proceeding. For small screens
    // and large WATERFALL_SECS_Y we might have less than one
    // block per pixel.  In this case increase m_dT and perform draw
    // less often

    float px_per_sec = (float)m_rGrid.GetHeight() / WATERFALL_SECS_Y;
    float dy = m_dT * px_per_sec;
    
    if (dy < 1.0) {
        m_dT += DT;
        return false;
    }
[...]

PR #487, opened 2023-07-28
Adjust waterfall settings to better visualize 2 Hz fading.

### LINKED ISSUE #79: Waterfall maintenance
Requirements:
1. Ensure bandwidth of waterfall is sufficient to visualise fading at up to 2 Hz (we can provide test signals).

tmiw:
Resolves #79 and #148 by adjusting some of the waterfall settings to prevent filtering out fast fades. This includes:

* Adjusting the built-in IIR filter for the waterfall so that newer data is a bigger contribution to the calculated points.
* Adjusting how new lines are drawn in PlotWaterfall by drawing a `MODEM_STATS_NSPEC` x 1 temporary image and having wxWidgets resize to the correct width and `dy`.
* Use a fixed number of pixels/second to avoid issues with fractional positioning and improper WF scrolling speed.

Additionally, this PR adds the ability to average the spectrum plot across 1-3 samples.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -897,6 +897,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix various screen reader accessibility issues. (PR #481)
     * Use separate maximums for each slider type on the Filter dialog. (PR #485)
     * Fix minor UI issues with the Easy Setup dialog. (PR #484)
+    * Adjust waterfall settings to better visualize 2 Hz fading. (PR #487)
 2. Build system:
     * Update Codec2 to v1.2.0. (PR #483)
 3. Cleanup:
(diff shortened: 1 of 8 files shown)


Tyrbiter:
This is a nice improvement, seems to reduce CPU load quite a bit too.

tmiw:
> This is a nice improvement, seems to reduce CPU load quite a bit too.

Neat, I wasn't even aiming for lower CPU 👍

Tyrbiter:
Would it be possible to have an averaging function on the spectrum display? Maybe allow 1, 2 or 3 sweeps, the new spectrum is a bit noisier and I think it would help to make signals a bit clearer. I am assuming that the waterfall changes can be retained without the spectrum plot affecting those.

tmiw:
> Would it be possible to have an averaging function on the spectrum display? Maybe allow 1, 2 or 3 sweeps, the new spectrum is a bit noisier and I think it would help to make signals a bit clearer. I am assuming that the waterfall changes can be retained without the spectrum plot affecting those.

I added something that I think might do what you're asking. Let me know how it looks? It seems to smooth things out a bit more here, anyway.

Tyrbiter:
> > Would it be possible to have an averaging function on the spectrum display? Maybe allow 1, 2 or 3 sweeps, the new spectrum is a bit noisier and I think it would help to make signals a bit clearer. I am assuming that the waterfall changes can be retained without the spectrum plot affecting those.
> 
> I added something that I think might do what you're asking. Let me know how it looks? It seems to smooth things out a bit more here, anyway.

This is good for me, the only thing is that the averaging setting is not saved, would be nice if it were.

PR #489, opened 2023-07-31
Deprecate PortAudio support on Linux.

tmiw:
This PR updates the `build_linux.sh` script to default to using PulseAudio/pipewire instead of PortAudio. Additionally, a warning now prints during `cmake` if the user attempts to compile FreeDV with PortAudio on a Linux machine.


tmiw:
Been wanting to do this for a while as FreeDV is a much better experience using PulseAudio/pipewire. I'm not sure when PortAudio usage on Linux will be ultimately disallowed (if at all, especially since Windows and macOS still need it).

PR #492, opened 2023-08-02
Preserve averaging setting across restarts.

tmiw:
An additional setting for spectrum plot averaging was added to #487 but didn't preserve this across restarts. This PR adds the necessary configuration option to enable this.


tmiw:
@Tyrbiter ^

Issue #438, opened 2023-06-15
Confusion regarding Voice Keyer


barjac:
From what I can read/guess, the voice keyer .wav files should be unprocessed audio from the mic input.

I can see no option in the tools menu to record such a file un-digitized.

A CQ type message needs to be recorded using the same mic and signal path and levels as when transmitting normally, so it sounds seamless, but how?

I see the option to 'Start play file - mic in' but none for 'Start record file - mic in' .

What am I missing?

Tyrbiter:
This is from a FreeDV manual at v1.1.0, which is pretty old:

"You will need to record the audio you wish to send as a WAV file using whatever program you have available on your PC to do this. Note: the audio MUST be recorded with an 8K sample rate or it will not work correctly with FreeDV."

So the long and short is that you need some other software to record the audio. It would certainly be nice if the freedv program had such a facility added to make things more straightforward.

Tyrbiter:
How difficult would be a record from selected microphone to WAV file be to implement @tmiw ?

I would guess that the nuts and bolts to do this are already present in freedv.

barjac:
I suspected that it was not included so thanks for confirming.
Thinking ahead...
It would be nice to be able to click to start recording during a transmission and then hit the voice keyer button to continue using the recording with the existing timings.
Quick, easy to use on the fly and levels and acoustics can match perfectly.
Files could be sequentially named and manually renamed later for re-use (somehow) should the need arise.

tmiw:
> How difficult would be a record from selected microphone to WAV file be to implement @tmiw ?
> 
> I would guess that the nuts and bolts to do this are already present in freedv. Maybe the controls could be in the Options->Audio tab where the voicekeyer setup is already.

Yeah, much of the code is likely already there, just need to figure out how best to show the UI for it.

Though really, perhaps it's a good opportunity to do some cleanup. I'm not sure how frequently the "record from modulator" option's used, for example.

DG9BFC:
sdr console also has a builtin voice keyer (with recording option) .. it records in 16 bit mono 48 kS and you can have up to 50 files maximum (normally not needed but nice to have) ... not only cq calls but also other audio files (that are converted to above numbers) can be played ... be it a test tone signal, a cq call or a jingle or a previously made recording of the other stations signal ... play once or play in a loop (with setable pause in between) ... so most parts are built in freedv .. only needs a record from mic and stop button and save under xyz filename ... voice keyer then should have a drop down list with the filenames
[...]

PR #493, opened 2023-08-05
Record/Play Cleanup and Voice Keyer Improvements

tmiw:
Per #438, this PR does the following:

1. Removes "Record File - From Modulator" and "Play File - Mic In" options from the Tools menu as they aren't commonly used.
2. Enables inclusion of audio from the modulator when transmitting while recording from the radio. This allows a two way QSO to be easily recorded.
3. TBD


barjac:
Item 1.
This is a shame as I have used it many times to test microphone settings.

Item 2. 
If I start to transmit and hit Record while talking then hit Record again to stop the recording, then a 44 byte .wav is created.
So item 2 appears to not currently work.

Item 3.
I think this will need a separate Voice Keyer wav file directory to be defined, to keep raw mic audio files apart from encoder output audio files, since both use .wav extensions which can be confusing.

Not yet tested this (Item 3) as I will need to install Audacity and figure out how to use it (again). :(
It really needs this capability built-in to freedv to allow quick on the fly message recording.
The operator should also be able to listen to what is being transmitted by the Voice Keyer for many reasons!

tmiw:
> Item 1. This is a shame as I have used it many times to test microphone settings.

Are there any settings you're confirming other than the level? The Level indicator on the left hand side should work in both transmit and receive.

> Item 2. If I start to transmit and hit Record while talking then hit Record again to stop the recording, then a 44 byte .wav is created. So item 2 appears to not currently work.

Yep, this should be fixed now. It actually did work before but only if you start the recording while in RX.

> Item 3. I think this will need a separate Voice Keyer wav file directory to be defined, to keep raw mic audio files apart from encoder output audio files, since both use .wav extensions which can be confusing.

The files from the Record button have a specific file naming scheme: FreeDV_FromRadio_[date].wav. Not sure if that helps or not.

> Not yet tested this (Item 3) as I will need to install Audacity and figure out how to use it (again). :( It really needs this capability built-in to freedv to allow quick on the fly message recording. The operator should also be able to listen to what is being transmitted by the Voice Keyer for many reasons!

Pull the latest and right-click the Voice Keyer button :)

Tyrbiter:
Probably should have put this here:

https://github.com/drowe67/freedv-gui/issues/438#issuecomment-1666576663

tmiw:
> Probably should have put this here:
> 
> [#438 (comment)](https://github.com/drowe67/freedv-gui/issues/438#issuecomment-1666576663)

Try now?

Tyrbiter:
https://github.com/drowe67/freedv-gui/pull/493/commits/98c43d8388448825a1172bd17af2b70368da5af6 helps, although the menu is still outside the window a little.

Both entries show the same keyboard shortcut too, and the tool-tip ought to be modified to include the new function.

tmiw:
> [98c43d8](https://github.com/drowe67/freedv-gui/commit/98c43d8388448825a1172bd17af2b70368da5af6) helps, although the menu is still outside the window a little.
> 
> Both entries show the same keyboard shortcut too, and the tool-tip ought to be modified to include the new function.

Done.

barjac:
> Are there any settings you're confirming other than the level? The Level indicator on the left hand side should work in both transmit and receive.

I was using this mainly for testing filter settings.

Since you mention the level gauges. In dark mode the pale blue is very hard to see, especially when the main window loses focus and dims due to the system volume control taking focus to set the level. Something with higher contrast would be better.

> 
> > Item 2. If I start to transmit and hit Record while talking then hit Record again to stop the recording, then a 44 byte .wav is created. So item 2 appears to not currently work.
> 
> Yep, this should be fixed now. It actually did work before but only if you start the recording while in RX.

Ah OK there were no stations around to RX when I tested this.
> 
> > Item 3. I think this will need a separate Voice Keyer wav file directory to be defined, to keep raw mic audio files apart from encoder output audio files, since both use .wav extensions which can be confusing.
> 
> The files from the Record button have a specific file naming scheme: FreeDV_FromRadio_[date].wav. Not sure if that helps or not.

Not really, I think off-air recordings should be kept separate from VK files. The recordings directory will get cluttered, TX messages need to be quick and easy to find.

> 
> > Not yet tested this (Item 3) as I will need to install Audacity and figure out how to use it (again). :( It really needs this capability built-in to freedv to allow quick on the fly message recording.
> 
> Pull the latest and right-click the Voice Keyer button :)

Will do :)

tmiw:
> Since you mention the level gauges. In dark mode the pale blue is very hard to see, especially when the main window loses focus and dims due to the system volume control taking focus to set the levels. Something with higher contrast would be better.

Might be a window manager issue. For example, the gauges appear orange for me:

Screenshot 2023-08-05 at 1 29 20 PM

barjac:
Re. Pull the latest and right-click the Voice Keyer button :)

Yes that is good, I like it.
Can the last selected message remain as the default (across re-starts) until another is selected?
.... and the Tools->Options->Audio tab 'Wave file' entry box become 'Voice keyer file directory'?

As it is now it appears to assume that one message is always default and set in the config and any others are just temporary.

I think that whichever keyer message was last selected should remain the default and any one of a small library of recorded messages may be selected as the current default at any time.
This is more fluid and the Voice Keyer button would be used without needing to touch the context menu for a full session.
Just like selecting a frequency, you would select a message and continue to use it, tomorrow as well :)
WDYT?

tmiw:
> Re. Pull the latest and right-click the Voice Keyer button :)
> 
> Yes that is good, I like it. Can the last selected message remain as the default (across re-starts) until another is selected? .... and the Tools->Options->Audio tab 'Wave file' entry box become 'Voice keyer file directory'?
> 
> As it is now it appears to assume that one message is always default and set in the config and any others are just temporary.
> 
> I think that whichever keyer message was last selected should remain the default and any one of a small library of recorded messages may be selected as the current default at any time. This is more fluid and the Voice Keyer button would be used without needing to touch the context menu for a full session. Just like selecting a frequency, you would select a message and continue to use it, tomorrow as well :) WDYT?

I could see a situation where someone forgets what they used previously and ends up transmitting the wrong WAV file out the next day. How about the following?

1. The selected VK file is persistent but just for the current FreeDV run.
2. The default VK file is still set in Tools->Options->Audio and is what starts off as the "selected" one. 
3. When doing (1), the folder the selected file is in persists across restarts (i.e. if you restart FreeDV and attempt to choose a VK file again, the window will start off in that folder instead of whatever random folder).

barjac:
> > Re. Pull the latest and right-click the Voice Keyer button :)
> > Yes that is good, I like it. Can the last selected message remain as the default (across re-starts) until another is selected? .... and the Tools->Options->Audio tab 'Wave file' entry box become 'Voice keyer file directory'?
> > As it is now it appears to assume that one message is always default and set in the config and any others are just temporary.
> > I think that whichever keyer message was last selected should remain the default and any one of a small library of recorded messages may be selected as the current default at any time. This is more fluid and the Voice Keyer button would be used without needing to touch the context menu for a full session. Just like selecting a frequency, you would select a message and continue to use it, tomorrow as well :) WDYT?
> 
> I could see a situation where someone forgets what they used previously and ends up transmitting the wrong WAV file out the next day. How about the following?

Which is why we need to hear what is being transmitted :)

I have just gone through all your suggestions and at first was against them, however I now see the advantages of doing what you suggest.
The 'default' remains with the config settings unchanged.
Any new folder(s) created for VK files may be used and selected on the fly and the current one will survive a re-start/boot.
I assume the current one will be stored in .freedv?
So after first use the 'default' is useless and ignored - perfect!  ;)

That sounds good for me :)

tmiw:
> I assume the current one will be stored in .freedv? So after first use the 'default' is useless and ignored - perfect! ;)

`playFileToMicInPath` in ~/.freedv is where that gets stored. The trigger for that being updated is when you push OK after selecting a new voice keyer file; it'll basically split the selected file into path and file name and save the former. When you go to choose another VK file (or record a new one), the open/save window will default to showing the folder saved in `playFileToMicInPath`.

Also, I updated the behavior of the VK file selection so that it doesn't go into TX after selection (it will still go into TX if you record a new file, though). Left-clicking the Voice Keyer button will transmit the newly selected file (or the one configured in Tools->Options->Audio if you didn't choose another file yet for that session).

barjac:
Bug:
After recording a new message it cannot be selected as the file selector is not refreshed and the new file is not shown.

tmiw:
How's the latest in this PR/branch working for you guys? Think we're ready to merge?

barjac:
Probably yes.
My only concerns are:
1. Being forced to transmit while recording, with no way to easily disable audio out to TX without potentially upsetting level. This also still reports a TX in the reporter which is also undesirable.
If TX while recording was optional then it certainly would have it's uses for quick on the fly custom messages, but I really think record without TX should be an option.
2. We need to listen to messages during TX - I know I keep going on about it.  /o\
Maybe another issue should be started for this?

It does seem good to go as it is, without major bugs, however I think the above should be later enhancements.

Tyrbiter:
That last change works for me, new voice file is recorded and reporter doesn't see my Tx.

Real Tx still works and shows up in the reporter.

barjac:
Nice! - Brilliant job :) 
Works for me too!

I won't mention the other thing :)

Tyrbiter:
> Nice! - Brilliant job :) Works for me too!
> 
> I won't mention the other thing :)

This other thing?

https://github.com/drowe67/freedv-gui/issues/498

barjac:
Yep :)

PR #497, opened 2023-08-09
Save and restore tab state

### LINKED ISSUE #495: Restore screen layout between re-starts of FreeDV
I have been asked to add this 'Enhancement Request' by Matt who reads the RSGB news in the UK using FreeDV every Sunday.
He has a rather complex system and uses this screen layout for FreeDV during the broadcasts:

image

As you can imagine it takes a bit of setting up each time, so it would be good if it would restore to this layout on each start.

73
Matt & Barry

tmiw:
Experimental change to enable saving and restoring of tab state. Theoretically resolves #495.


src/topFrame.cpp
```diff
@@ -32,18 +32,179 @@ extern int g_playFileFromRadioEventId;
...
 extern int g_txLevel;
 
-// Override for wxAuiNotebook to prevent tabbing to it.
-class TabFreeAuiNotebook : public wxAuiNotebook
+// THIS IS VERY MUCH A HACK! wxTabFrame is not in the public interface and should
+// not be here. Unfortunately this is needed to get the tab state loaded and
+// saved. Here's hoping this interface remains stable.
+//
+// (Last retrieved from wxWidgets 3.2 on August 8, 2023.)
+class wxTabFrame : public wxWindow
```
tmiw, line 40: This is the change I'm not comfortable with considering it's a non-public interface from wxWidgets itself. I don't *think* it changes all that often but still.

tmiw:
Hmm, apparently on wxWidgets 3.0 they actually prevent wxTabFrame from being in the dynamic symbol table for Linux builds. This might be a problem :(

tmiw:
Anyway, this code works on macOS but that's not going to be much help. Maybe this stays open in case someone eventually comes up with a better idea in the future.

barjac:
This appears to work OK in Linux with the proviso that the main window is set to full screen and the program closed and re-started BEFORE setting up the screen layout. Otherwise the previously saved main window and FRDVR window location over-rides it and layout is lost.

tmiw:
I haven't run into that issue on my Linux install, unfortunately, but rearranging the ordering of the tabs does seem a bit flaky and I'm not sure why. As mentioned before, this is a pretty big hack and will probably need to be worked on for a while before it can be merged (if at all).

tmiw:
I'm going to wrap this in an "experimental features" checkbox, as follows:

Screenshot 2023-09-10 at 9 50 12 PM

This is so that we can get some runtime with other people and hopefully get additional feedback on whether this should be an official feature.

PR #500, opened 2023-08-11
Adds VK/TX monitoring options

### LINKED ISSUE #498: "Monitor mode" for TX
Basically, this would be an option that could be enabled to allow one to listen to themselves as they're transmitting. This would allow live feedback for filter changes (among other things).

Proposed workflow:

* User right-clicks on the PTT button and chooses "Monitor" to enable or disable this option.
* User left-clicks to transmit/stop transmitting as normal. If "Monitor" is checked", they'll hear themselves :)

tmiw:
Resolves #498 by implementing two different monitoring modes:

1. "Monitor recorded audio" (when right clicking on the Voice Keyer button): this allows monitoring of the audio being recorded by the voice keyer.
2. "Monitor transmitted audio" (when right clicking on the PTT button): this allows monitoring of the audio being transmitted over the air prior to being modulated.

In both instances, the audio is taken just after input equalization is applied.


tmiw:
Not updating changelog just yet as I'm not sure if this is going into 1.9.0 or a 1.9.1 test build.

Tyrbiter:
Have built and installed this, will do some testing later if I have time or over the weekend.

barjac:
Just installed in my PA-pipewire system and it's not playing any sound during PTT or VK TX. 
The GUI selectors are all fine.
Just no audio monitored. :(

Tyrbiter:
> Just installed in my PA-pipewire system and it's not playing any sound during PTT or VK TX. The GUI selectors are all fine.
> 
> EDIT: But it works in analogue mode - not in digital :)
> 
> EDIT2: Also tested in PA only system with same result.

I can confirm this as well, I didn't try the analogue mode but I expect that will be the same as @barjac reports.

tmiw:
@Tyrbiter, @barjac, try the latest :)

barjac:
Hmm not quite what I was expecting!

There seems to some confusion about what is actually required.

1. I don't want to hear myself while recording a VK message, it's impossible to do with the delay in my headset . 
3. I want to hear what is being sent to the encoder during transmission of a VK message.
4.  I don't want to hear myself (with a delay) in the headset while transmitting normally, again it's impossible to do.

tmiw:
> Hmm not quite what I was expecting!
> 
> There seems to some confusion about what is actually required.
> 
> 1. I don't want to hear myself while recording a VK message, it's impossible to do with the delay in my headset .
> 2. I want to hear what is being sent to the encoder during transmission of a VK message.
> 3. I don't want to hear myself (with a delay) in the headset while transmitting normally, again it's impossible to do.

Wait, wouldn't just the one setting for TX monitoring be sufficient, then? You could specifically enable the checkbox in the PTT right-click menu right before sending the voice keyer message and disable it afterward. Unless I'm missing something?

barjac:
Sorry please see my edited message above.

> > Hmm not quite what I was expecting!
> > There seems to some confusion about what is actually required.
> > 
> > 1. I don't want to hear myself while recording a VK message, it's impossible to do with the delay in my headset .
> > 2. I want to hear what is being sent to the encoder during transmission of a VK message.
> > 3. I don't want to hear myself (with a delay) in the headset while transmitting normally, again it's impossible to do.
> 
> Wait, wouldn't just the one setting for TX monitoring be sufficient, then? You could specifically enable the checkbox in the PTT right-click menu right before sending the voice keyer message and disable it afterward. Unless I'm missing something?

No I don't want to be messing with context menus during operating. 
I want to set the voice keyer monitoring ON once and hear what the VK is transmitting. Once set it should stay that way.
That is all that is needed.

tmiw:
OK, *this* time it should match the desired behavior.

Tyrbiter:
I think this is a good solution, I am also happy that the option to monitor audio on the PTT button is there too in case I want to check something during setting up when the short delay (which can be longer depending on your hardware) exists but does not too seriously affect the speech/hearing loop in the brain.

barjac:
> OK, _this_ time it should match the desired behavior.

Yes that looks ideal in testing :)
However if the 'Record' button is pressed and VK started, the recording stops when the VK is pressed again to interrupt the VK message.
Would it be possible to have the general 'Record' button in total control of the the recording and not have it stop when VK and PTT are used? The PTT does not currently interrupt the recording, but the VK does.

barjac:
Ah It's now not recording though. Just tried to record a new VK file and it's not doing anything after the filename is entered. Just sits in RX.

Tyrbiter:
> Ah.... VK record is not working now, it was in previous build. I just tried to record a new CQ_40m.wav VK file and it's not doing anything after the filename is entered. Just sits in RX. A 44 Byte file is created.

Yes, just had that happen here. I thought it was OK yesterday but didn't check properly.

barjac:
> > Ah.... VK record is not working now, it was in previous build. I just tried to record a new CQ_40m.wav VK file and it's not doing anything after the filename is entered. Just sits in RX. A 44 Byte file is created.
> 
> Yes, just had that happen here. I thought it was OK yesterday but didn't check properly.

I have yesterday's build (b7596e4) in my PA/PW system and it does record a VK file OK.

Tyrbiter:
Just need to let @tmiw get the logic sorted out, I think the function is OK but control interaction with the various buttons can confuse what gets saved.

tmiw:
I think I got the problem with the VK recording fixed as of the latest commit. It also doesn't seem to interrupt the normal recording if you're doing a VK recording, and I made sure that one's recorded audio wasn't seeping into the other somehow. Let me know how this goes.

Tyrbiter:
OK, not good, although the file is now recorded there is a massive buzzing feedback into my headphones while the voicekeyer is recording. I get .wav files of appropriate size, I expect they sound like the speaker/headphone output.

tmiw:
I'm not fully sure why, but memset() didn't work to zero out the sound before (and I didn't notice the issue on my ARM Mac, either, just my Intel one). It should be good now.

Tyrbiter:
Seems to be fine now, thanks for the fix. I will test more over the next day or two, I imagine @barjac will too.

I don't know why some of these functions don't work either, I would have thought memset would be common enough that it's not down to a particular compiler.

barjac:
Using build of #1317040.
I am not seeing any difference regarding the Main recording being interrupted by the VK TX start/stop.

I see that you mentioned making a main recording of making a VK recording! Why would anyone do that? ;)

Steps to reproduce:
Start modem - now hearing nothing (or RX noise in analogue) .
Use context menu in VK to select a CQ call
Hit Main Record button on left of screen.
Hit VK button - listen to VK message now being transmitted
Hit VK button to stop transmission.
The Main recording stops, so any remote reply is not recorded nor are further PTT transmissions

/OT
I do see another issue in that during RX breaks between individual repeats of a VK message, there is no indication that the VK is still running. There is no knowing which toggle state it is in. Maybe the button could change to a colour when it is active. As it is now there is no way to know whether the VK is about to transmit again or not.

Tyrbiter:
> Using build of #1317040. The main issue I am seeing here now is that there is no way to know whether the Main recording is still ongoing, since the VK playback message wipes the 'recording' message. My assumption that the recording had stopped was maybe down to the message being wiped. We really need the Main Record button to be red while recording is ongoing or it gets very confusing.
> 
> There is another similar issue in that during RX breaks between individual repeats of a VK message, there is no indication that the VK is still running. (No sound in headset and no TX light) There is no knowing which toggle state it is in. Maybe the button could change to a colour, maybe amber, when it is active? As it is now there is no way to know whether the VK is about to transmit again or not.

These are good suggestions @barjac I think we're still feeling our way with the UI on these features and indicating the status is a real need.

barjac:
> > Using build of #1317040. The main issue I am seeing here now is that there is no way to know whether the Main recording is still ongoing, since the VK playback message wipes the 'recording' message. My assumption that the recording had stopped was maybe down to the message being wiped. We really need the Main Record button to be red while recording is ongoing or it gets very confusing.
> > There is another similar issue in that during RX breaks between individual repeats of a VK message, there is no indication that the VK is still running. (No sound in headset and no TX light) There is no knowing which toggle state it is in. Maybe the button could change to a colour, maybe amber, when it is active? As it is now there is no way to know whether the VK is about to transmit again or not.
> 
> These are good suggestions @barjac I think we're still feeling our way with the UI on these features and indicating the status is a real need.

Thanks, I agree and also think that if it can be done by changing a button colour rather than wasting screen space with a 'status bar' then so much the better. Nobody looks at status bars anyway.

tmiw:
I thought the various buttons were supposed to be appearing pressed when active? Though I can see how that might not be sufficient, especially since there was a previous request from someone else to make the PTT button red for example. I'll play some more tonight.

BTW:

> I don't know why some of these functions don't work either, I would have thought memset would be common enough that it's not down to a particular compiler.

I just realized this morning that the size passed into memset() should have been `*numOutputSamples * sizeof(short)`, not just `*numOutputSamples`. I'll try that as well tonight if only to make the mute step a tiny bit faster. :)

barjac:
> Well the PTT button is red when active now, so having the voicekeyer button similarly red when it's active would fit in. As @barjac mentioned you can't easily see when voicekeyer has returned to Rx hence the suggestion for a different colour during the receive part of the voicekeyer cycle.

No that is wrong - you can see when it is in RX as it's not transmitting, it's whether it is 'active' that needs to be shown by a colour! Nothing to do with TX or RX but whether the VK is still cycling or has timed/counted out.

Tyrbiter:
Maybe that would work, I think that clarity of Tx being active is important, some radios don't have a massive red LED on them or are not in immediate line of sight from the operator. Things for @tmiw to consider.

barjac:
> Maybe that would work, I think that clarity of Tx being active is important, some radios don't have a massive red LED on them or are not in immediate line of sight from the operator. Things for @tmiw to consider.

The PTT currently goes red during VK TX so it is clear when it is transmitting. However there is no distinction between a PTT TX and a VK TX. In my suggestion for orange for the VK button when active, then maybe the PTT should also be orange during a VK TX to differentiate. This would make it clear that the PTT could still be pressed to switch to (red) manual PTT operation which would cancel the VK session (all orange goes off).
I really like the work flow that would offer :)
Orange is just an example BTW, no specific reason other than it being a warning colour.

tmiw:
How about blue? This is based on https://www.theverge.com/23650428/colorblindness-design-ui-accessibility-wordle, where yellow might be problematic.

Tyrbiter:
I could go with blue, it needs to be bright enough to get the user's attention.

Was also thinking about that record button, maybe the text on it should turn red when active. I'd say it doesn't want to be as attention-getting as the PTT button.

tmiw:
Unfortunately it doesn't look like I can adjust the button foreground colors in a platform independent way so I just went with the following for background colors:

* VK button: blue-ish when transmitting the VK file, red when recording
* Left-hand Record button: red background when recording

Hopefully that helps a bit.

Tyrbiter:
Thanks, that seems to work OK, although I think 55, 155, 175 is a better blue colour.

Small bug, if you click the voicekeyer button and the file specified in the Options->Audio tab doesn't exist then the button stays blue after dismissing the error message, I would expect it to return to the inactive state/colour.

I thought wxWindows was supposed to *be* platform independent :)

tmiw:
> Thanks, that seems to work OK, although I think 55, 155, 175 is a better blue colour.
> 
> Small bug, if you click the voicekeyer button and the file specified in the Options->Audio tab doesn't exist then the button stays blue after dismissing the error message, I would expect it to return to the inactive state/colour.

Fixed :)

> I thought wxWindows was supposed to _be_ platform independent :)

On that note, I think I found an issue with macOS and light mode of all things where the background colors on buttons don't show up at all. I'm not even sure if there's anything FreeDV can do to work around that :/

barjac:
Testing #0fa6cea
Bug:
After the VK transmits it's message plus the correct number of repeats, it drops back to inactive, but the button remains pale blue. Clicking the button again in an attempt to turn it off fails, it remains blue and another set of messages start.
Clicking the VK button while VK is active works correctly to stop the VK session.

I tested the PTT button while the VK was active in the hope that you may have implemented my suggestion to allow the PTT button (while VK is active) to stop the VK session and switch to normal TX. Very handy to cut a VK message short and seamlessly add a live tail to it.

barjac:
Testing #0fa6cea
Could the display switch automatically from 'waterfall' to 'Frm Mic' during VK message recording (as it does between  TX/RX)?

Tyrbiter:
> Testing #0fa6cea Bug: After the VK transmits it's message plus the correct number of repeats, it drops back to inactive, but the button remains pale blue. Clicking the button again in an attempt to turn it off fails, it remains blue and another set of messages start. Clicking the VK button while VK is active works correctly to stop the VK session.

Probably missing a statement to set the button null colour, should be easy to find.

> 
> I tested the PTT button while the VK was active in the hope that you may have implemented my suggestion to allow the PTT button (while VK is active) to stop the VK session and switch to normal TX. It would be very handy to cut a VK message short and seamlessly add a live tail to it, or indeed to reply to a caller with one click, without having to click VK and then PTT!

The voicekeyer is supposed to turn off if some seconds of valid sync are received. I have not been able to test this yet, I am only recently using the voicekeyer facility.

tmiw:
> Probably missing a statement to set the button null colour, should be easy to find.

Yep 👍 

> Could the display switch automatically from 'waterfall' to 'Frm Mic' during VK message recording (as it does between TX/RX)?

And done.

barjac:
> > I tested the PTT button while the VK was active in the hope that you may have implemented my suggestion to allow the PTT button (while VK is active) to stop the VK session and switch to normal TX. It would be very handy to cut a VK message short and seamlessly add a live tail to it, or indeed to reply to a caller with one click, without having to click VK and then PTT!
> 
> The voicekeyer is supposed to turn off if some seconds of valid sync are received. I have not been able to test this yet, I am only recently using the voicekeyer facility.

Yes I thought that was the case and it probably worked last night, however with the blue staying on I was not aware. IIRC when I hit VK to stop it, it started again so that is probably working.
I did have a nice contact with W1IZZ from my CQ using VK for the initial call and it mostly went well including recording of the QSO.
The major issue was that FDV crashed 3 times during the QSO - always during RX while decoding, sadly I did not really have time to restart with gdb and the system in question does not have all the debug packages installed anyway.
When using the VK I noticed some glitching/missing sylables/very short gaps in the audio. Whether this was only in the local monitor or also the TX I am not yet sure. The glitches appear in the same place on each transmission, so it's more likely the recording that has the problem. I will get the chance to test again with the latest later today with luck. Re-recording produced similar faults.  I have not noticed this in earlier versions.

In light of the crashes which on restart set mode, power attenuation, FDVR band etc.etc. to what was set at last shutdown, it makes me wonder whether these should be saved more regularly, either timed or on each TX.
Reason: I had been testing with minimum power and on another band on 700E when last closed. This session I was using normal power and on a different band  and 700D. The program crashed and on re-start I had no output power, had switched to 700E and was now watching the wrong band in the reporter. Not very convenient to say the least as I had taken half an 'over' before I noticed the power situation  and it was two overs later before I noticed the mode change. I guess that if the frequency had been controlled by freedv that would have been reset too :(

tmiw:
I just checked in a few changes to address the monitor audio dropouts and bring in the other bugfixes done thus far. Let me know if that helps 👍

barjac:
So far all seems good. I left it recording the net and a few hours afterwards when I finally remembered it was still running with no crashes. The dropouts are also gone :)

Tyrbiter:
Seems OK to me too, I have a build here with all of the recent changes included except the layout restoration branch and it seems very good. With an activity weekend on the way I expect it will get stressed some more so I will report anything odd that I find.

Top job @tmiw :)

tmiw:
> Seems OK to me too, I have a build here with all of the recent changes included except the layout restoration branch and it seems very good. With an activity weekend on the way I expect it will get stressed some more so I will report anything odd that I find.
> 
> Top job @tmiw :)

Thanks!

Anyway, I'm thinking I'm going to release 1.9.1 tonight with this PR simply because the current changelog has enough crash-type bug fixes where it'd probably be a good idea. Feel free to open new GitHub issues if you end up finding other stuff wrong.

PR #503, opened 2023-08-16
Revert BETA back to prior 1.9.0 value.

tmiw:
As part of #487, BETA was set to 0.75 to improve fast fading visualization. This PR reverts that change based on user feedback of version 1.9.0. All other waterfall changes from that PR seem okay (so far).


Tyrbiter:
This appears to have changed the spectrum display too, I find that the averaging needed was 2 with beta=0.75 but now it looks similar to the original now that beta=0.95 again.

I'd leave the averaging settings though, they're useful in themselves.

PR #506, opened 2023-08-17
Fix serial port contention when attempting to test PTT multiple times.

tmiw:
Additionally, this PR adds some resiliency to the Hamlib cleanup logic to hopefully avoid an intermittent crash I've been seeing on exit.


tmiw:
> Am I correct in thinking that this does not relate to use of net_rigctld in any way? My understanding is that rigctld/hamlib talks to the radio and freedv talks to rigctld so freedv has no direct communication with the serial port.

Correct. This is mainly an issue for those who configure hamlib in FreeDV to go directly to the serial port instead of through rigctld.

PR #518, opened 2023-09-01
Adjust length of vertical frequency ticks to prevent overlap with text.

### LINKED ISSUE #517: Waterfall frequency scale - a bit inconsistent and cluttered
I am not a heavy waterfall user but recently I have been using it more than I have previously and have noticed that it needs a bit of tidying up, in my humble opinion :) The actual waterfall is much improved, it's the labelling that is not great.

The frequency annotation and scale markings appear to be overlaid at the top of the waterfall, which makes them more difficult to read than the same information on the spectrum display which is presented below the display. In particular the 500Hz label and multiples of it are difficult to place as the longer marks are unclear on the waterfall, but are perfectly legible on the spectrum display.

I think it would be better to be consistent with the spectrum tab if possible, I don't think there is a convention for where the frequency scale is displayed on a waterfall, however both my IC-7300 and TS-890 have the finer markings below their waterfalls.

Worth looking at?

tmiw:
Fixes #517 by adjusting the size of the major and minor vertical frequency ticks on the waterfall plot. This brings the waterfall more in line with the spectrum plot in terms of (lack of) overlapping text.

Additionally, this PR also fixes an issue with the coloring of the tick marks on the spectrum plot when in dark mode.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -911,6 +911,8 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix issue with Voice Keyer file changes via Tools->Options not taking effect until restart. (PR #511)
     * Eliminate mutex errors during Visual Studio debugging. (PR #512)
     * Add timeout during deletion of FreeDVReporter object. (PR #515)
+    * Adjust vertical tick lengths on waterfall to prevent text overlap. (PR #518)
+    * Adjust coloring of text and ticks on spectrum plot to improve visibility when in dark mode. (PR #518)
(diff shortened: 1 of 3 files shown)


Tyrbiter:
This is an improvement, but the long tick marks still touch the characters in at least one place.

Can the wxWidgets pixel offsets be adjusted in steps less than 5? A couple of extra pixels shift would help.

tmiw:
How about now?

Tyrbiter:
That has nailed it, a visible gap between the tick and the text.

Thanks @tmiw

Issue #417, opened 2023-06-06
Load different configuration without restarting the app


tmiw:
Basically something similar to what WSJT-X does where you can load a separate configuration while FreeDV is still running (i.e. no need to re-execute the process with `-f` as we do now). It may help with situations such as switching between a WebSDR and a transceiver for receive.

tmiw:
Closing as it should go through the existing feature request process.

PR #519, opened 2023-09-01
FreeDV Reporter usability enhancements

### LINKED ISSUE #516: More options for the FreeDV (int) Reporter
Since the QST article about FreeDV has been published it has become clear that the reporter needs to have more fine grained setting options.
It's great to see more stations on the bands, but on 20m especially the reporter is becoming difficult to read in use.
Suggestions:
1 A filter to remove calls that have not been updated in the last N mins.
2 An option to only display (or highlight?) calls who have received me in the last N mins).
3 Show my own call line in a different colour.
4 An option to switch the reporter band filter automatically to the currently reported band (but only when that changes). i.e. The reporter band selector would still over-ride.

Just few ideas for discussion ;)

tmiw:
Resolves #516 by adding the following to the built-in FreeDV Reporter client to improve usability:

* Add support for sorting columns in FreeDV Reporter.
* Add Distance column (vs. currently configured location).
* Highlight just-received rows in ~green~ light blue (like with web version).


barjac:
Working OK, except that 'N/A' sorts above latest TX time ;)
Where would '--N/A--' sort I wonder? Is 'N/A' really needed at all?

tmiw:
I addressed the concern with the date sorting as well as added green highlighting for RX rows (like with the web version) and a Distance column. Seems to work over here but let me know how it goes. 👍

barjac:
Super! That's really nice. Good work Mooneer.
Can you think of a way to reduce the width of these as the table is now too wide again ;)
Headings of Last RX callsign and Last RX mode
Maybe change Frequency heading to Freq MHz (or just MHz) and remove 'MHz' from each line?
Maybe change Distance headindg to Dist km and remove km from each line?

Screenshot_20230902_190354

tmiw:
Does the change I just pushed help a bit with the widths?

Tyrbiter:
For the distance column, I see it says (km), but what about a locale with miles as unit of distance?

tmiw:
> For the distance column, I see it says (km), but what about a locale with miles as unit of distance?

Is that anywhere other than the US or UK? On one hand it's just those two places, but on the other, a lot of hams are in the US.

barjac:
> Does the change I just pushed help a bit with the widths?
Will test shortly.
> 
> BTW what are your thoughts about changing the RX background color to the same color as used for the VK button? This would be done on the web version too and would mainly be for colorblindness accessibility.

Yes that would be OK for me, however I think the foreground should stay black as the white on blue or red background is less readable than black. (See screenshot)

I think the distance unit should be in misc settings :)

Tyrbiter:
> > For the distance column, I see it says (km), but what about a locale with miles as unit of distance?
> 
> Is that anywhere other than the US or UK? On one hand it's just those two places, but on the other, a lot of hams are in the US.

I am not entirely sure, but the only other places are probably insignificant, 2 are Liberia and Myanmar. As @barjac suggests, perhaps a setting somewhere. I can do km, but miles work best for me.

barjac:
Yes the spacing now works OK bearing in mind my long version.
Screenshot_20230902_203434

barjac:
I think the sort settings should be remembered between sessions.

barjac:
I notice that the sort is not updated on changing band filter.

tmiw:
I'm not 100% sure about the change I made for the sorting on filter change (pretty much everyone was on 14.236 MHz at the time of testing), but I believe I've addressed all other concerns thus far. The setting for miles vs. km is here:

Screenshot 2023-09-03 at 1 34 45 PM

Tyrbiter:
Very small truncation in the Options->Reporting tab distance setting



Everything else seems good.

barjac:
I just see:
Screenshot_20230904_003428

tmiw:
My Linux VM looked more like @Tyrbiter's. Hopefully the recent change I pushed helps @barjac too.

barjac:
> My Linux VM looked more like @Tyrbiter's. Hopefully the recent change I pushed helps @barjac too.

Not quite yet - can the settings dialogues be made variable geometry so they can be stretched for different screen aspect ratios?

Screenshot_20230904_124054

barjac:
https://github.com/drowe67/freedv-gui/pull/519/commits/fae9808bfff46be4981b2693f5a2792f850d18f4

This is saving the sort column but not the sort direction. If I close while sorting 'Last TX' with latest at the top, on re-start program and Start modem the latest is at the bottom.

barjac:
Not sure if this should be another enhancement bug or not.
This causes a second entry in freedv reporter if they have 'reporting' enabled. This is undesirable as the report is effectively fake (not from their own station) and clutters the reporter.
However if they disable reporting then they cannot use the internal reporter on that machine as it stops working with 'reporting' unchecked.

I'm sat here with my laptop with no rig or sdr running and would like to use the internal reporter to monitor activity on the bands but I can't without 'reporting' and creating a useless entry on both web and int. reporters. :(

barjac:
This seems excellent now (apart from the sort order polarity on restart). We had a good net today with 5 stations and with the sort set to 'Last TX' with most recent at the top it was very useful to see stations in order of appearance!
I have been having some internet connection issues and slow connections as well, so I think the duplicate calls that I reported were down to my end and not software.
If there is no use case for running two instances of FreeDV at the same time (on the same machine) then it may be worth adding a warning when attempting to start the second instance as it gets very confusing! I accidentally saw myself twice on the reporter and it was only when I stopped FreeDV that I found the other instance running! /o\

tmiw:
> Not quite yet - can the settings dialogues be made variable geometry so they can be stretched for different screen aspect ratios? My desktop monitor is 1280x1024 which shows this problem but on a wider aspect screen laptop it looks fine.

I removed the forced sizing and it seems to work on Windows and Linux (albeit both in VMs on a 16:10 laptop display). Any improvement on your end?

> "Save last sorted column and restore on next startup." This is saving the sort column but not the sort direction. If I close while sorting 'Last TX' with latest at the top, on re-start program and Start modem the latest is at the bottom.

Fixed.

> Not sure if this should be another enhancement bug or not. Many people use a second machine to run freedv to monitor a web sdr whilst operating. This causes a second entry in freedv reporter if they have 'reporting' enabled. This is undesirable as the report is effectively fake (not from their own station) and clutters the reporter. However if they disable reporting then they cannot use the internal reporter on that machine as it stops working with 'reporting' unchecked. Can the internal reporter be made to report activity when not itself reporting?
> 
> I'm sat here with my laptop with no rig or sdr running and would like to use the internal reporter to monitor activity on the bands but I can't without 'reporting' and creating a useless entry on both web and int. reporters. :(

Doesn't the second machine/instance show up as "RX Only" in that case? That might be a way for people to differentiate between the two. (To make it report "RX Only", just set each of the TX sound devices to "none" in Tools->Audio Config.)

But yeah, it'll be a more involved change to have the internal one work regardless of whether it's running so I'd say make it a separate issue.

> If there is no use case for running two instances of FreeDV at the same time (on the same machine) then it may be worth adding a warning when attempting to start the second instance as it gets very confusing! I accidentally saw myself twice on the reporter and it was only when I stopped FreeDV that I found the other instance running! /o\

With virtual sound devices/audio routing the web SDR use case mentioned above is an example of where having two instances running on the same machine would work. There are also other SDRs (e.g. Flex) that support multiple audio streams and could potentially TX on either instance, so one FreeDV process per band could work well.

Tyrbiter:
The sizing thing is now fine on my Linux machine, so that is sorted out.

The multiple copies issue probably ought to be part of the multiple configuration issue, https://github.com/drowe67/freedv-gui/issues/417 was created a while ago, I failed to make my attempts at it work about 2 years ago so clearly my coding skills are inadequate.

barjac:
This is now fine for me - excellent job Mooneer!

PR #529, opened 2023-09-07
Connect to FreeDV Reporter even when no session is running

### LINKED ISSUE #527: Allow running FreeDV Reporter (internal) when not reporting.

This causes a second entry in freedv reporter if they have 'reporting' enabled.

This is undesirable as the report is effectively fake (not from their own station) and clutters the reporter.

If they disable reporting to avoid the above then they cannot use the internal reporter on that machine as it stops working with 'reporting' unchecked.

tmiw:
Resolves #527 by implementing logic (client and server side) to enable hiding of our own connection from other users (and showing it again). Configuration changes for FreeDV Reporter also take effect immediately, but can still only be done while a session isn't running.


tmiw:
@barjac, give this a shot?

Tyrbiter:
I have this built and installed now, it seems to work as expected. Very fine work @tmiw

barjac:
+1 agreed - seems to work perfectly, nice! :)

PR #534, opened 2023-09-12
Add FreeDV Reporter option to have the band filter track the current frequency

### LINKED ISSUE #530: Option to have the internal  Reporter band filter track the Reported Frequency.
My thoughts are that with this option enabled that the reporter band filter would track the Reported Frequency no matter whether this is selected in FreeDV (when hamlib is used to control the rig), or whether it is reported to FreeDV from the rig when in 'read-only' mode.

Changing the band selector manually in the Reporter would still over-ride the tracking until either the modem is re-started and/or the Reported Frequency changes. Effectively this means that only 'changes' in Reported Frequency need to trigger an update of the band filter.

This could be a check box on the 'control bar' in the reporter. Its state should survive re-launches of FreeDV.

When the 'Track' check box is un-checked operation would be as now.

tmiw:
Resolves #530 by adding a new checkbox in the FreeDV Reporter window to allow the band filter to track the "Report Frequency" field on the main window. Example usage:

Screenshot 2023-09-11 at 11 35 40 PM

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -935,6 +935,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
         * *NOTE: Requires 'Enable Experimental Features' to be turned on, see below.*
     * Adds configuration item allowing optional use of experimental features. (PR #497)
         * This option is called "Enable Experimental Features" in Tools->Options->Debugging.
+    * Add FreeDV Reporter option to have the band filter track the current frequency. (PR #534)
 3. Build system:
     * Upgrade wxWidgets on binary builds to 3.2.2.1. (PR #531)
     * Fix issue preventing proper generation of unsigned Windows installers. (PR #528)
--- a/src/config/ReportingConfiguration.cpp
+++ b/src/config/ReportingConfiguration.cpp
@@ -41,6 +41,7 @@ ReportingConfiguration::ReportingConfiguration()
     , freedvReporterHostname("/Reporting/FreeDV/Hostname", wxT(FREEDV_REPORTER_DEFAULT_HOSTNAME))
     , freedvReporterBandFilter("/Reporting/FreeDV/CurrentBandFilter", 0)
     , useMetricDistances("/Reporting/FreeDV/UseMetricDistances", true)
+    , freedvReporterBandFilterTracksFrequency("/Reporting/FreeDV/BandFilterTracksFrequency", false)
         
     , useUTCForReporting("/CallsignList/UseUTCTime", false)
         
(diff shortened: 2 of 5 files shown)


barjac:
Perfect! Working exactly as I envisaged in Linux with nothing to say other than well done again!!

Tyrbiter:
This latest change appears to have prevented the in app reporter showing anything with either reporting disabled or with the modem not running.

Probably a logic bug with a simple fix I hope.

tmiw:
Can you provide a list of steps to run to duplicate?

Tyrbiter:
Previously without the ms-reporter-track-freq changes I could use the in-app reporter with the reporting disabled in Tools->Options and see stations with and without the modem running.

Now I cannot see anything in the reporter window when FreeDV Reporter is disabled, the modem state makes no difference to this.

tmiw:
> Previously without the ms-reporter-track-freq changes I could use the in-app reporter with the reporting disabled in Tools->Options and see stations with and without the modem running.
> 
> Now I cannot see anything in the reporter window when FreeDV Reporter is disabled, the modem state makes no difference to this.

The current behavior is actually the correct behavior as the server still requires a callsign to log in. The FreeDV application just sends a command immediately after connecting to hide itself from the user listing (and sends another to show itself again once you push Start).

That said, nothing in this PR should have affected that, which is weird. /shrug

Tyrbiter:
OK, well it's definitely not what I had working before.

I sometimes want to be able to sit and listen to a QSO without other stations noticing my presence and wanting me to join in, I can't do that any more with this change.

It would be good if you can work out why the logic has changed.

tmiw:
Yeah, the way it's supposed to work is that you need both "Enable Reporting" and "Enable FreeDV Reporter" checkboxes checked (along with a valid hostname, callsign and grid square/locator) in order for the application to connect to FreeDV Reporter, even when a session isn't running. For instance, something like this:

Screenshot 2023-09-12 at 12 36 32 PM

Once connected, if a session isn't running, the FreeDV application sends a command to the server that basically removes itself from the FreeDV Reporter list. When you push Start, another command gets sent to re-add itself to the list. This prevents people from seeing you unless you have a session running.

Anyway, while testing I found an issue where it was connecting to FreeDV Reporter even with "Enable Reporting" unchecked. This should now be resolved in https://github.com/drowe67/freedv-gui/pull/535. Having it connect regardless of a valid configuration is going to be a bigger change, especially since the FreeDV Reporter window does have dependencies on having a valid configuration (such as for distance calculation). I'd rather we wait on that until after 1.9.2 if we do decide this is something that's useful.

Tyrbiter:
OK, I suspect that this is what I was seeing and somehow the Enable FreeDV Reporter checkbox (which is what I was changing) made a difference, but I think when I rebuilt with the previous change to track the selected band it overrode this.

The new tightening of the conditions to set up reporting properly will have covered over the original hole that I happened to find by accident.

Since I can do what I am asking for using the web reporter I don't see another change request as worth pursuing.

PR #542, opened 2023-09-17
Fix FreeDV Reporter regressions

### LINKED ISSUE #527: Allow running FreeDV Reporter (internal) when not reporting.

### LINKED ISSUE #540: 1.9.2 Regression. 'Use UTC Time' in reporter settings is ignored.
All in the summary.

tmiw:
Contains the following changes:

1. Updates FreeDV to always connect to FreeDV Reporter, regardless of whether there is a valid configuration (Resolves #527 
    * Options for FreeDV Reporter hostname and metric measurements are moved to a separate section in Tools->Options and are no longer dependent on whether reporting via FreeDV Reporter is enabled.
    * Reporting checkboxes are renamed to better describe their behavior.
    * FreeDV Reporter hostname defaults to qso.freedv.org if blank.
2. Fixes regression preventing proper display of "RX Only" for receive-only stations.

(This also implicitly resolves #540.)


barjac:
Looks really good now - will test live this afternoon. Thanks, super work!

PR #545, opened 2023-09-18
Add configuration for background/foreground colors in FreeDV Reporter

### LINKED ISSUE #541: Option to change internal reporter Transmitting and Receiving colours in settings.
I have had a report today from an elderly user who struggles to read the coloured highlighted lines in the internal reporter.

It's nothing to do with colour blindness, just lack of contrast as I understand it.
If an option was available to maybe select paler colours (something like the web version) then it would help.

He finds black and white as foreground and background, or flipped, readable but not the heavily saturated coloured backgrounds.

A lot of this can be down to different operating systems and monitors, so some user control could be beneficial for many users myself included.

Any other thoughts on this?

tmiw:
Resolves #541 by adding configuration options in Tools->Options->Display to control the RX/TX background and foreground colors in FreeDV Reporter.

Example UI:

Screenshot 2023-09-17 at 6 24 33 PM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ endif()
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11" CACHE STRING "Minimum OS X deployment version")
 
 set(PROJECT_NAME FreeDV)
-set(PROJECT_VERSION 1.9.2)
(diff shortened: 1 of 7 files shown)


barjac:
Super! Does what it says on the tin! That should keep him happy :) Thanks!

PR #549, opened 2023-09-21
UX improvements for "On Top" menu option

### LINKED ISSUE #546: File->On Top does not seem to do anything
As above, tried to get main window on top of reporter window. Selecting this menu entry does nothing and there is no indication it is active.

tmiw:
Resolves #546 by doing the following:

1. On Linux (or any other platform using GTK), hide the "On Top" menu option as it's not reliable (note: [support was even removed entirely in GTK4](https://discourse.gnome.org/t/gtk-4-how-to-replace-gtk-window-set-keep-above-and-gtk-window-set-keep-below/3550) for this reason).
2. For other platforms, rename "On Top" to "Keep On Top" and indicate whether it's enabled via a checkmark next to the menu item. Example below:

Screenshot 2023-09-20 at 10 09 39 PM


barjac:
I think the common term for this is 'Keep above others' as in KDE Plasma5.

tmiw:
Windows calls it "Always on Top", IIRC. I don't know if macOS uses the same or different terminology for this. We'd probably want to use one of those if we were to use something other than "Keep On Top".

Issue #539, opened 2023-09-16
Allow configuration of hamlib (net rigctld) in RX only systems to set reporter band automatically over LAN.


barjac:
The summary explains this adequately I think.

However long explanation:
The main system running TX/RX instance of FreeDV is wired to the radio via USB.
Hamlib rigctld daemon on localhost communicates with FreeDV.

Another instance of FreeDV running on another system could also receive the rig frequency, mode etc. from the main system over the LAN, however FreeDV does not currently cater for the configuration of this.

I already run a logging program alongside FreeDV that uses the same data via localhost:4532

Can configuration be added to facilitate this by adding settings hostname/IP:port and a way to enable this?

This also depends on #527 being fixed as the main reason for this would be to enhance that feature.

tmiw:
You should theoretically be able to provide the IP address of the other machine instead of localhost/127.0.0.1 in the Hamlib settings, i.e. "192.168.1.123:4532". Assuming Hamlib is okay with that setting, FreeDV should then pull the frequency/mode every 5 seconds or so. Is that currently producing an error or some other undesired behavior on either machine?

barjac:
As I see it there is only provision to enter a host:port in the PTT section, which for a receive only machine is greyed out with 'Use Hamlib PTT' unselected.
Selecting 'Use Hamlib PTT' to un-grey that section causes FreeDV to complain that it can't connect to the rig via Hamlib as it's looking for a serial device. Even using 'Hamlib  Dummy' as the radio it is the same. :(

tmiw:
Using "Hamlib NET rigctl" for the radio name shouldn't need a serial device, especially if you choose "CAT" for PTT instead of DTR or RTS. YMMV if you're choosing some other radio name instead, though. Can you confirm that it's still expecting a serial device with "Hamlib NET rigctl"?

barjac:
These two images are with and without 'Use Hamlib PTT' selected. However this request is not related to PTT as this machine does not have any TX audio settings and does not use any PTT.

In 'monitoring' mode neither does it have a 'Report Frequency' in the main dialogue as it's not reporting.

It does appear to be requiring serial info when the 'Use Hamlib PTT' is checked.

This PTT settings dialogue seems to be the wrong place for this to be configured.

![Screenshot_20230919_190147](https://github.com/drowe67/freedv-gui/assets/1324702/bf2ff012-acd3-445b-96fb-e3359f4f2521)

[...]

PR #550, opened 2023-09-21
Improve labeling of PTT/CAT control options.

tmiw:
Improves labeling of CAT control/Hamlib related options based on feedback in #539:

1. Rename "PTT Config" to "CAT and PTT Config" in the Tools menu:

Screenshot 2023-09-20 at 10 20 59 PM

2. Change title of PTT Config dialog and change checkbox for Hamlib to clarify that it also handles CAT control (along with PTT):

Screenshot 2023-09-20 at 10 21 03 PM


tmiw:
> When setting up a RX only system there is no 'None' option for 'PTT Uses:' which is very confusing. Maybe that box should be greyed when audio settings for TX are none. Although when an install is occasionally used for both RX only and TX, I think that the TX audio settings should stay and have a 'enable TX' check box somewhere to avoid constantly changing the TX audio configs. WDYT?

I think it might be confusing to gray out that setting if there are no transmit sound devices set. (The user could very well be setting Hamlib settings first, before audio devices.) Also, most people don't need to change that setting, anyway; it's there for the oddball radios that don't have working CAT support for PTT.

> It would be sensible to have this covered by the enhancement request for #417 multiple reloadable settings.

The labeling changes in this PR, or did you mean consolidating the config dialogs?

Tyrbiter:
> > It would be sensible to have this covered by the enhancement request for #417 multiple reloadable settings.
> 
> The labeling changes in this PR, or did you mean consolidating the config dialogs?

I was referring to @barjac comments on the issue with Rx only, confusion with config settings etc. If it is possible to create a PR for the  https://github.com/drowe67/freedv-gui/issues/417 enhancement then it would mean only doing the work once.

It seemed to me to be sensible to have templates of some kind for this, creating a new config would allow the selection of a Tx/Rx one, or Rx only. Additional templates could be added if needed. I realise that this is quite a big task, the wsjtx Configurations menu has been around a while, maybe something like it would give us a decent UI for this.

tmiw:
> I was referring to @barjac comments on the issue with Rx only, confusion with config settings etc. If it is possible to create a PR for the #417 enhancement then it would mean only doing the work once.
> 
> It seemed to me to be sensible to have templates of some kind for this, creating a new config would allow the selection of a Tx/Rx one, or Rx only. Additional templates could be added if needed. I realise that this is quite a big task, the wsjtx Configurations menu has been around a while, maybe something like it would give us a decent UI for this.

I think if we were to do any consolidation, Easy Setup would likely get rolled into Tools->Options, with the existing PTT and audio device settings still accessible via their respective Advanced buttons. That would still be a lot of work and I'd like to do some refactoring and other cleanup first before attempting.

Tyrbiter:
Yes, I wasn't suggesting that you should just dive in and hack everything about.

Thought required, and you have clearly already done some thinking about it.

barjac:
> I think it might be confusing to gray out that setting if there are no transmit sound devices set. (The user could very well be setting Hamlib settings first, before audio devices.) Also, most people don't need to change that setting, anyway; it's there for the oddball radios that don't have working CAT support for PTT.

OK understood , so could there be a 'none' option in the list?
I have scratched my head looking at this setting while setting up a system with no transmit.
It's asking for something impossible as no option is applicable.

> > It would be sensible to have this covered by the enhancement request for #417 multiple reloadable settings.

Yes but it still needs to make sense.

tmiw:
> OK understood , so could there be a 'none' option in the list? I have scratched my head looking at this setting while setting up a system with no transmit. It's asking for something impossible as no option is applicable.

The typical way of setting up a RX only station is to not specify any TX sound devices, but adding None probably won't hurt: https://github.com/drowe67/freedv-gui/pull/556

barjac:
Not had chance to check fully but selecting 'None' seems to be a switch from transceive to receive only mode without needing to touch the audio settings.
If I have the above correct then it's a nice feature but not labelled as such.
It would be hard to find without reading docs and who does that? ;)
I really think that the feature needs a separate 'Receive Only' check box maybe above the 'PTT Uses:' (or somewhere else entirely) that when enabled greys out 'PTT Uses:'.

Tyrbiter:
@tmiw mentioned it in this PR

https://github.com/drowe67/freedv-gui/pull/556

tmiw:
What about renaming "None" to "None (RX Only)" in that drop-down?

But yeah, the user manual could probably be improved in general. not just when talking about this feature. PRs to update it are welcome :)

Issue #464, opened 2023-07-09
Freedv Reporter (not web) starts to duplicate and triplicate users after a long session. 


barjac:
This applies to a build of ms-preserve-reporter-layout branch yesterday.
After about an hour in QSO with two other stations I began to notice that I and the other stations were appearing twice and then three times in the list.
After signing I closed and restarted the reporter and the excess entries cleared.
Could this be a feature to preserve reporting for all members of the group? I doubt it since I was seeing lots of other stations duplicated.

tmiw:
I wonder if it was some sort of temporary network glitch on the server side or somewhere between it and you. It might be worth pinging qso.freedv.org and seeing if there are any dropouts when this happens.

Anyway, if it's operating correctly the duplicates should drop off on their own within a minute or so if it is indeed some sort of network glitch. Let me know if that's not actually happening.

barjac:
I will build master and test again during the net later today.
When I noticed this it was there for much longer than a few minutes, so it did not seem to be clearing itself, only when I finally killed the reporter and restarted it were all the many duplicates removed.
I have been suffering with occasional slow www speeds recently, not sure about during the OSO, but there was nothing else apparent in the reporter. Everyones TX was indicated instantly and SNRs were being updated OK.
Remember that I was using a ms-preserve-reporter-layout branch build if this is relevant.

barjac:
Current master in QSO now:
![Screenshot_20230710_160632](https://github.com/drowe67/freedv-gui/assets/1324702/228ce46f-c146-442e-80f1-4253a87e0dc4)

barjac:
Now there are 3 of me!

![Screenshot_20230710_161537](https://github.com/drowe67/freedv-gui/assets/1324702/37e2ea90-972d-47a7-a130-3688a55f02d6)

barjac:
And this is at the end of the 1 hour net.

![Screenshot_20230710_170115](https://github.com/drowe67/freedv-gui/assets/1324702/54cb9f06-0eaa-4219-8f10-ce42064e530f)

tmiw:
OK, I think I misunderstood what was happening. I thought this was something that was happening on the website, too.

Anyway, I was able to reproduce the issue by pushing Start and disabling IPv6 on my laptop afterward (to force a reconnect onto IPv4). #468 should fix it; turns out the socket.io client doesn't call close and open when reconnecting, just the "reconnect" event--so the necessary GUI event handlers to kick off the refresh weren't firing.

[...]

PR #557, opened 2023-09-28
Client disconnects should delete from the internal data store no matter what.

tmiw:
Possible followup fix for #464. Basically, the current code only deletes users from the internal data store on disconnect if they're already displayed on the screen. Good when everyone was always displayed, but now that this can't be assumed anymore, causes problems with duplicate entries (especially when users end up disconnecting and reconnecting).


barjac:
Yes!
Today I left the RPi4b running this branch all day along with my main system running master.
The main system has lots of multiple entries, the RPi has none so unless there any unexpected side effects this has fixed it. :)

barjac:
On another slightly related subject, it seems that the 'receiving' colour highlight seems to turn off randomly (or maybe 10s after the last update?) irrespective of whether there is a loss of decode or not.
I have heard users commenting about the colour continually turning on and off during overs, despite good conditions and solid signals.
This is causing some confusion.
I don't know how this works, but if it is a simple timeout then maybe it could be longer?

tmiw:
> On another slightly related subject, it seems that the 'receiving' colour highlight seems to turn off randomly (or maybe 10s after the last update?) irrespective of whether there is a loss of decode or not. I have heard users commenting about the colour continually turning on and off during overs, despite good conditions and solid signals. This is causing some confusion. I don't know how this works, but if it is a simple timeout then maybe it could be longer?

Yeah, it's a 10 second timeout, similar to the web version. Suggestions on how long it should be?

barjac:
If it was double the expected update period then it would be unlikely to drop between updates I guess.
However the updates appear to be quite random and sometimes only a few seconds.
Is there an explanation like a logic diagram of how this works without trying to read the sources? ;)

tmiw:
I'll go ahead and spin up another PR real quick after merging this one.

tmiw:
https://github.com/drowe67/freedv-gui/pull/558

PR #558, opened 2023-09-29
Increase RX coloring timeout to 20s and use UTC for time comparisons.

tmiw:
This PR updates the RX coloring code for the internal version of FreeDV Reporter as follows:

1. Increases the time that the coloring is displayed without an additional RX report from 10 seconds to 20 seconds. 
2. Time comparisons for determining whether to hide the coloring are now done explicitly using UTC time to ensure that differing time zones don't cause the coloring to prematurely be removed.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -925,6 +925,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Add configuration for background/foreground colors in FreeDV Reporter. (PR #545)
     * Always connect to FreeDV Reporter (in view only mode if necessary), regardless of valid configuration. (PR #542, #547)
     * Add None as a valid PTT method and make it report RX Only. (PR #556)
+    * Increase RX coloring timeout in FreeDV Reporter to 20 seconds. (PR #558)
 3. Documentation:
     * Add information about multiple audio devices and macOS. (PR #554)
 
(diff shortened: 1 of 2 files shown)


barjac:
I now have both systems running this branch and should be able to test with other stations later today.

I just noticed that when I change frequency my 'last transmit' time still reports for the old frequency. I have not transmitted on the new frequency so this could be misleading.
Maybe the 'last transmit' time should be cleared on frequency change?

Tyrbiter:
> I just noticed that when I change frequency my 'last transmit' time still reports for the old frequency. I have not transmitted on the new frequency so this could be misleading. Maybe the 'last transmit' time should be cleared on frequency change?

That last transmit time is band independent, it's useful for guessing if a station is monitoring or prepared for QSOs. I'd prefer it to remain as it is, the alternatives involve extra complexity but maybe there could be a frequency label on it to indicate band but not exact frequency.

tmiw:
> > I just noticed that when I change frequency my 'last transmit' time still reports for the old frequency. I have not transmitted on the new frequency so this could be misleading. Maybe the 'last transmit' time should be cleared on frequency change?
> 
> That last transmit time is band independent, it's useful for guessing if a station is monitoring or prepared for QSOs. I'd prefer it to remain as it is, the alternatives involve extra complexity but maybe there could be a label on it to indicate band but not exact frequency.

Adding a frequency label would likely add a fair bit of complexity as the current UI control used for the FreeDV Reporter window doesn't support multi-lined rows/columns. We'll leave for now and revisit if there is similar feedback from others about the last TX time.

barjac:
This PR seems a reasonable compromise, certainly now the RX colour remains steady for all receiving stations.

PR #560, opened 2023-10-01
Add decimal separator force to "." dot for frequencies translation on…

Artemia76:
On French Computer, locale use the comma as decimal separator. This make rig exchange translate frequencies without the decimal part { 14,236 -> 14.000) } by using wxWidget::Format when using Frequency reporter

I add a small check and force locale for numeric decimal separator to dot "." to avoid

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -210,6 +210,14 @@ bool MainApp::OnInit()
     // Initialize locale.
     m_locale.Init();
 
+    // Check System language to force dot "." as decimal separator
+    int sys_lang = wxLocale::GetSystemLanguage();
+    if( sys_lang != wxLANGUAGE_DEFAULT )
+    {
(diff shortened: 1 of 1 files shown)


tmiw:
Thanks for this code change. I'm not comfortable with just forcing C locale for number processing, especially when this behavior would likely be different from other applications on this system. I created an alternative PR (https://github.com/drowe67/freedv-gui/pull/561) that updates the frequency list to display properly for the user's current locale. For example, with this PR and forcing `LANG` to be `fr_FR.utf8`, the frequency list now looks something like this:

Screenshot 2023-10-01 at 11 54 23 PM

Let me know how that works for you. If that solves your issue, we'll go ahead and close this PR.

tmiw:
Looks like #561 is good so I'm going to go ahead and close this PR. Let me know if you have any questions.

PR #589, opened 2023-10-30
Fix bug preventing 60 meter frequencies from using USB with DIGU/DIGL disabled

tmiw:
This PR fixes a bug reported on the digitalvoice list where when "Use USB/LSB instead of DIGU/DIGL" is checked, the specific check to force USB for 60 meters is missing. This results in the radio improperly switching to LSB when a 60 meter frequency is selected or entered.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -913,6 +913,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 1. Bugfixes:
     * Fix bug preventing frequency updates from being properly suppressed when frequency control is in focus. (PR #585)
+    * Fix bug preventing 60 meter frequencies from using USB with DIGU/DIGL disabled. (PR #589)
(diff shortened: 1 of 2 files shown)


barjac:
I have not had time to test this PR but there is an occasional specific use case for using LSB on 60m in the UK, so 'forcing USB' would not be welcome. We do normally use USB.
If this comment is not relevant please ignore it. I am maybe taking the above out of context while quickly glancing through the mail while eating lunch. 73 Barry.

Tyrbiter:
Yes, occasionally LSB is used to avoid the Olivia users on 5.3665MHz

We could disable the red mode indication for this case, but that's really a minor issue. It is a useful reminder that USB is the expected mode on 60m.

tmiw:
Yeah, USB is standard but this shouldn't prevent anyone from manually setting their radios to LSB as needed.

PR #596, opened 2023-11-01
FreeDV Reporter: Add support for filtering the exact frequency.

### LINKED ISSUE #587: Option to allow Internal Reporter to filter on current frequency.
Currently the reporter can filter on the band in which the current frequency is operating.

This works well but when a net is operating on a specific frequency in a band and other stations around the world are using the same band on different frequencies the display becomes cluttered.

This enhancement request is to optionally allow tighter control of the filter down to maybe the current frequency +/- 250Hz.

Maybe the 'Track current frequency' box could be replaced with 'Track band' and a new check box for 'Track frequency' added to track exact frequency?

tmiw:
Resolves #587 by adding an additional option for filtering on the exact frequency. Example:

Screenshot 2023-10-31 at 9 47 14 PM


barjac:
That looks ideal Mooneer, nice work as always :)
I can't test today as XYL is finally coming out of hospital at last after six weeks!

barjac:
No it's not working for me building from this branch.
The GUI has not changed from 1.9.4 :(

Tyrbiter:
That git hash looks a bit odd @barjac, it builds for me, displays the radio buttons and I am still testing.

tmiw:
Yeah, I'd try deleting your `build_linux` folder and trying to build again.

barjac:
> Yeah, I'd try deleting your `build_linux` folder and trying to build again.
I don't use build_linux, never have.
I clone the repo, checkout the branch by name, create a tarball from it and build a proper package for Mageia from it.
The git# is the last commit in the branch at the time I built it.
I can re-build it again but the whole thing is scripted and was done the same way as I always test pull requests. Very odd.

tmiw:
> > Yeah, I'd try deleting your `build_linux` folder and trying to build again.
> 
> I don't use build_linux, never have. I clone the repo, checkout the branch by name, create a tarball from it and build a proper package for Mageia from it. The git# is the last commit in the branch at the time I built it. I can re-build it again but the whole thing is scripted and was done the same way as I always test pull requests. Very odd.

Indeed. A full rebuild would be good. It might also be worthwhile to make sure that the code it did download has the changes in the PR, just in case it's some sort of code fetch issue.

barjac:
Yes my error - sorry. I forgot to remove a specific git# from the tarball script so it used that.
Never pays to rush the job.
Internet playing up as well so cannot get current github clone at present :(

Tyrbiter:
Not sure what is happening, this may have caused problems with updates to the reporter, both app and web are showing slow or non-existent last update times. I suppose it could be a server issue.

barjac:
Hi Mooneer, Sorry but this is really not working. The GUI is fine now I have rebuilt from the correct sources, but the functionality is not there in many different ways with some existing functions also broken.
I am testing this in x86_64 and aarch64.
The first major item is that it appears to behave as though 'Tools->Options->Rig Control->Enable frequency and mode changes' is checked when it is not.
It simply does not read (read-only) the rig frequency.
I suspect that maybe un-breaking this may fix all the problems. :)

With only one system updated to this PR the other was working as before when the updated one was broken. Both were previously tracking the band correctly before update.

Sorry I don't have more time to explain this in more detail just now.

tmiw:
I think the frequency/mode update stuff is working a bit better now, which will probably help with how this PR behaves. FWIW, it is tracking frequency for me but I also don't run a RX only version of FreeDV that grabs frequency from a second machine. /shrug

Tyrbiter:
With these latest changes @tmiw would you expect the Last Update column to stop updating every 5 seconds? Or is this a server side change to reduce the CPU load and only reflect the most recent change of details for each station?

tmiw:
> With these latest changes @tmiw would you expect the Last Update column to stop updating every 5 seconds? Or is this a server side change to reduce the CPU load and only reflect the most recent change of details for each station?

Not intentional, so I restored the every 5s updates.

Tyrbiter:
> > With these latest changes @tmiw would you expect the Last Update column to stop updating every 5 seconds? Or is this a server side change to reduce the CPU load and only reflect the most recent change of details for each station?
> 
> Not intentional, so I restored the every 5s updates.

That's fixed it for me, but many other stations connected are not showing the 5 second update rate. Did this get into the last release version somehow?

tmiw:
> That's fixed it for me, but many other stations connected are not showing the 5 second update rate. Did this get into the last release version somehow, I see updating from v1.9.3 users but not v1.9.4?

Yeah, that's what I'm thinking--that it got into v1.9.4.

barjac:
Looks good now at #21737 :)

It's now working as expected on both x86_64 and aarch64.
A second system on the LAN picks up the frequency from the radio (read only) correctly with it's reporting turned off as before on clicking 'Start'.

A minor GUI issue is that the Frequency and Band radio buttons are not active when 'Track Current' is already checked at program start. The 'Track Current' check box needs to be clicked twice to activate them.

tmiw:
> A minor GUI issue is that the Frequency and Band radio buttons are not active when 'Track Current' is already checked at program start. The 'Track Current' check box needs to be clicked twice to activate them.

Should be fixed in 35f8133. 👍

barjac:
> > A minor GUI issue is that the Frequency and Band radio buttons are not active when 'Track Current' is already checked at program start. The 'Track Current' check box needs to be clicked twice to activate them.
> 
> Should be fixed in [35f8133](https://github.com/drowe67/freedv-gui/commit/35f8133b06ee9851a2379180ffabc1d116c2ab21). 👍

Great, I will test.

Tyrbiter:
I have been trying this out, including the last commit and it all seems to have calmed down and be working as expected for me.

Thanks @tmiw there is a lot of subtle interaction between various settings that takes time to notice. For various reasons I've not been on air for a while and hence things related to net use and transmitting have been less well tested here.

barjac:
Nice! #35f8133 has fixed the GUI issue.
I will test on air as time allows, this feature should make a big difference to legibility of the reporter.

barjac:
> 
> Not intentional, so I restored the every 5s updates.

I have been noticing recently that transmitting stations do not update during a transmission, is that also intentional? At first I thought RF was blocking the internet, but now unsure :)

Tyrbiter:
> > Not intentional, so I restored the every 5s updates.
> 
> I have been noticing recently that transmitting stations do not update during a transmission, is that also intentional? At first I thought RF was blocking the internet, but now unsure :)

It seems to be a bug that crept in during the 1.9.4 development, so 1.9.5 will be fixed but current git in this branch is back to the 5s updates again.

barjac:
All worked as expected during the 60m UK net where there were some other stations on slightly different frequencies, one reporting only 500Hz off our freq who was correctly filtered out.
Updating of all stations (about 7 of us) seemed solid with the highlights stable.
Looks good to go for me - thanks Mooneer :)

PR #599, opened 2023-11-02
Add ability to force RX Only reporting in Tools->Options.

tmiw:
Per request from the digitalvoice mailing list, this PR adds the ability to force RX Only reporting (and disable PTT/Voice Keyer buttons) through a new Tools->Options item:

Screenshot 2023-11-02 at 12 55 08 AM

The result on pushing Start is identical to what happens if only RX audio devices are configured or if Hamlib PTT mode is set to "None".


barjac:
This is something some UK users have been considering in advance of creating an enhancement request.
We think that another status of 'Web RX' should be added along with 'RX Only' and both settable in a more prominent location in the main GUI (seems preferred by most) or possibly the reporter bottom bar.
When a station wants to indicate that he is not wishing or able to TX he could easily set status to 'RX Only' and when a station is reporting and in a QSO or net while using a Web SDR he could indicate this by setting the 'Web RX' status so others don't assume the SNR is from direct reception at his own station.

Use cases:
Today as I joined the net late, was still otherwise engaged and set 'RX Only'.
When I became free to join I unset 'RX Only' and was called into the group.

One of the group has such a high noise level that he can only receive by monitoring a web SDR. To make this clear to other users, the 'Web RX' status would be very useful.
Note that this would only be relevant for stations with TX capability as reporting an unknown web SDR would be pointless. If he can transmit the other party can ask or not, but obviously would not wish to log such a QSO.

I hope this can be considered,
Thanks

barjac:
Saving the 'RX Only' setting in the reporting options with the modem running does not change the reported status until the modem is stopped and re-started.

tmiw:
> Saving the 'RX Only' setting in the reporting options with the modem running does not change the reported status until the modem is stopped and re-started.

Should be fixed now.

> This is something some UK users have been considering in advance of creating an enhancement request. We think that another status of 'Web RX' should be added along with 'RX Only' and both settable in a more prominent location in the main GUI (seems preferred by most) or possibly the reporter bottom bar. When a station wants to indicate that he is not wishing or able to TX he could easily set status to 'RX Only' and when a station is reporting and in a QSO or net while using a Web SDR he could indicate this by setting the 'Web RX' status so others don't assume the SNR is from direct reception at his own station.

I'm not fully sure about this one yet. Seems like the same could be accomplished by having two entries listed (one "RX Only" with the grid square/locator of the SDR and the other that's just "Receiving" with the grid square of the transmit location), right?

barjac:
> > Saving the 'RX Only' setting in the reporting options with the modem running does not change the reported status until the modem is stopped and re-started.
> 
> Should be fixed now.

Great!

> 
> > This is something some UK users have been considering in advance of creating an enhancement request. We think that another status of 'Web RX' should be added along with 'RX Only' and both settable in a more prominent location in the main GUI (seems preferred by most) or possibly the reporter bottom bar. When a station wants to indicate that he is not wishing or able to TX he could easily set status to 'RX Only' and when a station is reporting and in a QSO or net while using a Web SDR he could indicate this by setting the 'Web RX' status so others don't assume the SNR is from direct reception at his own station.
> 
> I'm not fully sure about this one yet. Seems like the same could be accomplished by having two entries listed (one "RX Only" with the grid square/locator of the SDR and the other that's just "Receiving" with the grid square of the transmit location), right?

That scenario was not considered.
We try to avoid having two reports showing for the same callsign in nets, which can happen when someone also monitors an SDR as a backup but is primarily working direct.
We encourage those users to turn off reporting on the monitoring system.

Would anyone notice the changed locator without 'Web RX' in the status I wonder?

I agree that more thought is required on this one, although I do think that a 'RX Only' toggle switch is needed in the Main GUI rather than only in the settings.

Tyrbiter:
> I'm not fully sure about this one yet. Seems like the same could be accomplished by having two entries listed (one "RX Only" with the grid square/locator of the SDR and the other that's just "Receiving" with the grid square of the transmit location), right?

I think it needs to be explicit that a Web Rx is in use, not quite sure how to indicate that but a change of locator is a bit too obscure unless you have a Maidenhead grid map in your head :)

tmiw:
I'd suggest creating a new issue with the Web status request so we don't clutter this one. As for being able to dynamically adjust RX Only, I'm leaning towards moving the option to the Reporter window, though it sounds like somewhere on the main window would be preferred. If so, where do you think is good? 

Additionally, I'm not sure how to handle TX related UI. Right now, if RX Only is detected (or forced), we disable the Voice Keyer and PTT buttons on start. That'd require doing the same if someone wants to change their RX Only status while running, right? That may be a bit more involved of a change, so we may want to think that one out before actually making the changes.

barjac:
See Issue #622 before getting bogged down in this. I think the main use case for needing to switch to 'RX Only' on the fly would be where someone wants to indicate that they have been called away. Using the proposed 'Msg' field would allow the station to simply display 'QRX' for example.

tmiw:
> See Issue #622 before getting bogged down in this. I think the main use case for needing to switch to 'RX Only' on the fly would be where someone wants to indicate that they have been called away. Using the proposed 'Msg' field would allow the station to simply display 'QRX' for example.

Cool. I'll consider that request.

In the meantime I'll go ahead and merge this in (after making a minor change to disallow changing RX Only while active).

PR #601, opened 2023-11-08
Add new 160m, 80m and 40m calling frequencies for IARU R2.

### LINKED ISSUE #580: Add US/Region 2 specific frequencies for 40 and 80 meters
Previously discussed on the digitalvoice list; creating here to keep it from getting lost. Selected frequencies should be usable by US General class licensees (the original reason for the request).

Tasks:

- [x] Come up with the frequencies.
- [x] Update configuration defaults in the FreeDV application to include these frequencies. (#601)
- [x] Update FreeDV Reporter, website, announcement templates to list frequencies.

tmiw:
Resolves #580 by adding the following frequencies to the default frequency list:

* 1997 KHz (160 meters)
* 3850 KHz (80 meters)
* 7197 KHz (40 meters)

The new 40/80m frequencies are intended to be used as alternates for IARU R2 for those who can't use the existing frequencies (i.e. US General class licensees).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -924,6 +924,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * FreeDV Reporter: Add support for filtering the exact frequency. (PR #596)
     * Add confirmation dialog box before actually resetting configuration to defaults. (PR #593)
     * Add ability to double-click FreeDV Reporter entries to change the radio's frequency. (PR #592)
+    * Add new 160m/80m/40m calling frequencies for IARU R2. (PR #601)
 
 ## V1.9.4 October 2023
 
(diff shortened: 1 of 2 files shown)


tmiw:
> I have patched this PR in, but I don't see the new frequencies in the drop-down list. I am guessing this is because I have added in some additional frequencies myself. I can add them manually of course, but maybe there should be an automatic process.

This just affects the default list, so there currently won't be changes unless you reset your configuration to defaults or if you install FreeDV brand new.

Tyrbiter:
That's what I thought given that there was no code added to identify an existing expanded list and insert the new ones.

Is there a need to help out an existing user, if we are happy that they know how to add new frequencies then this can be left alone.

tmiw:
> That's what I thought given that there was no code added to identify an existing expanded list and insert the new ones.
> 
> Is there a need to help out an existing user, if we are happy that they know how to add new frequencies then this can be left alone.

It might be a good idea to post something to the mailing list about it but if additional help is needed, we can always revisit this.

PR #607, opened 2023-11-12
Add Help button on the main window.

tmiw:
Per PLT discussion in this month's meeting, this PR adds a button on the main window to make it easier for users to get help. 

Example of how it currently looks (feedback welcome):

Screenshot 2023-11-11 at 7 59 34 PM

Also including an upgrade of wxWidgets for Windows/macOS binaries to 3.2.4 and changes to release 1.9.5.


Tyrbiter:
Thoughts:

Having this right below the PTT button could lead to unexpected trips to the browser :)   Maybe place it somewhere further away from the main modem controls.

Would it make sense to allow this button to be hidden once the user is familiar with freedv and has a working setup?

tmiw:
> Having this right below the PTT button could lead to unexpected trips to the browser :) Maybe place it somewhere further away from the main modem controls.

Any suggestions on where else it could go?
 
> Would it make sense to allow this button to be hidden once the user is familiar with freedv and has a working setup?

How would this be done? Something in Tools->Options that the user has to select and/or something more automated?

Tyrbiter:
I don't know if wxWidgets allows it, but could there be a Get Help button placed on the right of the menu bar? Not sure if you can mix a menu and an action button this way. Doing it this way would make my second suggestion unnecessary, although it could still have a setting in Tools->Options to hide it if the user chooses to.

The more I look at the current position the more I think it doesn't belong in the Control group, so if the above suggestion is not possible then how about putting it in an unlabelled box below the Report Frequency controls, maybe separated by more space than the usual amount?

tmiw:
> I don't know if wxWidgets allows it, but could there be a Get Help button placed on the right of the menu bar? Not sure if you can mix a menu and an action button this way. Doing it this way would make my second suggestion unnecessary, although it could still have a setting in Tools->Options to hide it if the user chooses to.
> 
> The more I look at the current position the more I think it doesn't belong in the Control group, so if the above suggestion is not possible then how about putting it in an unlabelled box below the Report Frequency controls, maybe separated by more space than the usual amount?

I'm thinking just another item in the Help menu. Right now the Help menu looks like this:

Screenshot 2023-11-12 at 7 53 59 PM

so perhaps having it be the very first item in that menu would be sufficient?

Tyrbiter:
That's possible, but is it obvious enough? David was keen on a button in plain sight.

tmiw:
How about having the button on the lower left instead?

Screenshot 2023-11-14 at 8 49 05 AM

Tyrbiter:
I think that's better, but since I have a shorter vertical window I get a second column with the level meter and the Assistance box just below it. Is there a way to anchor the Assistance box at the bottom left as shown in your image above, with Assistance decoupled from the stack of other boxes?

tmiw:
I tried a few things to get it to anchor but all that seems to have done is make the other panels expand out horizontally instead (and leave a bunch of space left over on the bottom left). Maybe putting the button to the top left instead will help?

Tyrbiter:
Sounds OK to me, but perhaps then my other suggestion of allowing the button and box to be hidden might be useful.

tmiw:
Or another option: we can hide the Stats box entirely unless "Enable Experimental Features" is tuned on? Not sure how valuable that stuff is for the new user.

Tyrbiter:
I use BER and the Sync value regularly, everything else very rarely. I doubt I am a normal case :)

It's hard to create rules like this as an experienced user, some time ago I could recognise 3 or 4 UK stations by their frequency offset ;-)

Maybe experimental features need to be more granular...

Tyrbiter:
I have now had a look at this and it seems good to me, it brings up the right anchor point in the web page and is nicely out of the way of the more used buttons in the main window.

I did notice that with the stats disabled the level meter is also not shown, which I think is probably not intentional?

Very small thing: the buttons on the left of the main window on Linux are touching the bottom of the box they are contained in, on the right side the PTT button has a small gap clear above the box outline, it would be nice to make this the standard arrangement.

tmiw:
I did notice that the sizing was still a bit off when using Ubuntu's wxWidgets (which is still 3.0 I think) and trying to make the window smaller, but forcing 3.2 to be used instead seems to fix that. I think that's what you meant by the level slider getting hidden as there's no specific setting to hide the Stats window right now.

Anyway, I tweaked the spacing and it mostly looks okay here. Occasionally the buttons end up touching the bottom of the box but I think that's a window manager/GTK issue more than anything.  Going to do some final checks here and then release 1.9.5.

PR #611, opened 2023-11-19
Fix #610 by using SetSize instead of SetClientSize.

### LINKED ISSUE #610: 1.9.5 left side of main window widgets vanish when resizing main window height, this did not happen in b5d73 before release.
A bug was introduced just before release of 1.9.5 that breaks the re-arrangement of the widgets on the left side of the screen.
They are removed from view so that stats can not be monitored nor the level meter.
I run the main window at down to half screen height, so this is a major problem.
I am seeing the same issue in x86_64 and aarch64.

tmiw:
`SetClientSize()` as was used in previous releases to restore window size doesn't seem to work well with the new Get Help button. This PR switches to using `SetSize()` instead (and subsequently also uses `GetSize()` to save the size of the window).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ endif()
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11" CACHE STRING "Minimum OS X deployment version")
 
 set(PROJECT_NAME FreeDV)
-set(PROJECT_VERSION 1.9.5)
+set(PROJECT_VERSION 1.9.6)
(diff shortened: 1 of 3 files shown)


barjac:
Yes, PR #611 seems to have fixed it, nice! Thanks again for your expertise :)

barjac:
Screenshot_20231119_210403

Tyrbiter:
> Yes, PR #611 seems to have fixed it, nice! Thanks again for your expertise :)

Indeed. I saw a few issues that seemed to have gone away so I decided to wait before reporting anything. @barjac beat me to it.

PR #620, opened 2023-12-10
Add logic to report status message to FreeDV Reporter.

### LINKED ISSUE #602: Add new 'Msg' column to reporter.
This is a request for a new 3 char wide column in the internal reporter to display up to any 3 ASCII characters.

A 3 character 'Msg' entry box in the reporter bottom bar could be used by any station to have the characters display on their reporter line.

Groups could decide on the content. Some may wish to use Q codes, other CW shorthand or even texting abbreviations or all of the above. 

This could make complex hard wired features as being discussed currently in PR#599 unnecessary.

e.g.:
In a net where someone has to leave urgently they could enter 'g2g' or 'QRT'.
Someone receiving via a web SDR could use 'WRX'.
One member of a group could ask what is my signal strength like? All the others could easily type in "S9+" etc. to give an instant report.
BRK could be very handy if used with care :)

The possibilities are endless and for just a 3 character width field it could be very useful.

I mentioned this idea in the UK net today and it was met with enthusiasm by all stations.

Not sure if this is possible to implement, but if it can then with a 5 second update cycle messages would be adequately fast.

tmiw:
Resolves #602 by adding functionality to the FreeDV Reporter window to allow live reporting of a user-defined status message. Example:

Screenshot 2023-12-10 at 2 28 31 PM


tmiw:
> Message is appearing in the app reporter, but not on the web version. Message column is showing after a js refresh.

Does holding down Shift while pressing F5 on the Web version improve this at all?

Tyrbiter:
> > Message is appearing in the app reporter, but not on the web version. Message column is showing after a js refresh.
> 
> Does holding down Shift while pressing F5 on the Web version improve this at all?

Not using Firefox 120.0.1, shift-refresh doesn't, shift-F5 opens a browser recording console window of some sort.

I also tried Chromium, that doesn't display the message text either, although it displays the Message column.

Tyrbiter:
I saw your message when I looked at the page without reloading it, but nothing from me showed up.

Refreshing it again cleared your message, so clearly there is something that needs fixing.

tmiw:
> I saw your message when I looked at the page without reloading it, but nothing from me showed up.
> 
> Refreshing it again cleared your message, so clearly there is something that needs fixing.
> 
> Editing my message showed it until I refreshed the page, so it's repeatable.

I think I found out what was causing this. Restarted the server with the required changes, hopefully that works a bit better.

Tyrbiter:
Server fix looks good.

Am I right in thinking that the message needs to be edited to be active? It seems that way. A previous message is not shown until I edit it, either in the app or on the web.

tmiw:
> Server fix looks good.
> 
> Am I right in thinking that the message needs to be edited to be active? It seems that way. A previous message is not shown until I edit it, either in the app or on the web.

Do you mean when you restart FreeDV? For me, restarting FreeDV brings back the previous message I entered.

Tyrbiter:
Yes, it does for me now, might have been down to whatever the server change did. I wonder if it would be useful to remind a user of the message somehow, or perhaps have an option to use the existing message on start.

tmiw:
> Yes, it does for me now, might have been down to whatever the server change did or maybe some changes with another PR merged in here. I wonder if it would be useful to remind a user of the message somehow, or perhaps have an option to use the existing message on start.

I'm not sure we need a separate config option for this. Maybe @barjac can chime in?

Tyrbiter:
I was thinking about something to indicate that a message is set when the reporter window is closed.

tmiw:
> I was thinking about something to indicate that a message is set when the reporter window is closed.

Since the FreeDV Reporter window state persists between executions (i.e. it reopens if it was open when you terminated FreeDV), seeing a message there when it comes back might be enough, right? I'm also presuming that a lot of the use cases for this (i.e. indicating that you're using a WebSDR for receive) won't necessarily make the previously set message invalid on next execution.

Anyway, we'll see what @barjac (and anyone else who happens on this PR) think before proceeding either way.

barjac:
Hi,
Thanks :)
It works communication wise, but for me even only three characters are off the screen because the text in the column is centred. The text in the Msg column could be left justified and the column label only needs three characters: 'Msg'.
This was my reasoning for using only 3 characters to keep the width of the reporter to a minimum for this aspect ratio of screen.
It is of course not an issue for high aspect ratio screens.
Maybe an option for users to select which columns to show in the reporter would help. Most users will not need to always see the 'Locator' column or 'Last update' perhaps.

I think more space saving in all columns needs to be made for it to work well as it stands.

Screenshot_20231215_152637

barjac:
> > I was thinking about something to indicate that a message is set when the reporter window is closed.
> 
> Since the FreeDV Reporter window state persists between executions (i.e. it reopens if it was open when you terminated FreeDV), seeing a message there when it comes back might be enough, right? I'm also presuming that a lot of the use cases for this (i.e. indicating that you're using a WebSDR for receive) won't necessarily make the previously set message invalid on next execution.
> 
> Anyway, we'll see what @barjac (and anyone else who happens on this PR) think before proceeding either way.

Maybe a 'Clear' or 'X' button to the right of the message entry box is needed, as the intention is that these messages in the main are temporary, so rather than having to insert the cursor and backspace over the last message a single click will remove it?

barjac:
Regarding space saving:
The Frequency column title could be 'MHz' (we all know that 'MHz' indicates frequency) and the values in each record could be only numeric. This would save repeating 'MHz' over and over.

Under 'Status' the word 'Receiving' is too long, 'RX' would suffice. Likewise 'Transmitting'  could be TX.

'RX Only' could be totally superseded by simply using 'RXO' in the Msg column.

'TX mode' heading label is forcing the column to be too wide for the data. 'Mode' would suffice.

The display of all the data looks rather scrappy due to the centre justification. It would be more readable if it was all left justified. Especially the 'Version' column.

@Tyrbiter
I limited my reported 'Version' git# to 5 chars to be no wider than 'FreeDV x.x.x-devel' however you now exceed that. Maybe FDV-x.x.x-devNN would suffice for devel versions?
I'm not sure yet where the 'FreeDV' comes from but I guess it's possible to patch it.

A lot to discuss but it's a great start and from comments on the air we are all looking forward to seeing this in action.

Tyrbiter:
Regarding the version reporting, I wonder if it would be sensible to display a tool tip with the version in future, perhaps when this has all settled down. At present it's useful to see when a station is using an older, potentially buggy version.

tmiw:
> 'RX Only' could be totally superseded by simply using 'RXO' in the Msg column.

"RXO" in the message field won't necessarily be obvious to the new FreeDV user.

> 'TX mode' heading label is forcing the column to be too wide for the data. 'Mode' would suffice.

Eh...we list modes for both TX and RX. How would one be able to tell which is which if the TX Mode column is shortened as suggested?

> The display of all the data looks rather scrappy due to the centre justification. It would be more readable if it was all left justified. Especially the 'Version' column.

Unfortunately I did some experimentation on this and it doesn't look like wxWidgets allows the column names to be center aligned while also having the column data itself be left-aligned. To me, forcing both column names and column data to be left-aligned looks worse, so I'm leaving the alignment as-is for now.

> @Tyrbiter I limited my reported 'Version' git# to 5 chars to be no wider than 'FreeDV x.x.x-devel' however you now exceed that. Maybe 'FDV-x.x.x-devNN' would suffice for devel versions? I'm not sure yet where the 'FreeDV' comes from but I guess it's possible to patch it.

IIRC "FreeDV" comes from the CMake configuration somewhere. Probably not a good idea to change this.

> Maybe a 'Clear' or 'X' button to the right of the message entry box is needed, as the intention is that these messages in the main are temporary, so rather than having to insert the cursor and backspace over the last message a single click will remove it?

If these messages are temporary, should they not be preserved at all in the configuration, then? I assumed you guys would want them to be but it should be straightforward to remove the save/load code for this if that'll be better.

Tyrbiter:
A compromise might be to have a clear button, then allow the last 2 or 3 messages to be selected from a drop-down list? A new message would remove the earliest one from the drop-down.

Of course, we could wait to decide until more people have had the chance to play with it live.

barjac:
> > 'RX Only' could be totally superseded by simply using 'RXO' in the Msg column.
> 
> "RXO" in the message field won't necessarily be obvious to the new FreeDV user.

How many characters do you envisage finally allowing? If more than 3 then 6 would cope with 'RXonly'.

> 
> > 'TX mode' heading label is forcing the column to be too wide for the data. 'Mode' would suffice.
> 
> Eh...we list modes for both TX and RX. How would one be able to tell which is which if the TX Mode column is shortened as suggested?

'RX mode' is to the right of RX call which makes fairly obvious, so 'Mode' in that position with a tooltip would I think be adequate. Space needs to come from somewhere :/

There is existing padding in all columns which could be reduced. I can do it manually in GUI headers. Can this be tightened up in the code?
This alone would possibly be enough gain assuming 3 characters (and 'Msg' as the column label).

> 
> > The display of all the data looks rather scrappy due to the centre justification. It would be more readable if it was all left justified. Especially the 'Version' column.
> 
> Unfortunately I did some experimentation on this and it doesn't look like wxWidgets allows the column names to be center aligned while also having the column data itself be left-aligned. To me, forcing both column names and column data to be left-aligned looks worse, so I'm leaving the alignment as-is for now.

OK, if we stuck to 3 characters then this may not be a problem, but at present the Msg field is so wide that 3 characters centred puts them so far to the right that they are off screen.

I think that the 'Msg' column should be to the right of the current 'Status' column in the centre of the display.

> 
> > @Tyrbiter I limited my reported 'Version' git# to 5 chars to be no wider than 'FreeDV x.x.x-devel' however you now exceed that. Maybe 'FDV-x.x.x-devNN' would suffice for devel versions? I'm not sure yet where the 'FreeDV' comes from but I guess it's possible to patch it.
> 
> IIRC "FreeDV" comes from the CMake configuration somewhere. Probably not a good idea to change this.

Maybe just remove it then for the reporter? 'x.x.x-devNN', x.x.x-123abc or x.x.x.
> 
> > Maybe a 'Clear' or 'X' button to the right of the message entry box is needed, as the intention is that these messages in the main are temporary, so rather than having to insert the cursor and backspace over the last message a single click will remove it?
> 
> If these messages are temporary, should they not be preserved at all in the configuration, then? I assumed you guys would want them to be but it should be straightforward to remove the save/load code for this if that'll be better.

I did say 'in the main' since some should be preserved like WRX (WEB sdr receiving), which for someone who has to use remote RX due to noise it will be a permanent message.
Maybe a 'default' message could be in settings, always used at program start but manually overwritable during a session from the GUI.
I think a 'clear' button and saving between sessions would then be optimum.
If nothing was saved at last close down then use the default if there is one.

barjac:
> A compromise might be to have a clear button, then allow the last 2 or 3 messages to be selected from a drop-down list? A new message would remove the earliest one from the drop-down.

An MRU list was in the original enhancement request, so yes, I would think 10 a more suitable limit.


> Of course, we could wait to decide until more people have had the chance to play with it live.

tmiw:
I managed to get a bit more space, I think. However, I'm not fully sure there's enough room for both a Clear and an Add button (the latter I would think being required to have a MRU feature). For example:

Screenshot 2023-12-18 at 8 39 57 PM

Additionally, the alternative for Add (simply implying it when one pushes Enter on the soon-to-be combo box) might not be 100% obvious, either.  Suggestions would be good here.

barjac:
Can we start with room for only 3 characters and call the column 'Msg' as originally intended?
Change the entry box label also to 'Msg'.
Change 'Show stations on:' to 'Show:'.
Change 'Frequency' to 'Freq'.
Change 'meters' to 'm' (in the band selector) and '6 meters and above' to '6m up'
Then see how space is going and get comments?

I don't think there is any need to add a second footer line and lose a line of data.
If there is any chance that shorter labels will confuse or be unclear then tooltips can be used to clarify - users only need to use them once or twice to become familiar.

The MRU could be just like the band selector. Click an item to use it.
'Clear' could be an 'X' to use minimal space.

tmiw:
How about now?

Screenshot 2023-12-19 at 7 23 50 AM

barjac:
Why is there a 'Set' button? It seems to do nothing, can it be removed and 'Clear' be just [X]? What about 'Show' and 'freq' would they not then allow a single footer line?

The 'Clear' button fails to clear the message entry box.
Anyone typing a long message will screw up everyone else's display - there must be a limit and I think ten chars is more than adequate, we only really wanted three to avoid all these issues.
As users type different length messages all user displays will jump around to accommodate the longest one. It needs to be a fixed width column at a sensible limit.
The Msg column seems to be padded out to be very wide.
Is it intended that the MRU saves/has a blank entry?

Screenshot_20231220_224625

Having used a long message, any subsequent message is padded to the same length as the long one!

Screenshot_20231220_225659

Usage expected (_most_ of which is working):
Type a new message, hit Enter and message appears in the Reporter.
Hit 'Clear' and it is wiped from the entry box, saved in the MRU list and cleared from the Reporter.
Activate the MRU drop down, click on a previous message, it is copied to the entry box and appears in the reporter.
Hit 'Clear' and the message is removed from the entry box and the reporter and not duplicated in the MRU list.

I think a context menu for the MRU list with the option to delete entries would be useful. Currently a right click on an entry does the same as left click.

tmiw:
> Why is there a 'Set' button? It seems to do nothing, can it be removed and 'Clear' be just [X]? What about 'Show', 'freq' and 'Msg', would they not then allow a single footer line?

The Set button actually does the same as pushing Enter in the drop-down list. It doesn't look like it does anything unless you modify the text, though. I don't necessarily think it's a good idea to remove it as I'm not sure the Enter behavior is 100% obvious to users, but perhaps it can be grayed out unless the text is in fact different?

> The 'Clear' button fails to clear the message entry box. 

This has been fixed.

> Anyone typing a long message will screw up everyone else's display - there must be a limit and I think ten chars is more than adequate, we only really wanted three to avoid all these issues. 

I've been testing with the message "M1 Mac Mini" to indicate which machine I'm on, which is more than 10 characters long. How about 15 for the limit?

> It needs to be a fixed width column

With a character limit, wouldn't the maximum end up being a certain width (with the minimum being the width of the "Msg" label + a bit of spacing)? Would it be okay to resize this column on every single change instead, or should there be no resizing at all (even if everyone is using shorter/empty messages)?

> Is it intended that the MRU saves/has a blank entry?

No. This has been fixed as well.

> I think a context menu for the MRU list with the option to delete entries would be useful. Currently a right click on an entry does the same as left click.

Based on the documentation I'm seeing, it doesn't look like this would be trivial to do (if possible at all). Would a separate "Remove" button work instead?

tmiw:
OK, I did some additional tweaks to the window to get it to fit on a 720p display:

Screenshot 2023-12-20 at 11 26 20 PM

(This includes a 15 character limit for the message.)

Hopefully this works out a bit better. The Msg column is dynamically changing size still but with the character limit, there should be a maximum width that it gets set to.

barjac:
> > Why is there a 'Set' button? It seems to do nothing, can it be removed and 'Clear' be just [X]? What about 'Show', 'freq' and 'Msg', would they not then allow a single footer line?
> 
> The Set button actually does the same as pushing Enter in the drop-down list.

Why would anyone use Enter in a drop down list? It's a mouse click action that everyone uses.

> It doesn't look like it does anything unless you modify the text, though. I don't necessarily think it's a good idea to remove it as I'm not sure the Enter behavior is 100% obvious to users 

Where did this 'use of Enter' come from? I agree its totally alien when using a mouse.
Re-read my last message part about usage near the end. The only mention of enter key is when typing a new text. I see no need for a Set button or use of Enter in connection with the MRU list.

> 
> > The 'Clear' button fails to clear the message entry box.
> 
> This has been fixed.
> 
> > Anyone typing a long message will screw up everyone else's display - there must be a limit and I think ten chars is more than adequate, we only really wanted three to avoid all these issues.
> 
> I've been testing with the message "M1 Mac Mini" to indicate which machine I'm on, which is more than 10 characters long. How about 15 for the limit?
> 
> > It needs to be a fixed width column
> 
> With a character limit, wouldn't the maximum end up being a certain width (with the minimum being the width of the "Msg" label + a bit of spacing)? Would it be okay to resize this column on every single change instead, or should there be no resizing at all (even if everyone is using shorter/empty messages)?

If it always defaults to the width of the 'Msg' column title, irrespective of what messages have gone before then 15 may be OK. Resizing on every change sounds OK, since in most cases the longest will still control the width so there may be no visible change.

> 
> > Is it intended that the MRU saves/has a blank entry?
> 
> No. This has been fixed as well.
> 
> > I think a context menu for the MRU list with the option to delete entries would be useful. Currently a right click on an entry does the same as left click.
> 
> Based on the documentation I'm seeing, it doesn't look like this would be trivial to do (if possible at all). Would a separate "Remove" button work instead?

We don't have room for that. Maybe if the MRU are stored in an accessible text file then users could add/remove their favourites at will.
I wonder if a right (or middle) click on the MRU list could be made to delete the item under the cursor?
Left click to use, right/middle click to delete? 

I have not had chance to test this yet, but it's looking good in your screen shot and getting better - nice!

barjac:
I just managed a quick test and it fits my 1280x1024 screen OK.

Screenshot_20231221_111806

However there is a bug when switching from the above All bands view to 30m where there are no messages:

Screenshot_20231221_112218

Tyrbiter:
Just checked this new build and I don't see anything wrong with the VK2 on 30m with band filtering selected, the message column shows up and the -- entries are correctly centred under the column headings.

barjac:
> Just checked this new build and I don't see anything wrong with the VK2 on 30m with band filtering selected, the message column shows up and the -- entries are correctly centred under the column headings.

No neither do I if I restart the reporter.
The problem happened when switching the filter to 30m from All with a message showing in the All view.
Restarting the reporter cleared the fault, so maybe some sort of re-set is required when the filter is changed?

tmiw:
I'm able to duplicate by selecting the filters in this order: All->160m->80m->60m. The fix I just checked in seems to resolve it for me, but give it a shot.

> Why would anyone use Enter in a drop down list? It's a mouse click action that everyone uses.

I should have clarified. Enter (or pushing the Set button) would be for adding new messages. No need to do either of those things if you select an existing item. 

> I see no need for a Set button or use of Enter in connection with the MRU list.

Agreed. Hopefully the graying out of it (and the Clear button) when there's nothing new to potentially commit to the list will make that clearer.

> Maybe if the MRU are stored in an accessible text file then users could add/remove their favourites at will. 

`~/.freedv` (or `~/Library/Preferences/FreeDV Preferences` on macOS), specifically this section:

```
[Reporting/FreeDV]
Enable=1
...
StatusText=M1 Mac Mini
RecentStatusTexts=M1 Mac Mini,012345678901234,012345678901236,M1 Mac Mini2,M1 2
```

May be a bit more complicated on Windows, though, since the Registry's used for that platform.

> I wonder if a right (or middle) click on the MRU list could be made to delete the item under the cursor? Left click to use, right/middle click to delete?

We'd only have access to the selected item, not the item under the mouse cursor (and selection doesn't change until the box closes again). Also, the proposed right-click behavior definitely wouldn't be obvious to users.

> Have you had any thoughts about notification of a change? I was thinking that the line that changed text should change background colour until the end of the next update cycle. (Unless it changes to empty).

What color should be used if this were to be implemented? Same as for RX, or a different one?

barjac:
> What color should be used if this were to be implemented? Same as for RX, or a different one?

User selectable like the others would be consistent.

barjac:
> I'm able to duplicate by selecting the filters in this order: All->160m->80m->60m. The fix I just checked in seems to resolve it for me, but give it a shot.

NO it's not fixed.
Starting the reporter with 20m filter enabled and me reporting on 60m I see this which is fine:
 
Screenshot_20231221_203852

If I then check 'Track' then this also OK:

Screenshot_20231221_204405

If I now un-check 'Track' then we hit a problem:

Screenshot_20231221_204746

tmiw:
> > What color should be used if this were to be implemented? Same as for RX, or a different one?
> 
> User selectable like the others would be consistent.

Sure, but what about the default?

> Edit - Just thinking more about this... If possible maybe changing only the colour of the Message itself rather than the whole line would draw better attention to it.

I'll have to look at how this is being done for the other colors. It's possible we can only do it for the entire row.

barjac:
> > > What color should be used if this were to be implemented? Same as for RX, or a different one?
> > 
> > 
> > User selectable like the others would be consistent.
> 
> Sure, but what about the default?

The palest purple looks good to me but I have forgotten what the defaults are.
I use the palest orange for TX and palest blue for RX.
The very pale colours provide the highest contrast with black text.

barjac:
See EDIT in my post with screen shots.

tmiw:
> Edit - Just thinking more about this... If possible maybe changing only the colour of the Message itself rather than the whole line would draw better attention to it.

Unfortunately it does indeed look like it has to be adjusted on a row by row basis.  The latest implements this with a 5 second timeout (TBD) and the purple background as the default.

> Moving from a filter _with_ a message to a filter _without_ causes the problem.

Yeah, I think this is a wxWidgets bug, but I think I was able to work around it now with the latest in this PR.
 
> O/T The message entry box will only display 14.5 characters without scrolling here ;)

The latest widens that field a bit more. I don't think I can go further without making the window unusable on 720p displays, though.

barjac:
Tested @ #aff64 and it's looking good, not much left to do :)
Three issues:
1. When changing from a wide message to a narrow one the Msg column header stays at max width.
2. Not sure I agree with 'Don't highlight when we change our own message.' I rather think it would be reassuring if all users see the same. I would like to see confirmation that my new message was highlighted for everyone.
3. The widened message entry box did not need such a big increase, less than half would have been adequate for my '0123456789abcde' test on the low aspect Dell office type screen.

tmiw:
> Three issues: When changing from a wide message to a narrow one the Msg column header stays at max width.

I tried to force that column to always auto-resize on any updates to it but that doesn't seem to be working for some reason. It's weird because I'm pretty sure other columns are able to auto-resize on changes without any issues.

Maybe we can leave it for now and revisit if it ends up becoming a serious issue.

> Not sure I agree with 'Don't highlight when we change our own message.'

Reverted.

> The widened message entry box did not need such a big increase, less than half would have been adequate for my '0123456789abcde' test on the low aspect Dell office type screen. See first image.

I shrunk it to 180px but let me know if it needs a tiny bit more.

barjac:
> > Three issues: When changing from a wide message to a narrow one the Msg column header stays at max width.
> 
> I tried to force that column to always auto-resize on any updates to it but that doesn't seem to be working for some reason. It's weird because I'm pretty sure other columns are able to auto-resize on changes without any issues.

No they don't.
'Version' is the same. If I filter to where users are all using 1.9.x release then switch to that band and start the modem, then the column widens to accomodate my 'FreeDV 1.9.6-0b234' but it fails to revert back when I stop the modem.

Also even if the filter is set to not include me I see a purple row (presumably me) flash up briefly on program start (full FreeDV not just reporter) and the Msg column is set wide for my message even though it's not on the screen.
> 
> Maybe we can leave it for now and revisit if it ends up becoming a serious issue.

OK but With a long Msg and a wide Version the low aspect screen is not far from the limit.

> > Not sure I agree with 'Don't highlight when we change our own message.'
> 
> Reverted.

Good that is better :)

> > The widened message entry box did not need such a big increase, less than half would have been adequate for my '0123456789abcde' test on the low aspect Dell office type screen. See first image.
> 
> I shrunk it to 180px but let me know if it needs a tiny bit more.

Hmm... this is my fault for not finding the widest characters and considering that it's not using a fixed width font.
Only 10 'M's fit the entry box now.
15 'M's cause the Last Update column to be truncated, however this is a corner case.

tmiw:
> No they don't. 'Version' is the same. If I filter to where users are all using 1.9.x release then switch to that band and start the modem, then the column widens to accomodate my 'FreeDV 1.9.6-0b234' but it fails to revert back when I stop the modem.

For me, it was changing column widths when I was switching band filters. Examples:

Screenshot 2023-12-23 at 10 03 07 PM

Screenshot 2023-12-23 at 10 03 15 PM

I did add some additional logic to force resize all columns when rows are removed from the list, though. This should apply both to the filter change case as well as simple disconnects/modem stops.

> Also even if the filter is set to not include me I see a purple row (presumably me) flash up briefly on program start (full FreeDV not just reporter) and the Msg column is set wide for my message even though it's not on the screen.

Should be fixed as of the latest, too.

> Hmm... this is my fault for not finding the widest characters and considering that it's not using a fixed width font. Only 10 'M's fit the entry box now. 15 'M's cause the Last Update column to be truncated, however this is a corner case and I think that messages that use mainly lower case will not cause a problem.

👍

barjac:
All seems fine now :)
One minor improvement may be to make the most recent MRU items appear at the bottom of the list since this list will generally display above the entry box which is at the bottom of the screen.
This will put the most commonly used items in easy reach.

tmiw:
> All seems fine now :) One minor improvement may be to make the most recent MRU items appear at the bottom of the list since this list will generally display above the entry box which (for most users) is at the bottom of the screen. This will put the most commonly used items in easy reach.

👍 

I'll leave the drop-down for now and revisit if it ends up that this is preferred instead.

PR #621, opened 2023-12-12
Allow display and entry of frequencies in KHz

### LINKED ISSUE #583: Add option for frequency entry in KHz instead of MHz
A request was submitted via the digitalvoice mailing list to provide an option to enter frequencies in KHz instead of MHz (i.e. "14236" instead of "14.236"). This issue was created so that this request isn't lost and can be evaluated further.

tmiw:
Resolves #583 by adding a new configuration option to allow display and entry of frequencies in KHz.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -368,10 +368,11 @@ along with the reporting feature as a whole:
 * [FreeDV Reporter](https://qso.freedv.org/) -- also accessible via the Tools->FreeDV Reporter menu option.
 
 The frequency that FreeDV reports is set by changing the "Report Frequency" drop down box in the main window. This 
-is in kilohertz (kHz) and will turn red if the entered value is invalid. If Hamlib support is also enabled, 
-this frequency will automatically remain in sync with the current VFO on the radio (i.e. if the frequency is changed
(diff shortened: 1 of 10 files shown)


Tyrbiter:
This seems to work OK, but should the entry list in the Options dialog also change to kHz when this option is selected?

tmiw:
> This seems to work OK, but should the entry list in the Options dialog also change to kHz when this option is selected?

It should already be doing that (I tested last night to be sure), albeit you need to push Apply or OK first.

Tyrbiter:
OK, yes, seems to work. If a value already exists in the entry box, say 1.997MHz,  then this turns into 1.997kHz after changing the setting and applying it. Might be a bit confusing.

tmiw:
> OK, yes, seems to work. If a value already exists in the entry box, say 1.997MHz, then this turns into 1.997kHz after changing the setting and applying it. Might be a bit confusing.

Should be fixed now :)

Tyrbiter:
> > OK, yes, seems to work. If a value already exists in the entry box, say 1.997MHz, then this turns into 1.997kHz after changing the setting and applying it. Might be a bit confusing.
> 
> Should be fixed now :)

And indeed it is, thanks.

PR #632, opened 2023-12-29
Status message improvements

tmiw:
This PR adds several improvements to the recently-added status message feature:

1. Allows user to refresh the status update (and cause it to highlight to all users) by pushing the Set button. (1.9.6 previously disabled this button unless the message was actually changing.)
2. Increases the priority of the message highlight color so that it takes precedence over all other highlight types (i.e. updating the message while transmitting or receiving will cause a purple highlight before returning to RX/TX highlighting).


barjac:
Hi,
Tested in x86_64 and aarch64 @ #a7198 and it's looking good :)

I can't test purple replacing blue but it replaces red just fine.

One issue. The Msg column is again not resizing. This time the Version column is not affected, only Msg. I'm now wondering if this is a feature not a bug? It is always at max. width.
Not sure I like that, as it occupies a lot of useless space when not in use now.

tmiw:
> One issue. The Msg column is again not resizing. This time the Version column is not affected, only Msg. I'm now wondering if this is a feature not a bug?
> 
> Edit: Ah, it is a bug, as the Msg column does return to narrow after a full FreeDV restart. Re-starting only the reporter retains the wide column.
> 
> To reproduce, Start FreeDV Add a short 'QRL' message. Change message to '0123456789abcde' Change message back to 'QRL'

It looks like wxWidgets purposely doesn't shrink columns unless everything is deleted first. I checked in a change that will allow you to at least reset the column widths if you change the band filter. I don't think we can do much more unless wxWidgets itself changes.

barjac:
> It looks like wxWidgets purposely doesn't shrink columns unless everything is deleted first. I checked in a change that will allow you to at least reset the column widths if you change the band filter. I don't think we can do much more unless wxWidgets itself changes.

IIRC This was working fine until the last changes to allow the colour highlight to change from red or blue to purple.

tmiw:
> > It looks like wxWidgets purposely doesn't shrink columns unless everything is deleted first. I checked in a change that will allow you to at least reset the column widths if you change the band filter. I don't think we can do much more unless wxWidgets itself changes.
> 
> IIRC This was working fine until the last changes to allow the colour highlight to change from red or blue to purple.

I'm not sure how considering that the only sizing related change was the very last commit to fix it for the band filter change case. Can you confirm by running `master` / 1.9.6 just to be sure?

barjac:
It's a subtle difference. I was testing previously by clicking 'Track' to move to a filter with some wide messages and my own which was narrow. In this case it still works fine the Msg column changes width correctly. If I do the same as above to show the current issue and click Track or Band or Freq while the message column is wide and the message 'QRL' is narrow the column width does in fact reduce to fit.

tmiw:
> So I'm guessing that whatever happens on a filter change is needed on a message change too?

Unfortunately with the current wxWidgets library, that means deleting all the rows from the list, then re-adding them individually. Here's how `SetColumnWidth` is defined:

```c++
    if ( width == wxLIST_AUTOSIZE_USEHEADER || width == wxLIST_AUTOSIZE )
    {
        wxListCtrlMaxWidthCalculator calculator(this, col);

        calculator.UpdateWithWidth(AUTOSIZE_COL_MARGIN);

        if ( width == wxLIST_AUTOSIZE_USEHEADER )
            calculator.UpdateWithWidth(ComputeMinHeaderWidth(column));

        //  if the cached column width isn't valid then recalculate it
        wxColWidthInfo* const pWidthInfo = m_aColWidths.Item(col);
        if ( pWidthInfo->bNeedsUpdate )
        {
```

`bNeedsUpdate` seems to get set to `true` only if the new width is longer than the existing one (i.e. when modifying or adding rows), using similar logic to the below:

```c++
        // calculate the width of the item and adjust the max column width
        wxColWidthInfo *pWidthInfo = m_aColWidths.Item(col);
        int width = GetItemWidthWithImage(&item);
        item.SetWidth(width);
        if (width > pWidthInfo->nMaxWidth)
        {
            pWidthInfo->nMaxWidth = width;
            pWidthInfo->bNeedsUpdate = true;
        }
```

Anyway, I don't think doing a full erase and re-add (especially as common as it would likely end up being given that people can re-trigger reporting of existing messages) is going to look all that great for users, either.

barjac:
OK, I agree that it does cause a full re-write of the table which is visible to the user, whether any would actually notice is debatable ;)
I can always just click whichever is already in use of the Band or Freq buttons to clean up the table manually as a workaround.

barjac:
Ah! I have just noticed a related issue doing some more testing in a QSO with another station.

Using any of the Band filters and 'Track' with 'Band' selected all is OK, but with 'Freq' selected the whole table is refreshing (visibly the colours and text momentarily flashes) on every 'Last Update' change by any station, transmitting or receiving.

I noticed it in this branch, but I am building master now to double check in Mageia 9 stable, rather than Cauldron (dev).

PR #637, opened 2023-12-31
Fix flicker in FreeDV Reporter window when tracking by frequency.

### LINKED ISSUE #635: 1.9.6 Reporter table flicker when using 'Track' with 'freq'
When using 'Track' with 'freq' set, the whole reporter table is refreshing on each 'Last update' update, maybe only the local user's update but not certain as some updates can display almost simultaneously.

The flicker effect happens about 150ms before the time changes in the user's 'Last update'. 

This is causing an annoying flicker, most noticeable when colour is showing, as it causes the colours to be extinguished for several seconds each time.

Tested also in master at #a8f58

/OT
A secondary effect of this relates to issue #634, as the refresh triggers the truncation of the callsign column in Win10.

tmiw:
Resolves #635 by preventing clearing of the reporter list if the current frequency hasn't changed.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -921,6 +921,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix intermittent crash resulting from object thread starting before object is fully initialized. (PR #630)
     * Prevent creation of filters if not enabled. (PR #631)
     * Fix issue preventing Start button from reenabling itself on audio device errors. (PR #636)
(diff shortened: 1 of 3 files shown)


barjac:
Looks good to me! :)

Maybe a 1.9.6.1 would be in order so Windows10 users get the 'Callsign' fix, if it is working?

Happy New Year!

barjac:
No problems noticed in a net today with four other stations, the purple highlight worked well to draw attention.

PR #641, opened 2024-01-03
Update Filter dialog to better handle resizing.

### LINKED ISSUE #639: Minor things in current git
Just a quick report of some very minor things.

In USER_MANUAL.html there is one instance of 'applicaton' when it should read 'application'.

In the filter dialog, for some reason the Default button moves slightly when selecting the Speaker Output equaliser tab for the first time. This is with neither equaliser enabled. Repeatable if window is closed and reopened, but not when switching tabs again.

tmiw:
Resolves #639 by updating various controls in the Filter dialog to better handle when the dialog is resized.


Tyrbiter:
Nailed it @tmiw 

Don't know how you know wxWidgets so well :)

PR #642, opened 2024-01-04
Fix capitalization of distance units in FreeDV Reporter window.

tmiw:
Followup request from https://github.com/drowe67/freedv-gui/issues/639#issuecomment-1875529675 to update distance units to use correct capitalization.


Tyrbiter:
Perfect.

PR #643, opened 2024-01-05
Rename KHz -> kHz.

tmiw:
Turns out that kilo does *not* use capital 'K'. This PR renames references in the UI and documentation accordingly.


Tyrbiter:
This is a good tweak, looks right now.

PR #644, opened 2024-01-05
Left justify data and labels, remove '--' and align data start with label start for easier readability and cleaner looks

barjac:
This PR attempts to improve the readability and looks of the internal reporter by left aligning all the data in each column rather than centring it within the current column widths which looks ragged. 

It makes sure that the start of each data item is below the start of the column label.

It removes the '--' items which indicate that there is no data. However we can see that there is no data if there is nothing there.
I can maybe understand using this in a purely testing environment but not for production as it is serving no purpose.

I will add a screen shot later.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -906,6 +906,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 2. Enhancements:
     * Allow user to refresh status message even if it hasn't been changed. (PR #632)
     * Increase priority of status message highlight. (PR #632)
+    * Adjust FreeDV Reporter data display to better match accepted UX standards. (PR #644)
 3. Build system:
     * Include PDB debugging file for FreeDV. (PR #633)
 4. Documentation:
--- a/src/freedv_reporter.cpp
+++ b/src/freedv_reporter.cpp
@@ -27,7 +27,7 @@
 
 extern FreeDVInterface freedvInterface;
 
-#define UNKNOWN_STR "--"
+#define UNKNOWN_STR ""
 #if defined(WIN32)
 #define NUM_COLS (14) /* Note: need empty column 0 to work around callsign truncation issue. */
 #else
--- a/src/freedv_reporter.cpp
+++ b/src/freedv_reporter.cpp
@@ -301,11 +301,11 @@ void FreeDVReporterDialog::refreshLayout()
 
     if (wxGetApp().appConfiguration.reportingConfiguration.useMetricDistances)
     {
-        item.SetText("km");
+        item.SetText("km ");
     }
     else
     {
-        item.SetText("mi");
+        item.SetText("Miles");
     }
 
     m_listSpots->SetColumn(2 + colOffset, item);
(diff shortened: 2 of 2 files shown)


barjac:
Screenshot_20240105_095612

barjac:
> Thanks for creating this PR! Were you able to do any experimentation to see if it were possible to keep the column titles center aligned while left aligning everything else? I remember trying before and being unable to.

I read some wx docs and could see nothing, but I really don't think that would be better.
Having the label accurately aligned with the data does not look bad at all. If the data includes narrow items as in the Msg column alongside the odd wide one, then a centred label could be very misaligned with the narrow item.
I used it in our daily 60m net today and took a screen shot. (Will add from a other machine shortly)

Tyrbiter:
I'm not against this, however I think it may not be ideal for the distance and SNR columns, I think that they look better if right-justified.

barjac:
> I'm not against this, however I think it may not be ideal for the distance and SNR columns, I think that they look better if right-justified. Perhaps for distance the own station value should be 0 or blank instead of 0.0 as well.

Yes I agree, seems odd that 0 is displayed as 0.0.

tmiw:
I was doing a bit more reading on UX standards for tables and they seem to have the following in common:

* Text: left aligned
* Numbers: right aligned
* Column header alignment should match data

Example guidance: https://medium.com/mission-log/design-better-data-tables-430a30a00d8c

With that in mind, I pushed some changes to right-align "km/mi", "kHz/MHz", and "SNR" columns. It actually significantly improved it for me to the point where I think it might be counterproductive to center align the column headings (assuming that's actually possible without significant changes to wxWidgets itself). Hopefully that's okay for you guys too, but if not, we can revert and just have everything left aligned.

barjac:
I did a similar thing today before reading your last above. I also removed the check for distance < 10.0 which solved the 0.0 issue for me. I'm getting a bit lost in github - I don't see your changes :(

Screenshot_20240106_124107

Tyrbiter:
I've just made the changes as suggested, with dates centre-justified but then other text left-justified and numbers right-justified.

Also removed the <10.0 distances units check and that's nicer IMHO.

Not completely sure about the centred dates, I think that it's probably quite a weak change and is probably font-dependent.

tmiw:
Yeah, considering the width of the text in the date columns, I don't think there's much difference other than for the headers. I'd say to back out that particular change to be more consistent with the other text-based columns.

barjac:
OK the dates are back to left justified and I am very happy with it now. Note I also changed mi to Miles which looks much better and in most cases causes no width overhead.

Screenshot_20240106_201857

tmiw:
> Note I also changed mi to Miles which looks much better and in most cases causes no width overhead.

Seems like we'd want to keep it consistent with both units, right? As in, either abbreviate both or use the full name for both.

Though I'm also fine just using "Miles" if that is indeed better.

barjac:
> > Note I also changed mi to Miles which looks much better and in most cases causes no width overhead.
> 
> Seems like we'd want to keep it consistent with both units, right? As in, either abbreviate both or use the full name for both.

No not really! 'Kilometers' is too long, 'Miles' is not and if you only use one of them then you will not care! ;)
> 
> Though I'm also fine just using "Miles" if that is indeed better.

Looks much better to me, 'mi' looks like an SI unit that I never heard of and it's not a common abbreviation for mile. :)

barjac:
I was trying to figure out what was causing the 'Last update' to be pushed off my screen. It appears to be mainly down to someone using an eight character locator. Can this field be limited to 6 characters? ...and lower case conversion for the last two?
Also SNR column seems over wide as it only ever needs 4 characters.
Miles/km is also wider than needed as it only needs 5 digits, unless it's catering for interplanetary contacts :)
By manually adjusting these the whole table fits again.

Screenshot_20240106_224900

tmiw:
I'll go ahead and merge, then work on the stuff mentioned in the very last comment above.

PR #646, opened 2024-01-08
More column sizing adjustments

tmiw:
This PR performs the following to address feedback posted at https://github.com/drowe67/freedv-gui/pull/644#issuecomment-1879870187:

1. Limits displayed grid squares to six characters in length and actually ignores invalid input (vs. trying to calculate a distance that's probably wrong).
2. Instead of using wxWidgets' column autosizing, implement our own, ensuring that column sizes don't drop below hardcoded minimums.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -903,10 +903,12 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Update Filter dialog to better handle resizing. (PR #641)
     * Fix capitalization of distance units in FreeDV Reporter window. (PR #642)
     * Rename KHz to kHz in documentation and UI. (PR #643)
+    * Avoid calculating distances in FreeDV Reporter window for those with invalid grid squares. (PR #646)
 2. Enhancements:
     * Allow user to refresh status message even if it hasn't been changed. (PR #632)
     * Increase priority of status message highlight. (PR #632)
     * Adjust FreeDV Reporter data display to better match accepted UX standards. (PR #644)
+    * Further reduce required space for each column in FreeDV Reporter window. (PR #646)
 3. Build system:
     * Include PDB debugging file for FreeDV. (PR #633)
 4. Documentation:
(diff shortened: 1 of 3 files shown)


tmiw:
@barjac, hopefully this works for you?

Tyrbiter:
I built and installed this latest version. There are some issues.

The 2 Mode columns are different widths for some reason. The SNR column is a bit too wide. The Locator column is now wider than the new maximum 6 characters displayed. The distance column could be narrowed a little, 5 digits should be sufficient.

I appreciate how difficult this is, it may also be affected by font sizes.

barjac:
Yes, here is a direct comparison of before this change and now:
Screenshot_20240106_224900c
Screenshot_20240108_142420

tmiw:
I fixed the Mode sizing discrepancy and added a change to the sizing calculation that might help. Unfortunately some space does need to be left for the sort indicator, so there might be a limit to how much further the other columns can shrink. For example, when sorting by SNR:

Screenshot 2024-01-08 at 9 16 31 AM

Tyrbiter:
This seems much better, and apologies for forgetting about the space used by the column sorting indicator.

I don't think much improvement is possible now.

barjac:
I agree much improved :)
I have added another screen shot to the group above.

tmiw:
> While on the looks of the GUI, could a 'dark mode' be made available from within the program? I can use the system theme in plasma5 on x86_64 but changing the theme in LXQT DE on the RPi4 does not affect FreeDV. Maybe I just can't find the correct setting.

If you're using wxWidgets 3.2, it should automatically detect the OS/dev environment's setting for that. Seems to on Ubuntu with GNOME, anyway, albeit not 100% perfect. For example, when starting in dark mode:

Screenshot 2024-01-08 at 11 07 43 PM

and then switching to light mode in the GNOME settings:

Screenshot 2024-01-08 at 11 07 54 PM

tmiw:
BTW, I just checked in a change that should make dark->light mode transitions work a bit better in the FreeDV Reporter window (at least on Linux and maybe macOS; Windows does have some separate logic for that control in wxWidgets so it might not have had the problem in the first place).

PR #647, opened 2024-01-10
FreeDV Reporter message behavior changes

### LINKED ISSUE #645: Minor enhancement to reporter message entry - discussion
At present any one off 'throw away' comments that are sent are added to the 'Saved' list, filling it up with 'never again to be needed' items. Also 10 items is not currently enough, maybe 20 would be?

One possible simple solution:
Use CTRL/'Set' to add to the list and use just 'Set' to send without saving.

My initial idea for the list was that it would contain the 'most used' items in popularity order and now realize that the 'most recently used' was the wrong terminology.

Maybe:
If the last 200 or so messages were logged and sorted into most used order then the top 10+ could appear in the list. When the log reaches max size then delete everything that only appears once.
(there is a bug in that idea but it's food for thought :)

tmiw:
Resolves #645 by changing FreeDV Reporter message behavior as follows:

1. Left-click of the Set button no longer saves the message on the server.
2. Right-click of the Set button brings up a new option to save the message (along with sending it to the server).

Additionally, tooltips were added to each button in the FreeDV Reporter window to ensure that users can find the expected behavior of each.


barjac:
> Resolves #645 by changing FreeDV Reporter message behavior as follows:
> 
>     1. Left-click of the Set button no longer saves the message on the server.

I assume that is a typo?
Did you mean that left-click sends the message to the server but does not save it to the local list?

> 
>     2. Right-click of the Set button brings up a new option to save the message (along with sending it to the server).

I don't see a need for a context menu, can right-click not simply send the message to the server and save it to the list?
> 
> Additionally, tooltips were added to each button in the FreeDV Reporter window to ensure that users can find the expected behavior of each.

That sounds good - I have not tested any of this yet though.

I think a GUI means of removing messages from the list is needed.

Tyrbiter:
Just built and installed this, the only problem is that the Set and Save button is appearing at the top left of the reporter window instead of next to the Set button. Other than that, it seems to do what it says.

tmiw:
> I don't really see a need for a context menu, can right-click not simply send the message to the server and save it to the list?

The context menu is mainly to help people see what will happen if you right-click. Otherwise, the only way someone would know that right-click == save would be to read the tooltip (which I'm not sure is visible enough). It also gives us flexibility to add other features in the future if we wanted.

> I think a GUI means of removing messages from the list is needed. Edit: Maybe a pick list from the 'Clear' button (same as the Message list) to remove an item from the list? Possibly with confirmation?

Maybe another right-click menu for the Clear button that has "Remove selected from list" and "Remove all from list" options?

> Just built and installed this, the only problem is that the Set and Save button is appearing at the top left of the reporter window instead of next to the Set button. Other than that, it seems to do what it says.

This should be fixed now.

Tyrbiter:
> 
> Maybe another right-click menu for the Clear button that has "Remove selected from list" and "Remove all from list" options?

Yes, that would do it for me as long as the clear all entry is not the default selection.

> 
> > Just built and installed this, the only problem is that the Set and Save button is appearing at the top left of the reporter window instead of next to the Set button. Other than that, it seems to do what it says.
> 
> This should be fixed now.

Thanks.

tmiw:
> Yes, that would do it for me as long as the clear all entry is not the default selection.

Added a right-click context menu for the Clear button as well with "clear selected" and "clear all" items.

barjac:
I don't like this, sorry. :(
I was expecting the right click on clear to display the list of messages currently saved to allow one to be selected and deleted.

Consider this, I am in a QSO and decide to remove an unimportant message to replace it with another I want to use.

With the implementation as it is now I would have to send the message (that I am about to delete) to the server and make it current and world visible before I could delete it, unless I am mistaken?
For this to work it would need to have the click to send removed so that a message can be selected and deleted without automatically sending it.

This currently only works if the modem is stopped, which is never the case while operating FreeDV.

tmiw:
> For this to work it would need to have the 'click to select and send' removed so that a message could be selected and deleted without automatically sending it. This would then require an extra click on 'Set' when using the list.

If that's okay with you, I can remove that behavior and require users to explicitly click Set to send the message.

barjac:
> > For this to work it would need to have the 'click to select and send' removed so that a message could be selected and deleted without automatically sending it. This would then require an extra click on 'Set' when using the list.
> 
> If that's okay with you, I can remove that behavior and require users to explicitly click Set to send the message.

I don't think it's ideal but it is acceptable and I appreciate it's a lot less work :)
Whether it will be confusing for users is debatable.
As I said above "I was expecting the right click on 'Clear' to display the list of messages currently saved to allow one to be selected and deleted totally independently of any message entry in use." Similar to the chooser on the Message box but if it's a pain to do then I suppose it's only one extra click to hit 'Set' each time.

tmiw:
> I don't think it's ideal but it is acceptable and I appreciate it's a lot less work :)
> 
> That would have the advantage that the message selected would be transferred to the 'Message' box and could be edited before sending :)

Done.

> BTW I don't like 'Set' as it's not very meaningful. I think 'Send' would be more acceptable.

Also changed :)

barjac:
Looks fine now to me - everything works as expected - Well done!
I don't understand why Americans appear to think that the Msg field is for QTH though. Very strange. :)

Tyrbiter:
> I don't understand why Americans appear to think that the Msg field is for QTH though. Very strange. :)

Maybe adding a bearing column would avoid this ;-) The US is a big place so beam headings can be almost any direction even on HF.

PR #649, opened 2024-01-12
Fix issue preventing usage of the 5th and 6th letter of the grid square.

tmiw:
Followup to PR #646 that fixes a bug introduced in the grid square math calculation. After this fix, distances appear to calculate correctly using the [ARRL grid square distance calculator](https://contest-clubs.arrl.org/griddistancecalc.php):

Screenshot 2024-01-12 at 12 46 59 AM


barjac:
Yes looks fine now :)

PR #650, opened 2024-01-13
Add guard code to prevent FreeDV Reporter window from being off screen on startup.

tmiw:
While helping a user get their FreeDV Reporter window back on screen, I noticed that the left and top properties for the window in the Registry were set to values close to `UINT32_MAX`. This PR adds additional guard logic to prevent FreeDV from loading invalid window X and Y positions on startup, thus ensuring that the window can be recovered without wiping the FreeDV configuration.


barjac:
I wondered about the jiggle workaround that you introduced some time ago for the main window positioning issue. If it was already at max h then using h + 1 on start appeared to cause an accumulating decrease in window height on one particular window manager between re-starts. I patched this locally to not touch w (which was not needed) and to only use h - 1.
Not sure if this relates to this issue, but it reminded me to mention it.

tmiw:
> I wondered about the jiggle workaround that you introduced some time ago for the main window positioning issue. If it was already at max h then using h + 1 on start appeared to cause an accumulating decrease in window height on one particular window manager between re-starts. I patched this locally to not touch w (which was not needed) and to only use h - 1. Not sure if this relates to this issue, but it reminded me to mention it.

It doesn't look like that happens in the FreeDV Reporter window but should be straightforward to do in another PR.

PR #656, opened 2024-01-16
Fix issue causing intermittent crashes when filters are enabled while running.

tmiw:
Reported on the digitalvoice mailing list. This is due to changes in 1.9.7 with how the filters are re-created upon modification. The workaround in this PR is to ensure that the filters are actually deleted before recreating them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ endif(APPLE)
 
 # Adds a tag to the end of the version string. Leave empty
 # for official release builds.
-set(FREEDV_VERSION_TAG "devel")
+set(FREEDV_VERSION_TAG "")
 
 # Prevent in-source builds to protect automake/autoconf config.
 # If an in-source build is attempted, you will still need to clean up a few
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ endif()
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11" CACHE STRING "Minimum OS X deployment version")
 
 set(PROJECT_NAME FreeDV)
-set(PROJECT_VERSION 1.9.7)
+set(PROJECT_VERSION 1.9.7.1)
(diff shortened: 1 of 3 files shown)


Tyrbiter:
I have installed an updated build, it appears to function normally although I did not suffer from the crashes that prompted the patch.

tmiw:
> I have installed an updated build, it appears to function normally although I did not suffer from the crashes that prompted the patch.

Checking the Enable checkbox in the Filter dialog for the Speaker Out filters while FreeDV is running is the quickest way to duplicate it (though you may need to do it a few times). Looks like it happens on multiple platforms, not just Windows.

Tyrbiter:
> Checking the Enable checkbox in the Filter dialog for the Speaker Out filters while FreeDV is running is the quickest way to duplicate it (though you may need to do it a few times). Looks like it happens on multiple platforms, not just Windows.

OK, have just spent a good few minutes repeatedly enabling and disabling the Speaker Out filter with the patched version and absolutely nothing bad has happened. As long as the Mic Input filter doesn't misbehave then it looks good here.

tmiw:
> > Checking the Enable checkbox in the Filter dialog for the Speaker Out filters while FreeDV is running is the quickest way to duplicate it (though you may need to do it a few times). Looks like it happens on multiple platforms, not just Windows.
> 
> OK, have just spent a good few minutes repeatedly enabling and disabling the Speaker Out filter with the patched version and absolutely nothing bad has happened. As long as the Mic Input filter doesn't misbehave then it looks good here.

Would you be able to try that test again on 1.9.7 and make sure it crashes for you? That way we can be fully sure this PR solves that problem.

Tyrbiter:
Yes, I will try this out this evening and comment a bit later.

tmiw:
Weirdly, I can't duplicate it now on multiple machines. Based on where the assertion was happening, I'm still fairly confident this will fix it, so we'll run with it. Worst case, we have to do a .2.

PR #670, opened 2024-01-25
Move TX Mode column to left of Status in FreeDV Reporter window.

tmiw:
Example display with this change:

Screenshot 2024-01-24 at 9 36 24 PM


tmiw:
@barjac, does this look reasonable?

barjac:
Yes, I think it makes sense to have the TX mode and frequency together on the left away from the RX info on the right so there is no chance of confusion between the the two 'Mode' columns.
That looks good to me :)

PR #672, opened 2024-01-26
Add heading column to FreeDV Reporter window.

### LINKED ISSUE #671: Add a beam heading column to the (internal) reporter.
This was suggested by Mooneer during another bug/PR a few weeks ago.

It would be a good addition to the reporter and would only require a 3 digit wide column next to the Locator column to show the great circle beam heading from the current station.

Adding this enhancement request so as to not forget about it :)

tmiw:
Resolves #671 by adding a heading for each user in the FreeDV Reporter window. For example:

Screenshot 2024-01-25 at 8 05 46 PM

Values have been cross-checked with https://www.chris.org/cgi-bin/finddis and appear to be correct.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -902,6 +902,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Update tooltip for the free form text field to indicate that it's not covered by FEC. (PR #665)
     * Enable use of space bar for PTT when in the FreeDV Reporter window. (PR #666)
     * Move TX Mode column to left of Status in FreeDV Reporter window. (PR #670)
+    * Add heading column to FreeDV Reporter window. (PR #672)
 3. Code cleanup:
     * Move FreeDV Reporter dialog code to dialogs section of codebase. (PR #664)
 
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -1092,7 +1092,8 @@ double FreeDVReporterDialog::DegreesToRadians_(double degrees)
 
 double FreeDVReporterDialog::RadiansToDegrees_(double radians)
 {
-    return (radians > 0 ? radians : (2*M_PI + radians)) * 360 / (2*M_PI);
+    auto result = (radians > 0 ? radians : (2*M_PI + radians)) * 360 / (2*M_PI);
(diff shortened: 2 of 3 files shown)


barjac:
Wow! that was quick :)
Why is the column so wide? You seem to have lost part of the 'Last update' as a result.
Also why is your distance sort not working? You seem to be sorting on distance but it's not sorted.
I will test here later on my low aspect system.

Tyrbiter:
> Wow! that was quick :) 
@tmiw magic at work.

> Why is the column so wide?
Might need a tweak, but it's not much too wide.

>You seem to have lost part of the 'Last update' as a result. 
There's a horizontal scroll bar, so it's just that the window hasn't been resized.

>Also why is your distance sort not working? You seem to be sorting on distance but it's not sorted.
Looks OK to me, sort appears to be working on both distance and heading for me, the image above shows increasing distance in km.

I think it's fine although the column width could be reduced a little.

barjac:
Yes it's not too wide but does look the same width as the distance column which requires 5 digits.

My mistake - must have still been half asleep when I looked at that image the sort is fine /o\.
On my Low aspect screen there is plenty of room:

Screenshot_20240126_140919

Interesting that it shows the heading to me as 360deg :) Maybe that should be just blank?

Tyrbiter:
> Interesting that it shows the heading to me as 360deg :) Maybe that should be just blank?

You should be able to hear yourself at any one of those 360 degrees :-p

tmiw:
> Interesting that it shows the heading to me as 360deg :) Maybe that should be just blank?

Yeah, I just made it blank. I also added logic to change 360 degrees to 0 degrees since that probably makes more sense to people. (Though not everyone does that; for example, airport runways actually go from 1 to 36, meaning that a completely north/south runway is labeled "18/36".)

tmiw:
Also, for me, it doesn't look like I can make the Hdg column any narrower without truncating the column name itself when sorting on it:

Screenshot 2024-01-26 at 10 40 55 PM

I'll merge for now and we can always revisit formatting if needed.

barjac:
OOps!
I only just noticed that the version column is labelled 'MHz'

tmiw:
> OOps! I only just noticed that the version column is labelled 'MHz'

Fixed: https://github.com/drowe67/freedv-gui/pull/675

Issue #677, opened 2024-01-28
Further GUI development - discussion


barjac:
Now that the GUI of both the main program and the reporter are compact and looking neat and clean thanks to great work by Mooneer I have been wondering how difficult it would be to combine the reporter into the main window.

Most users I know arrange the main window to occupy the top half of the screen and fit the reporter below it which involves some window dragging and sizing to achieve.

My ideal arrangement would have the reporter below the existing main window content by default, with an adjustable dividing line between them which could be dragged up and down to give either the main program or the reporter more height as required/desired.

If the reporter were turned off then the main window would get the whole screen.

Any thoughts anyone?

tmiw:
That's actually a significant amount of work for I'm not sure how much gain. Additional feedback from others would be good (and submitting this via the existing feature request template would be good if this is something others want as well).

Tyrbiter:
Currently the reporter windows stays on top of the main window when open, if it could be dropped down the Z-order I would find that more useful.

tmiw:
> Currently the reporter window stays on top of the main window when open, if it could be dropped down the Z-order I would find that more useful.

Played around with this a little bit and haven't found a way to change this behavior yet. I also noticed that the Filter dialog is the same way. Not fully sure why.

BTW I haven't noticed this on any other platform except for Linux, either. /shrug

tmiw:
How about https://github.com/drowe67/freedv-gui/pull/679? That should at least solve the Z ordering issue.

tmiw:
Just merged #679, any other action needed here?

barjac:
> Just merged #679, any other action needed here?

That was a side issue, it does not affect the main suggestion which needs discussion between users to decide on whether or not it needs escalating to a 'new feature request'.

I would suggest leaving this open to allow time for users to consider and become aware of this idea.

Tyrbiter:
I would say that not having the reporter window forced on top allows for different ways of using the whole application. If we are suggesting a different top window configuration then that should be against the background of what freedv-1.9.8 will offer which already has additional flexibility over the current version.

barjac:
> I would say that not having the reporter window forced on top allows for different ways of using the whole application. If we are suggesting a different top window configuration then that should be against the background of what freedv-1.9.8 will offer which already has additional flexibility over the current version.

Yes, I agree that the new configuration that allows the main GUI and the reporter to be minimized to separate task bar items does make a big difference and is much more flexible than the current one.

[...]

PR #679, opened 2024-01-30
Prevent FreeDV Reporter window from being above the main window.

tmiw:
Per comment at https://github.com/drowe67/freedv-gui/issues/677#issuecomment-1913747293, this PR updates the FreeDV Reporter window's initialization so that it's not always kept above the main window.


Tyrbiter:
Seems to work as expected on GNOME 45.x Wayland here, so it gets a +1 from me.

PR #685, opened 2024-02-03
Add support for displaying cardinal directions instead of headings.

### LINKED ISSUE #682: Option to display cardinal points rather than degrees in the reporter heading  column.
Is there a possibility of extending the usability of the reporter heading column to display 16 approximate cardinal points as an alternative to degrees?

I have found a C++ routine that should handle the conversion:

```
int cardinalDirectionNumber( static_cast<int>( ( ( currentGustDirection / 360.0 ) * 16 ) + 0.5 )  % 16 );
const char* const cardinalDirectionTexts[] = { "N", "NNE", "NE", "ENE", "E", "ESE", "SE", "SSE", "S", "SSW", "SW", "WSW", "SW", "WNW", "NW", "NNW" };
string dirString = cardinalDirectionTexts[ cardinalDirectionNumber ];
```

Reference:
https://cumulus.hosiene.co.uk/viewtopic.php?p=9802&sid=bdb6d3c1f7f482e87ca10856dec6c1ff#p9802

tmiw:
Resolves #682 by adding a new setting in Tools->Options to show directions (e.g. N/S/E/W) instead of degrees for the Heading/Direction column.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -906,6 +906,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Move TX Mode column to left of Status in FreeDV Reporter window. (PR #670)
     * Add heading column to FreeDV Reporter window. (PR #672, #675)
     * Prevent FreeDV Reporter window from being above the main window. (PR #679)
+    * Add support for displaying cardinal directions instead of headings. (PR #685)
 3. Code cleanup:
     * Move FreeDV Reporter dialog code to dialogs section of codebase. (PR #664)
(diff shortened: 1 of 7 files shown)


barjac:
Great!
Just one thing, since the headings in cardinal mode are text, should they not be left justified to use the same convention as other columns?

tmiw:
> Great! Just one thing, since the headings in cardinal mode are text, should they not be left justified to use the same convention as other columns?

Updated.

barjac:
Perfect!

PR #688, opened 2024-02-11
Fix typo in cardinal directions list.

tmiw:
Reported on IRC #freedv. Turns out "W" was incorectly showing up as "SW"; this PR resolves that issue.

EDIT: typo


barjac:
> Reported on IRC #freedv. Turns out "W" was incorectly showing up as "NW"; this PR resolves that issue.



> Reported on IRC #freedv. Turns out "W" was incorectly showing up as "NW"; this PR resolves that issue.

It was actually showing as "SW" but no matter, the fix is correct :)

Issue #703, opened 2024-03-28
When tracking 'band', a change to 'frequ' is ignored until tracking is stopped and restarted. 


barjac:
I was on 28.720 two others were on 28.330 when I had band selected. Clicking freq made no difference. I had to click 'track' twice to see only 28.720.

![Screenshot_20240328_131146](https://github.com/drowe67/freedv-gui/assets/1324702/7070a8c2-6117-4224-96f1-b28c94b7f322)

tmiw:
Thanks for finding this! Let me know how the fix in https://github.com/drowe67/freedv-gui/pull/704 goes. It seems that the user list *does* update in my testing, just not immediately (only when we receive an update for the users not matching the filter from the server, which typically happens within ~5 seconds with the more recent versions of FreeDV).

barjac:
Yes, tested and seems just fine now - thanks! :)

tmiw:
> Yes, tested and seems just fine now - thanks! :)

Sounds good. I just merged the appropriate PR. 👍

PR #704, opened 2024-03-30
Fix issue causing delayed filter updates when going from tracking band to frequency.

tmiw:
This PR resolves an issue discovered by @barjac in #703 where updates to the displayed list don't occur immediately when switching from tracking band to frequency, but do update the next time the user receives an update from the server for the users not matching the current filter.


tmiw:
> Using current master at #10ea1 in QSO I am seeing a very annoying flicker every time my reporter updates, which affects all reporting stations in the table which is set to track freq. The whole table flashes white briefly every 5 seconds. (Using a light theme) If 'only' the text was refreshed then it would not be such an annoyance, it's the whole table colour change from orange and blue (or red and green) to white and back that is the main problem.

Can you try https://github.com/drowe67/freedv-gui/pull/710 and let me know if that works any better?

barjac:
Yes that seems perfect! :) Sorry for the delay but I wanted to use it in a net to be sure all was well.

PR #710, opened 2024-04-07
Revert #704 and use alternate solution.

tmiw:
Per additional feedback, this PR reverts #704 and forces a re-filtering when changing the filter type instead. This should hopefully avoid the unnecessary flickering.


barjac:
This looks good so far but not yet had chance to test it in a net with lots of colour on the screen. I will report back when fully tested.

Issue #720, opened 2024-05-13
Reporter Msg column : Only add ellipsis (...) if the Msg column width will not fit the whole message. 


barjac:
In use I have been finding that many messages have the ellipsis even when I can easily widen my Msg column so it would fit the whole message.

At present widening the column does not expand the message and remove the ellipsis, the user has to still use the tooltip even when room has been made for the whole message.

Improved behaviour:
If a message has an ellipsis, allow widening of the column to adjust the number of characters allowed before an ellipsis is added.

tmiw:
How does https://github.com/drowe67/freedv-gui/pull/721 work for you?

PR #721, opened 2024-05-15
Allow dynamic updating of message column width.

tmiw:
Per request in #720, allow contents of message column to dynamically adjust based on the physical column width.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -899,6 +899,8 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 1. Bugfixes:
     * Revert PR #689 and reimplement fix for original startup delay issue. (PR #712)
+2. Enhancements:
+    * Allow "Msg" column to be resized by the user. (PR #721)
 
 ## V1.9.9 April 2024
 
--- a/src/config/FreeDVConfiguration.cpp
+++ b/src/config/FreeDVConfiguration.cpp
@@ -109,6 +109,8 @@ FreeDVConfiguration::FreeDVConfiguration()
     , monitorTxAudio("/Monitor/TransmitAudio", false)
 
     , txRxDelayMilliseconds("/Audio/TxRxDelayMilliseconds", 0)
+
+    , reportingUserMsgColWidth("/Windows/FreeDVReporter/reportingUserMsgColWidth", 130)
 {
     // empty
 }
@@ -198,6 +200,8 @@ void FreeDVConfiguration::load(wxConfigBase* config)
     load_(config, tabLayout);
 
     load_(config, txRxDelayMilliseconds);
+
+    load_(config, reportingUserMsgColWidth);
 }
 
 void FreeDVConfiguration::save(wxConfigBase* config)
@@ -282,6 +286,8 @@ void FreeDVConfiguration::save(wxConfigBase* config)
     save_(config, monitorTxAudio);
 
     save_(config, txRxDelayMilliseconds);
+
+    save_(config, reportingUserMsgColWidth);
     
     config->Flush();
 }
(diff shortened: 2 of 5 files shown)


barjac:
Hi Mooneer,
Yes that's looking good! Very nearly there ;)
Only minor issue is that when dragging out the column width sometimes the ellipsis does not follow it, so you can't tell whether you have reached the end of the message or not.

To reproduce this grab the right hand Msg heading divider and compress the messages so they all have ellipsis.
Now release the left button.
Now grab the divider again and slowly widen the column. The message width increases and then stops, (still with an ellipsis) no matter how wide the divider is taken, until the button is released, when the remainder of the message appears.

tmiw:
> Hi Mooneer, Yes that's looking good! Very nearly there ;) Only minor issue is that when dragging out the column width sometimes the ellipsis does not follow it, so you can't tell whether you have reached the end of the message or not.
> 
> To reproduce this grab the right hand Msg heading divider and compress the messages so they all have ellipsis. Now release the left button. Now grab the divider again and slowly widen the column. The message width increases and then stops, (still with an ellipsis) no matter how wide the divider is taken, until the button is released, when the remainder of the message appears.

Yeah, that was intentional originally since I didn't think it could be adjusted while dragging. Try now?

barjac:
Just came across another issue:
Sometimes if there is a long message with an ellipsis in the column and this is not of interest so the column is not expanded to include it all then even when the cursor is not over that window, the column expands by a few characters at a time every 5 seconds until the whole of the long message is finally displayed.
Not sure exactly how to reproduce reliably but I did get a screen video of the final few steps of it happening.

https://github.com/drowe67/freedv-gui/assets/1324702/a92b8ce7-6d84-4fa8-8815-e9f246cde339

tmiw:
I think it's fixed with the latest change. Try now?

barjac:
Yes that works fine now. :)

Would it be possible to retain the column width when switching between filters?
I am now (with these recent improvements) setting the Msg column width to the max that suits my screen so that I rarely see the ellipsis, however on switching filter or changing band at the radio, it is reset to the default again.

Maybe it could be retained between sessions too if it is not default.

tmiw:
> Would it be possible to retain the column width when switching between filters? I am now (with these recent improvements) setting the Msg column width to the max that suits my screen so that I rarely see the ellipsis, however on switching filter or changing band at the radio, it is reset to the default again.

Yep and done.

> /OT Have fun at the Hamvention! I just watched your EzDV talk :)

\o/

barjac:
Yes that is working as expected now, any chance of:
"Maybe it could be retained between sessions too?"
...if you agree that it would be useful.

barjac:
Just noticed that the Msg column width setting is lost when switching to analogue and back to digital.

tmiw:
> Just noticed that the Msg column width setting is lost when switching to analogue and back to digital.

I just tried on macOS and Linux and that doesn't seem to happen for me. Can you make sure you're running in the latest in the PR?

tmiw:
> Yes that is working as expected now, any chance of: "Maybe it could be retained between sessions too?" ...if you agree that it would be useful.

Done.

barjac:
> Just noticed that the Msg column width setting is lost when switching to analogue and back to digital.

I think I was being confused by the lack of my extra wide 'version' which was not there when  I switched to analogue, on re-test with latest it does in fact work as expected.

barjac:
> > Yes that is working as expected now, any chance of: "Maybe it could be retained between sessions too?" ...if you agree that it would be useful.
> 
> Done.

Yes that's great now. I think this is now ready. :)

I will test again during the UK Sunday RSGB News and following net and let you know if anything else crops up.

Issue #726, opened 2024-06-14
CMakeLIsts.txt line 51 overrides configuration option -DFREEDV_VERSION_TAG=


barjac:
Can CMakeLists.txt line 51 please be made conditional on -DFREEDV_VERSION_TAG not already being set to a value during cmake configuration.

At present it get clobbered.
Thanks :)

tmiw:
See https://github.com/drowe67/freedv-gui/pull/727 :)

barjac:
Confirm resolved - thanks!

PR #727, opened 2024-06-14
Allow overrriding the version tag when building.

tmiw:
By request in #726, this PR allows overriding of `FREEDV_VERSION_TAG` for custom builds.


Tyrbiter:
Very useful, should have thought of this!

Simplified my local dev builds and I only have to change one number in my spec file.

PR #733, opened 2024-07-13
Add monitor volume adjustment

### LINKED ISSUE #732: Add relative volume settings in config for both transmit monitors
When using the voice keyer or live transmit monitors to hear your own transmission, the volume is generally much higher than necessary.

If there were settings to attenuate these in maybe 1dB steps relative to the RX audio level then comfortable listening levels could be set for each. 

I suspect that the level needed for live TX will need to be considerably lower than that for the voice keyer so I think each does need its own setting.

tmiw:
Resolves #732 by adding new right-click menu items for adjusting monitoring volume under PTT and Voice Keyer. Example:

image<br/>
image

Clicking outside of the box that appears will close it. Additionally, updates here are live (i.e. if you're transmitting, you can hear the volume changing as you adjust the slider).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -896,6 +896,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 2. Enhancements:
     * Show green line indicating RX frequency. (PR #725)
     * Update configuration of the Voice Keyer feature based on user feedback. (PR #730)
+    * Add monitor volume adjustment. (PR #733)
 3. Build system:
     * Allow overrriding the version tag when building. (PR #727)
     * Update wxWidgets to 3.2.5. (PR #731)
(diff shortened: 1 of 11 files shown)


barjac:
Very nice!
However the slider widget melts into the display elements behind it especially when the main window is using multiple widget columns either side of the scope.

Is it possible to put a border line around it? 
Maybe even dim the whole display behind it while it has focus?
Or maybe give it a different background colour or invert it's foreground/background?

Screenshot_20240713_104349a

Tyrbiter:
The only thing I will mention as a minor thing, the new monitor settings and code have the word Monitor typo'd as Montior, it's not urgent but probably best to fix it sooner rather than later.

Will report on the actual function of the change later.

tmiw:
> Is it possible to put a border line around it? Maybe even dim the whole display behind it while it has focus? Or maybe give it a different background colour or invert it's foreground/background?

There's a border around it now as of 09b7275.

> The only thing I will mention as a minor thing, the new monitor settings and code have the word Monitor typo'd as Montior, it's not urgent but probably best to fix it sooner rather than later.

Fixed.

barjac:
Yes that looks and works fine for me now.
Nice job Mooneer! :)

PR #735, opened 2024-07-15
Avoid modifying the audio device configuration without the user explicitly doing so.

### LINKED ISSUE #734: If a USB audio card is unplugged FreeDV wipes audio settings for that card.
I have hit this many times and it is really frustrating and should not happen.

I use an external USB (CM108 based) audio dongle to handle audio to and from my transceiver.

I sometimes use that USB port with extension cable to connect other devices when not using FreeDV.

If I forget to re-connect it to the CM108 device afterwards and start FreeDV, then instead of just warning that the device is missing, FreeDV wipes it from the configuration as well!

Why, when it only needs to remind me to plug in the correct device and re-try?

tmiw:
Resolves #734 by doing the following:

1. Remove the logic that changes the audio device configuration after a failure. This was originally intended to force the analog side to use the system's default audio devices (i.e. if a USB headset was removed) but is likely still confusing to users.
2. Slightly update the error message to "Your %s device cannot be found and may have been removed from your system. *Please reattach this device and* go to Tools->Audio Config... to confirm your audio setup." (Note italics for new/modified text.)

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -897,6 +897,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Show green line indicating RX frequency. (PR #725)
     * Update configuration of the Voice Keyer feature based on user feedback. (PR #730)
     * Add monitor volume adjustment. (PR #733)
+    * Avoid modifying the audio device configuration without the user explicitly doing so. (PR #735)
 3. Build system:
     * Allow overrriding the version tag when building. (PR #727)
     * Update wxWidgets to 3.2.5. (PR #731)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3147,8 +3147,8 @@ bool MainFrame::validateSoundCardSetup()
     else if (!canRun)
     {
         wxMessageBox(wxString::Format(
-            "Your %s device cannot be found and may have been removed from your system. Please reattach this device and go to Tools->Audio Config... to confirm your audio setup.", 
-            failedDeviceName), wxT("Sound Device Removed"), wxOK, this);
+            "Your %s device cannot be found and may have been removed from your system. Please reattach this device, close this message box and retry. If this fails, go to Tools->Audio Config... to check your settings.", 
+            failedDeviceName), wxT("Sound Device Not Found"), wxOK, this);
     }
     
     engine->stop();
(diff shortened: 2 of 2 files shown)


barjac:
> Resolves #734 by doing the following:
> 
>     1. Remove the logic that changes the audio device configuration after a failure. This was originally intended to force the analog side to use the system's default audio devices (i.e. if a USB headset was removed) but is likely still confusing to users.
> 
>     2. Slightly update the error message to "Your %s device cannot be found and may have been removed from your system. _Please reattach this device and_ go to Tools->Audio Config... to confirm your audio setup." (Note italics for new/modified text.)

This is still confusing as it tells users who have simply unplugged a device to go to (scary for some) Tools->Audio Config... when all they need to do is re-plug the device, close the message dialog and hit Start (modem) again.

I tested this as follows:
1. Confirm I have a running system
2. Stop modem and program
3. Un-plug audio device
4. Re-start FDV
5. Hit Start (modem)
6. Got message
7. 
Screenshot_20240715_195238

8. Tried to go to Tools->Audio Config but of course can't as the message dialogue is still open.
9. Now I was unsure - do I re-plug the device before or after OKing the dialogue?
10. I re-plugged after and then went to Tools->Audio Config.
11. All was good, I know what looks good and what does not having had to re-configure over and over. Some users will hopefully only configure once so will not easily know.
12. Rather than go straight to Tools->Audio Config I tested the following:

a. I again followed 1..6
b. On getting the message I re-connected the USB audio device
c. Closed the dialogue with OK
d. Hit Start (modem) and all was back to normal.

I tested using both USB sound devices (audio to/from radio and USB headset) with the same results.
I then tried removing both together and on receiving the message I purposely inserted the wrong one (not in the message) first. I then got the correct message for the second one and again closed the dialogue before again hitting Start (modem) which worked fine.

I think the message should read:

"Your %s device cannot be found and may have been removed from your system. Please reattach this device, close this message box and retry. If this fails go to Tools->Audio Config... to check your settings."

tmiw:
I think people would typically know to push the OK button first, but I guess the proposed message doesn't hurt. Updated.

barjac:
> I think people would typically know to push the OK button first, but I guess the proposed message doesn't hurt. Updated.

Many thanks as always.

I appreciate that  most people would know that, however considering that hitting OK previously screwed up the settings I thought that precise instructions would provide some reassurance to anyone who does not read release notes.

Looks fine now for me :)

PR #738, opened 2024-08-08
If provided by user, add config file to titlebar.

### LINKED ISSUE #737: Add config file name to main and reporter window headings if not default.
Problem:
When using multiple instances of both freedv-gui and it's linked freedv reporter on one machine on different desktops it can be difficult to easily identify which instance is on the screen.

Possible solution:
Since both instances will be using different configuration files the config file name could be used to act as identification if it was added to the header message. It could be excluded if the default config file name was in use.

tmiw:
Resolves #737 by adding the path of a custom config file to the title bar. For example, if FreeDV is started via the following:

```
freedv -f `pwd`/test.conf
```

the titlebar could look something like this:

image

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -204,6 +204,12 @@ bool MainApp::OnCmdLineParsed(wxCmdLineParser& parser)
         fprintf(stderr, "Loading configuration from %s\n", (const char*)configPath.ToUTF8());
         pConfig = new wxFileConfig(wxT("FreeDV"), wxT("CODEC2-Project"), configPath, configPath, wxCONFIG_USE_LOCAL_FILE);
         wxConfigBase::Set(pConfig);
+        
+        // On Linux/macOS, this replaces $HOME with "~" to shorten the title a bit.
+        wxFileName fn(configPath);
+        fn.ReplaceEnvVariable("HOME", "~");
+        
+        customConfigFileName = fn.GetShortPath();
     }
     pConfig->SetRecordDefaults();
     
@@ -240,7 +246,7 @@ bool MainApp::OnInit()
     // displayed. But it doesn't when built from command line.  Why?
 
     frame->m_auiNbookCtrl->ChangeSelection(0);
-    frame->Layout();
+    frame->Layout();    
     frame->Show();
     g_parent =frame;
 
@@ -648,6 +654,12 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent, wxID_ANY, _("FreeDV ")
     pthread_setname_np(pthread_self(), "FreeDV GUI");
 #endif // defined(__linux__)
 
+    // Add config file name to title bar if provided at the command line.
+    if (wxGetApp().customConfigFileName != "")
+    {
+        SetTitle(wxString::Format("%s (%s)", _("FreeDV ") + _(FREEDV_VERSION), wxGetApp().customConfigFileName));
+    }
+    
     m_reporterDialog = nullptr;
     m_filterDialog = nullptr;
 
--- a/src/main.h
+++ b/src/main.h
@@ -168,6 +168,7 @@ class MainApp : public wxApp
         bool                    CanAccessSerialPort(std::string portName);
         
         FreeDVConfiguration appConfiguration;
+        wxString customConfigFileName;
         
         // PTT -----------------------------------    
         unsigned int        m_intHamlibRig;


barjac:
Working fine in main but not showing in reporter.

tmiw:
> Working fine in main but not showing in reporter.

Sorry, I missed the part in the original issue about FreeDV Reporter. I'm not sure about doing this there, too, since that would make the version string significantly longer. Could this be something that can be done manually via the existing Message field?

barjac:
> > Working fine in main but not showing in reporter.
> 
> Sorry, I missed the part in the original issue about FreeDV Reporter. I'm not sure about doing this there, too, since that would make the version string significantly longer. Could this be something that can be done manually via the existing Message field?

I don't understand, we are only talking about the reporter window decoration

tmiw:
OK, I think I understand now. Apologies for the confusion.

Anyway, I tried running multiple FreeDV instances on both GNOME and KDE to see what would happen. GNOME at least seems to show separate icons for each running FreeDV process (the generic "gear" one for some reason):

image

I can see how it'd be confusing on KDE, though:

image

Unfortunately I can't seem to find a setting on KDE that governs this behavior. I wonder if there's something that can be done at the app level to make it behave more like GNOME (i.e. display separate icons for each separate running process). If so, perhaps that's a better solution than displaying the config filename in every window that's open?

barjac:
In latest KDE Plasma6 there are separate task bar items for 'FreeDV' and 'FreeDV Reporter' rather than the Plasma5 method that shows FreeDV+ on a single taskbar item when there is more than one window open belonging to FreeDV.

When using multiple desktops it is still possible to inadvertently click a task bar item belonging to a different desktop such that you are seeing instance 1 of FreeDV and instance 2 of Reporter.

Hence my request to have a simple indication of which is which in the tile bars of both.

In use I have found the addition of the config filename really useful to see at a glance which is on screen.

Can the config filename be added to the reporter window titlebar string as in the main window, or is there technical reason that does not allow this?

tmiw:
> In use I have found the addition of the config filename really useful to see at a glance which is on screen. I don't think trying to change 'icons' is the way to go.

Noted. It's probably still a good idea to make the Linux behavior match Windows and macOS, though; showing the gear icon (with GNOME for example) isn't ideal no matter how many instances of FreeDV you're running. This can be done as a separate PR for sure.

> Can the config filename be added to the reporter window title bar string as in the main window, or is there technical reason that does not allow this?

AFAIK I don't think there's a technical reason, just figuring out if there's a better solution that will help achieve what you're looking for (ensuring you don't accidentally pull up the wrong window(s)).

tmiw:
So I have a better understanding, I went ahead and started my Fedora 40 VM again (KDE Plasma 6.1.3). I created a second virtual desktop and moved one of the FreeDV instances to that. When I hover over its icon, I only see the windows for the FreeDV instance on the current desktop as shown below:

image

However, as shown above, I can click on one of the icons next to the KDE button to switch between desktops. Are you inadvertently clicking on one of those icons and pulling up the wrong FreeDV that way, or does how KDE look for me totally not match what you're seeing? 

Also, wouldn't this confusion theoretically happen with other FreeDV windows as well (i.e. the filter one), or is there something unique about the FreeDV Reporter window that makes the config file path needed there?

Tyrbiter:
FWIW if you use a Fedora package (either distro or self built) then the FreeDV icon will show up under GNOME, not sure about KDE as I rarely change DE.

barjac:
I just spent an hour replying to everything with images etc. and then lost the lot.
No time now to re-do it all so it will have to wait. Sorry.

barjac:
> So I have a better understanding, I went ahead and started my Fedora 40 VM again (KDE Plasma 6.1.3). I created a second virtual desktop and moved one of the FreeDV instances to that. When I hover over its icon, I only see the windows for the FreeDV instance on the current desktop as shown below: (image removed here but shown in a previous post) 

> However, as shown above, I can click on one of the icons next to the KDE button to switch between desktops. Are you inadvertently clicking on one of those icons and pulling up the wrong FreeDV that way, or does how KDE look for me totally not match what you're seeing?

I agree after much testing that it is as you say although displayed differently in Plasma. (I don't use Gnome).

This shows a typical screen in Plasma showing the main FreeDV desktop with a remote SDR shown on all desktops. 
Screenshot_20240813_223401

This is in the same session with another instance of FreeDV on desktop 2 monitoring the same SDR using my callsign but displaying the locator of the remote SDR.
Screenshot_20240813_223512

This all works fine but as you say it's easy to get in a muddle with a wrong click and for this reason I feel there should be some way to quickly identify which windows belong to which instance.

> Also, wouldn't this confusion theoretically happen with other FreeDV windows as well (i.e. the filter one), or is there something unique about the FreeDV Reporter window that makes the config file path needed there?

Yes exactly, this had also crossed my mind but since I sensed some resistance to this idea I was reluctant to mention it. Ideally ALL configuration dialogue window titlebars would benefit from this too.

tmiw:
I'm not 100% convinced, but it's an easy enough change so I went ahead and added it to the FreeDV Reporter window. We can always back it out if it ends up not working out. Let me know how the latest commit in this PR goes.

barjac:
This is OK for me and makes it easy to check which instance is which.
I would prefer only the filename without path and extension to allow the use of multi-level paths from cluttering the titlebar, but this is already very good.
I think that all the config dialogue window titles should have the same treatment.
This would not affect users with only the default config file, so I see no detrimental impacts from incorporating this feature.

tmiw:
> I would prefer only the filename without path and extension to allow the use of multi-level paths from cluttering the titlebar, but this is already very good.

Any possibility of having multiple levels of folders for configs (similar to the voice keyer files from the previous PR)? Might still be a good idea to keep the folder names displayed too. There's also already some shortening of the name by using ~ instead of the full `/home/username` at the beginning.

> I think that all the config dialogue window titles should have the same treatment.

I took a look and it seems that the remaining ones would be pretty much just setup type dialogs (e.g. Tools->Audio Options). Any use case where you'd be changing settings in more than one copy of FreeDV at a time?

barjac:
> > I would prefer only the filename without path and extension to allow the use of multi-level paths from cluttering the titlebar, but this is already very good.
> 
> Any possibility of having multiple levels of folders for configs (similar to the voice keyer files from the previous PR)? 

Sorry I don't understand the question.

User config files can already be anywhere as any full path can be passed with the -f option. 

>Might still be a good idea to keep the folder names displayed too. 

Why is it useful to show any of the path in the titlebar when the user already MUST know where his config file is as he already passed it as a parameter on startup!

> > I think that all the config dialogue window titles should have the same treatment.
> 
> I took a look and it seems that the remaining ones would be pretty much just setup type dialogs (e.g. Tools->Audio Options). Any use case where you'd be changing settings in more than one copy of FreeDV at a time?

It's quite possible that since we cannot simply switch to RX only or stop reporting without stopping the modems and opening config files, that yes, we may be editing multiple config files at the same time.

If I were to use e.g "TS450S" as a config filename and put it in ~/ then OK it would keep the titlebar quite clean with "FreeDV 1.9.9 (TS450S)" however my config files would be scattered around my user home folder depending on the filename. - Messy!
If I add a config directory to keep them in one place then I get a messy titlebar with superfluous text.

I see this as very similar to the way other softwares work e.g. dolphin file manager that adds the remote machine hostname or IP to the titlebar when working on a remote machine. Or klog logging program which adds the log name that it is using to the titlebar.

tmiw:
> Why is it useful to show any of the path in the titlebar when the user already MUST know where his config file is as he already passed it as a parameter on startup!

Good point. I shortened it to only show the filename. Still leaves the scenario where you have the same-named files in e.g.

* `~/x/file.conf`
* `~/y/file.conf`

But this is probably pretty unlikely.

> It's quite possible that since we cannot simply switch to RX only or stop reporting without stopping the modems and opening config files, that yes, we may be editing multiple config files at the same time.
> 
> If I were to use e.g "TS450S" as a config filename and put it in ~/ then OK it would keep the titlebar quite clean however my config files would be scattered around my user home folder depending on the filename. - Messy! If I add a config directory to keep them in one place then I get a messy titlebar with superfluous text.
> 
> I see this as very similar to the way other softwares work e.g. dolphin file manager adds the remote machine hostname or IP to the titlebar when working on a remote machine. Or klog logging program which adds the log name that it is using to the titlebar.

*shrug* I added the name to all the other windows. We can revisit if that ends up becoming a problem.

barjac:
Perfect!
Many thanks for this, it is working exactly as I envisaged. No confusion possible now.
Using e.g. "TS450S" as one of the config file names which are all now in ~/.FreeDV/ directory I can have all the special config files together and the windows all show like:

Screenshot_20240825_213100

Brilliant!

Tyrbiter:
Looks fine to me, I reckon it can be merged into master.

PR #769, opened 2024-11-11
Report RX of RADE signals without callsign or SNR.

tmiw:
Per November 2024 PLT meeting, this PR adds special behavior for RADE that allows reporting to FreeDV Reporter that we're receiving *something* in RADE on the user's current frequency whenever we're in sync. Callsign is set to 'unk' with SNR to 0 as we're currently unable to retrieve either.


Tyrbiter:
Does the server support this now?

tmiw:
> Does the server support this now? Ah, yes it does.
> 
> I notice that the RADE rx report seems to stay up for some time after the signal has gone, maybe it needs something to clear it after a delay.

There's a 20 second timeout on the server side already, but that doesn't take into account that this code sends the RX report once a second or so (vs. once maybe every 5 seconds with the other modes). It could be cleared by just sending an empty mode and callsign on losing sync (like what happens when switching frequencies) but then we'd lose the 'RADE' RX mode display in the web interface. Does that sound like a reasonable tradeoff?

Tyrbiter:
> > I have not yet been able to see the sequence play out, but I found the RADE Rx mode showing for my callsign until I refreshed the reporter page.
> 
> Yeah, that's normal behavior for other modes (the RX callsign/mode/SNR sticking around until frequency changes or the user receives someone else).

Probably best left as it is then, I'd say. No benefit in making it more complicated, watching in real time should give a fairly good idea of which transmission a given station, or stations, can hear.

Issue #777, opened 2024-12-01
FreeDV resampling filters


drowe67:
In #776 there was some experimentation with resampling libraries in at attempt to mimimise distortion measured with the RADE `loss.py` tool.  However - we can do (and perhaps should do) better than swapping out 3rd party libraries and hoping for the best.  We have the knowledge to build our own, e.g. examples for fixed sample rate conversion ratios in fdmdv, and in radae/dsp.py have code for generating the filter coeffs at init time.  That would give us testable (against RADE) resampling code we understand completely, and can be tuned to be fit for purpose.  An alternative is to devise a suite of tests to evaluate third party libraries before adopting them.

As #776 & #761 (the PR around sample slips) has shown - the signal processing is very important and needs to be designed, engineered, and tested carefully.   We need to be very careful with making ad-hoc design changes (e.g. optimisation, changes to the audio I/O) without checking the signal processing fundamentals.   Recent changes  freedv-gui to support automated testing of the signal processing path, and the RADE `loss.py` tool will help here.

This Issue is a placeholder for notes on the topic of resampling filters in `freedv-gui`.  Before kicking off any work it should be discussed & approved at PLT.

tmiw:
There was actually a resampling ctest prior to the RADE work (src/pipeline/tests/ResampleTest.cpp). Currently it just generates a sine wave for one second and makes sure that the number of output samples is +/- 10% of the theoretical number (for example, if you go from 8 kHz to 48 kHz, the output can be anywhere between 43,200 samples and 52,800 samples long). This test could be restructured to check other parameters instead, like RADE loss vs. reference.

drowe67:
> Another thing--we probably need to be careful how testing is done. ....

If we get the resampler right , e.g. a transfer function of 1 between a pre-defined mask of frequencies - we can resample as many times as we like without any distortion and all these test/configuration problems go away.  This is why it's important to really think carefully about the resampler before we do any more coding in this area.

tmiw:
> If we get the resampler right , e.g. a transfer function of 1 between a pre-defined mask of frequencies - we can resample as many times as we like without any distortion and all these test/configuration problems go away. This is why it's important to really think carefully about the resampler (and set requirements) before we do any more coding in this area.

Should we go ahead and merge what's already in #776, then? Or is there additional work that's needed there?

drowe67:
[...]

PR #779, opened 2024-12-04
Investigate soxr as a replacement resampling library

tmiw:
Continuation of PR #776 with the changes needed to switch from libsamplerate to soxr as the resampler. 

TODO:

- [ ] Develop test procedure for comparing soxr with libsamplerate
- [ ] Test soxr at various quality levels against existing libsamplerate code
- [ ] TBD


drowe67:
@tmiw - I do not feel there is any justification for this PR - and the decision to do this work should be a decision made by the PLT.

Resamplers are not magic and we are not going to get significant improvements by swapping libraries in and out. A much better approach is our own, well understood resamplers, as per https://github.com/drowe67/freedv-gui/issues/777.

tmiw:
We can close, but I'd say not to delete this branch in case we decide to start working on #777 at some point in the future and find that the soxr work already done may be useful.

PR #780, opened 2024-12-04
Fix stuttering during EOO TX.

tmiw:
Resolves the following EOO related transmit issues:

1. Prevents multiple transmission of EOO by adding logic to suppress its generation during a TX cycle once it's requested.
2. Fixes issue preventing EOO from being transmitted when using the voice keyer.
3. Adjusts scaling of EOO to match regular RADE signal (i.e. multiply by 16383 instead of 32767).


tmiw:
Going to merge since #783 also has these changes. We can edit that PR if needed.

PR #794, opened 2025-01-05
If disabled, make sure we don't try calling rade_text_* with a null pointer.

tmiw:
This PR prevents calling `rade_text_rx()` on EOO if the user's disabled reporting. Without this change, errors similar to the following are possible:

image (1)

### DIFF
--- a/src/pipeline/RADEReceiveStep.cpp
+++ b/src/pipeline/RADEReceiveStep.cpp
@@ -110,12 +110,12 @@ std::shared_ptr<short> RADEReceiveStep::execute(std::shared_ptr<short> inputSamp
             int hasEooOut = 0;
             float eooOut[rade_n_eoo_bits(dv_)];
             nout = rade_rx(dv_, features_out, &hasEooOut, eooOut, input_buf_cplx);
-            if (hasEooOut)
+            if (hasEooOut && textPtr_ != nullptr)
             {
                 // Handle RX of bits from EOO.
                 rade_text_rx(textPtr_, eooOut, rade_n_eoo_bits(dv_) / 2);
             }
-            else
+            else if (!hasEooOut)
             {
                 if (featuresFile_)
                 {


Tyrbiter:
I think this commit may have stopped EOO text working in the most recent couple of builds I have tested.

Tyrbiter:
I have dropped this commit and will report what happens once I find some on air signals from stations with EOO-text enabled. There are a few in the UK.

barjac:
Are you receiving?

tmiw:
I wonder if it's some sort of audio delay between the application and say, pipewire. Maybe you can try different values for "RX/TX Delay" (Tools->Options->Rig Control) and seeing if that helps?

Tyrbiter:
Some experimentation this evening revealed that even with good 10dB+ SNR on RADE we are failing to get EOO-Text detection on a significant percentage of overs. What happens is that as the transmitting station drops PTT there is almost no noise heard if EOO detection is successful, but if it isn't there is a noticeable amount of garbage "speech-alike" sound which is a bit like 700 modes sound with marginal error rates.

I don't think this particular commit is the problem, but maybe some more debug output from the receive side at end of over would be useful.

tmiw:
> Some experimentation this evening revealed that even with good 10dB+ SNR on RADE we are failing to get EOO-Text detection on a significant percentage of overs. What happens is that as the transmitting station drops PTT there is almost no noise heard if EOO detection is successful, but if it isn't there is a noticeable amount of garbage "speech-alike" sound which is a bit like 700 modes sound with marginal error rates.
> 
> I don't think this particular commit is the problem, but maybe some more debug output from the receive side at end of over would be useful.

It does kinda sound like PTT may be dropping too soon at least some of the time. I'd suggest trying the RX/TX Delay option I mentioned above;  perhaps there's some value for delay that will make decodes more likely. 

Also, do you have any recordings that you can try playing back? We can at least see if that decodes reliably too.

Tyrbiter:
I would expect PTT to be held until EOO-Text blocks are sent, and indeed the debug shows that:

20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/TxRxThread.cpp:649: Triggering sending of EOO
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:178: Queueing 1152 EOO samples to output FIFO
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 992)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 832)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 672)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 512)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 352)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 192)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 32)
20:56:48 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/RADETransmitStep.cpp:97: Returning 32 EOO samples (remaining in FIFO: 0)
20:56:49 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/ongui.cpp:911: All TX finished, going out of PTT

What I was wondering is if some extra debug might be useful before these messages are output:

20:53:07 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/rade_text.c:210: mean amplitude: 1.798888
20:53:07 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/rade_text.c:220: Estimated BER: 0.000000
20:53:07 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/pipeline/rade_text.c:334: rxCRC: 7, calcCRC: 7, decodedStr: G4MKT
20:53:07 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/freedv_interface.cpp:114: FreeDVInterface::OnRadeTextRx_: received G4MKT
20:53:08 INFO /home/bdm/git/freedv-gui-v2.0-ms-rade-snr/src/main.cpp:1920: Reporting callsign G4MKT @ SNR 10, freq 3642500 to reporting services.

I have not looked at the rade_text.c code to see what might be available at that point, but maybe we need to see what has been detected to actually call the code that output the debug above. Is that part of the RADE demodulator/decoder function?

Tyrbiter:
That's very useful information @barjac,  @tmiw can have a look at the different code paths to see what might be happening.

tmiw:
I created https://github.com/drowe67/freedv-gui/pull/800 as a place to put some WIP stuff while we figure this out. It's possible that the PR as is will help, but I'm not fully sure yet. Give it a try and leave comments there?

barjac:
I have done some more testing tonight using latest ms-rade-snr (#d47135)

Following on from "I'd suggest trying the RX/TX Delay option I mentioned above; perhaps there's some value for delay that will make decodes more likely."

Yes, that makes a difference.
In a nutshell, 100ms is not enough, around 200ms and above seems to cause the receiving freedv instance to crash and 300ms on my system is working correctly with occasional crashes.

G4DYA was using this version today and he was also crashing during RX during the net.

First the crash:
`Traceback (most recent call last):
  File "/home/baz/freedv-rade/freedv-gui/build_linux/rade_src/radae_rxe.py", line 163, in get_snrdB_3k_est
    return int(self.receiver.snrdB_3k_est)
ValueError: cannot convert float NaN to integer
Error: call to get_snrdB_3k_est`

I will re-test the delay in #93c1a0 which did not crash.

tmiw:
Looks like #799 got created for the crash bug, so we'll add further comments over there.

g4dya:
barjac and I did a test earlier today, both running 2a40daab with no extra TX/RX delay in the Rig Control tab. Both eoo callsigns and snr are now showing in FreeDV reporter reliably. The callsign seems to need an snr of a plus a few dB to work reliably, a bit more than needed to decode speech.

PR #797, opened 2025-01-07
Integrate SNR reporting from latest RADE.

tmiw:
Adds support for SNR reporting from the latest RADE code.

### DIFF
--- a/cmake/BuildRADE.cmake
+++ b/cmake/BuildRADE.cmake
@@ -13,7 +13,7 @@ ExternalProject_Add(build_rade
    SOURCE_DIR rade_src
    BINARY_DIR rade_build
    GIT_REPOSITORY https://github.com/drowe67/radae.git
-   GIT_TAG main
+   GIT_TAG dr-est_snr
    CMAKE_ARGS ${RADE_CMAKE_ARGS}
    #CMAKE_CACHE_ARGS -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=${CMAKE_OSX_DEPLOYMENT_TARGET}
    INSTALL_COMMAND ""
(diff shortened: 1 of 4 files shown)


Tyrbiter:
I also noticed that the reporter is still showing SNR after Rx has finished. Is that intended? I forget what it does with the 700 modes.

tmiw:
> Could this affect the EOO text? I'm not seeing any call sign decodes but the 3 stations on the UK net are all using this commit so maybe it has upset the EOO sending or receiving.

This is just display/reporting type stuff, so I don't foresee it affecting EOO decode (especially since we're not keying reporting of callsigns off of SNR AFAIK).

> I also noticed that the reporter is still showing SNR after Rx has finished. Is that intended? I forget what it does with the 700 modes.

IIRC 700D etc. still showed an SNR, albeit something low like -5 or something depending on the band noise. The latest commit just shows "--" for RADE and the case when we're not in sync, but maybe we want to do the same for the other modes too?

Tyrbiter:
> IIRC 700D etc. still showed an SNR, albeit something low like -5 or something depending on the band noise. The latest commit just shows "--" for RADE and the case when we're not in sync, but maybe we want to do the same for the other modes too?

It would make sense for them all to behave in the same manner, a blank SNR on no sync makes sense to me.

tmiw:
> > IIRC 700D etc. still showed an SNR, albeit something low like -5 or something depending on the band noise. The latest commit just shows "--" for RADE and the case when we're not in sync, but maybe we want to do the same for the other modes too?
> 
> It would make sense for them all to behave in the same manner, a blank SNR on no sync makes sense to me.

Done.

tmiw:
Merging so we can have these changes for the other PR. We can open a new PR if any problems come up.

PR #803, opened 2025-01-13
Fix bug preventing Request QSY button from being enabled in RADE mode.

tmiw:
This PR fixes an issue preventing the Request QSY button in the FreeDV Reporter window from working when in RADE mode.

(Thanks @Tyrbiter for reporting this issue via email.)


Tyrbiter:
This seems to have fixed the problem with RADE. And indeed it's enabling the QSY button when the frequency is different in any mode, which it may not have been but I wasn't sure. Gets my vote.

PR #804, opened 2025-01-13
Fix issue possibly degrading RADE RX audio quality.

tmiw:
Walter K5WH reported that the latest test build of 2.0 didn't sound as good on RADE as the 20241018 version. Using the RADE tools and the automated test framework, I recorded a loss of 0.182 between freedv-gui and `radae_rx` in the RADE repo. After implementing the code change in this PR, the loss dropped to 0.001.

(See companion issue https://github.com/drowe67/radae/issues/41 indicating minimal loss between Python and C decoders.)

### DIFF
--- a/src/pipeline/RADEReceiveStep.cpp
+++ b/src/pipeline/RADEReceiveStep.cpp
@@ -102,7 +102,7 @@ std::shared_ptr<short> RADEReceiveStep::execute(std::shared_ptr<short> inputSamp
             // demod per frame processing
             for(int i=0; i<nin; i++) 
             {
-                input_buf_cplx[i].real = (float)input_buf[i] / 32767.0;
+                input_buf_cplx[i].real = input_buf[i] / 32767.0;
                 input_buf_cplx[i].imag = 0.0;
             }
 


Tyrbiter:
After a quick search, it seems that float gives 6 or 7 decimal places of precision, while double gives 15 digits. That could explain it, unless something changed in the compiler.

tmiw:
> After a quick search, it seems that float gives 6 or 7 decimal places of precision, while double gives 15 digits. That could explain it, unless something changed in the compiler.

My VM that I use for Windows builds got trashed recently so I had to rebuild it. It's possible that messed something up too (i.e. due to using a different llvm-mingw than before). Though I was also noticing higher loss on macOS before this change. 

Anyway, you didn't happen to notice any quality differences with this change, right?

Tyrbiter:
Between the October release and this commit in early December:

https://github.com/drowe67/freedv-gui/pull/776

the way the incoming data is processed changed a bit.

Personally I didn't notice much but then I keep running new code as it evolves so maybe the quality difference passed me by and off-air signals tend to be quite variable.

Tyrbiter:
Might there be a connection with this issue https://github.com/drowe67/freedv-gui/issues/777 ?

PR #808, opened 2025-01-14
Fix bug causing hangs in non-RADE mode at end of TX.

tmiw:
Reported by the RADE test group. This PR modifies the end of TX logic to add a missed check for RADE mode before waiting for EOO to be queued.


Tyrbiter:
Testing appears to show this works as expected in both RADE and legacy modes, at least in terms of expected debug output.

PR #809, opened 2025-01-14
Add support for only adjusting frequency, not mode.

tmiw:
Per request from the RADE test group, this PR adds an additional rig control option to allow only adjusting frequency and not mode. This is due to mode adjustment being fairly buggy for some radios in hamlib.

Example UI:

Screenshot 2025-01-14 at 12 49 11 PM

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -905,6 +905,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Minor adjustments to spectrum/waterfall tooltips. (PR #743)
     * Implement new logging framework. (PR #773)
     * Windows: Detect whether microphone permissions have been granted and display error if not. (PR #790)
+    * Add rig control option to prevent auto-adjustment of the radio's current mode. (PR #809)
 3. Build system:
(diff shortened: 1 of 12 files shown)


Tyrbiter:
I think this is a good option to have, despite enormous effort on hamlib-4.6 some of the mode switching and filter selection code is not finished because actually the radios have differences even between very similar models from the same manufacturer.

tmiw:
> I think this is a good option to have, despite enormous effort on hamlib-4.6 some of the mode switching and filter selection code is not finished because actually the radios have differences even between very similar models from the same manufacturer.

Yeah. Not strictly RADE specific but if it ultimately reduces support workload on FreeDV's part...

Anyway, does the added feature seem to work for you? I tested locally and it seems to for me, anyway.

Tyrbiter:
I have tested it here using RADE mode and it's fine, selecting frequency only no longer sets Digital mode on the TS-890. Together with the DIGU/DIGL setting these appears to cover pretty much all possible combinations.

I also have the EOO for non-RADE modes change in use, that too seems to be fine, it doesn't obviously break anything when transmitting and looking at the debug output, but I have not tried it in QSO.

PR #812, opened 2025-01-15
Make sure reporting message is sent using UTF8.

tmiw:
On Windows, it's apparently not possible to use anything other than English letters and numbers for the FreeDV Reporter message dialog (whereas it seems to work at least in macOS). This PR is a WIP that will hopefully allow the use of foreign language text and emojis.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -896,6 +896,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix bug preventing proper time display in FreeDV Reporter on macOS. (PR #748)
     * Set timeout for Hamlib comms to avoid GUI getting stuck. (PR #746)
     * Fix various audio dropout issues, especially on Linux. (PR #761)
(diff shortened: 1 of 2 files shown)


tmiw:
Don't have time to test this right now but maybe if Linux shows the same behavior (and is fixed by this PR) we can go with that.

@Tyrbiter, @barjac

Tyrbiter:
> Don't have time to test this right now but maybe if Linux shows the same behavior (and is fixed by this PR) we can go with that.
> 
> @Tyrbiter, @barjac

I've built this version and added some accented characters to the Message: box in the reporter window, then sent the message which has appeared as it should on the web and app reporter.

So that far it works, and presumably it should allow pretty much any charset that maps into UTF8 to be entered.

tmiw:
I was able to test on Windows. Looks like Japanese text appears fine on the website but not in the app yet. It's probably a similar change needed on the RX side.

tmiw:
Latest change looks like it did the trick, so I'll go ahead and merge.

Issue #815, opened 2025-01-16
[Software Issue] SNR display in freedv-gui


Tyrbiter:
### Platform

Linux

### Platform Version

Fedora 41

### FreeDV Version

FreeDV 2.0.0-devel-9de0abbb plus PR#813 and #814 fixes

### Steps to Reproduce

Look at the SNR graph with high SNR signals, it reaches the graph end point as the numerical value exceeds 20dB.

### Expected Behavior

I think that the graph should show SNRs higher than 20dB, RADE QSOs often reach above this level with stronger RF signal levels.

It would make sense to increase the range of the SNR graph, the exact value needed is at least 25dB perhaps higher.

### Actual Behavior

SNR graph limits at 20dB.

g4dya:
I have seen RADE SNRs up to 32 dB, so I would suggest a 40 dB limit.

tmiw:
Done. See https://github.com/drowe67/freedv-gui/pull/817.

Tyrbiter:
All working nicely now.

PR #816, opened 2025-01-17
Increase max SNR in the UI to 40 dB.

tmiw:
By request in #815, this PR increases the maximum SNR to 40 dB.


tmiw:
Hang on, something's off with this PR. Going to recreate.

PR #817, opened 2025-01-17
Increase max SNR in the UI to 40 dB.

tmiw:
By request in #815, this PR increases the max SNR to 40 dB.


Tyrbiter:
This seems to work as intended.

If RADEV2 does eventually demonstrate an SNR capability lower than RADEV1 then it would be worth changing the minimum SNR value to reflect this.

PR #818, opened 2025-01-19
Add missed save/load for new frequency change option.

tmiw:
Followup for #809 to add missed code to actually persist the new config option. Doh.


Tyrbiter:
This works for me and keeps the rig control settings over a restart although I had not noticed it failing previously.

PR #828, opened 2025-02-11
Don't adjust Msg column width when user disconnects.

### LINKED ISSUE #821: FreeDV Reporter 'Msg' column width resets to stored width every time a station leaves a net (stops modem).
### Platform

Linux

### Platform Version

Mageia, Ubuntu & Windows

### FreeDV Version

Happening in original Windows RADE version and latest git RADE1 in linux

### Steps to Reproduce

Join a group/net with the reporter filter set to track frequency.
Set the 'Msg' column width so the 'Last Update' column is within the screen width.
Wait for a station to leave and note that the 'Msg' column width more than doubles.

EDIT:
On further testing I have realized that it actually returns to the width that was set on a previous program close, which explains why this has not been reported previously. Most stations probably don't adjust the column widths, especially if they are using high aspect ratio screens.

### Expected Behaviour

During a session the 'Msg' column width should not change other than when manually adjusted by the user, even if a non-default config setting has been saved.

### Actual Behavior

I have been seeing this sudden increase in 'Msg' column width apparently happening at random, but only today managed to track down how to reproduce it for a bug report.

With six stations in the net today and with some specific testing after the net with different operating systems we managed to reliably reproduce it as above.

### Additional Comments

_No response_

tmiw:
Resolves #821 by suppressing auto-width changes for the Msg column regardless of any changes in the contents.


barjac:
I just did a quick test and this does fix the issue when a station leaves, but it now/still happens when a station joins.
I had not noticed this previously as the 'leaving' bug masked it during testing :(

tmiw:
> I just did a quick test and this does fix the issue when a station leaves, but it now/still happens when a station joins. I had not noticed this previously as the 'leaving' bug masked it during testing :(

I think the most recent change will fix this. Confirm?

barjac:
Yes that seems fine now :)

BTW, can we have the git# shown in the version string in the reporter as before, when using pull request branches etc. e.g. FreeDV 2.0.0-dev2#dc5f (I think 4 chars would be adequate).

It is useful to be able to see what other testers are using, as otherwise we can't tell if we are comparing apples with pears!

Tyrbiter:
> BTW, can we have the git# shown in the version string in the reporter as before, when using pull request branches etc. e.g. FreeDV 2.0.0-dev2#dc5f (I think 4 chars would be adequate).

It's nice in a way, but then you'll have people asking what that version is about and asking for one for their OS.

tmiw:
> > BTW, can we have the git# shown in the version string in the reporter as before, when using pull request branches etc. e.g. FreeDV 2.0.0-dev2#dc5f (I think 4 chars would be adequate).
> 
> It's nice in a way, but then you'll have people asking what that version is about and asking for one for their OS.

We were doing it before with longer commit IDs and I didn't hear anything like that from people. I don't expect anything to be different now tbh.

Tyrbiter:
> > > BTW, can we have the git# shown in the version string in the reporter as before, when using pull request branches etc. e.g. FreeDV 2.0.0-dev2#dc5f (I think 4 chars would be adequate).
> > 
> > 
> > It's nice in a way, but then you'll have people asking what that version is about and asking for one for their OS.
> 
> We were doing it before with longer commit IDs and I didn't hear anything like that from people. I don't expect anything to be different now tbh.

OK, let's just hope that it goes unnoticed then ;-)

PR #831, opened 2025-02-16
Update 80/160m default frequencies.

tmiw:
Updates default frequency list as follows:

1. 1.9970 MHz is now 1.8700 MHz.
2. 3.8500 MHz is now 3.8030 MHz.

The above changes are to better handle the Japanese band plan.


Tyrbiter:
I think this works OK, I tried removing .freedv and it created a new file with these frequencies added into the list.

PR #834, opened 2025-02-18
Save Hamlib rig names on startup as they're able to change out from under us.

tmiw:
Hamlib 4.6 added some logic to cause the rig name for FLrig (and possibly) others to change on first usage. For example, with FLrig, the rig name according to FreeDV could end up becoming something like "FLrig IC-7300(FLrig)" internally. As FreeDV uses the rig name and not its ID number to find the rig to use for connecting via Hamlib, the name changing without it knowing causes FreeDV to crash.

This PR saves off the names of the supported Hamlib rigs on startup and uses this list for ID lookups instead, preventing crashes on FreeDV's side.

### DIFF
--- a/src/rig_control/HamlibRigController.cpp
+++ b/src/rig_control/HamlibRigController.cpp
@@ -33,6 +33,7 @@
 #include "util/logging/ulog.h"
 
 HamlibRigController::RigList HamlibRigController::RigList_;
+HamlibRigController::RigNameList HamlibRigController::RigNameList_;
 std::mutex HamlibRigController::RigListMutex_;
 
 #if RIGCAPS_NOT_CONST
@@ -41,6 +42,7 @@ int HamlibRigController::BuildRigList_(struct rig_caps *rig, rig_ptr_t rigList) 
 int HamlibRigController::BuildRigList_(const struct rig_caps *rig, rig_ptr_t rigList) {    
 #endif // RIGCAPS_NOT_CONST
     ((HamlibRigController::RigList *)rigList)->push_back(rig); 
+    RigNameList_.push_back(std::string(rig->mfg_name) + std::string(" ") + std::string(rig->model_name));
     return 1;
 }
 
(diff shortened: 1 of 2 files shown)


Tyrbiter:
Interesting, this commit has returned my TS-890 to sounding mode changes (SSB->SSB-DATA and vice versa) in Morse. This stopped a while ago but I couldn't pin down what caused it.

I view this as a bug fix for a bug I didn't report.

tmiw:
> Interesting, this commit has returned my TS-890 to sounding mode changes (SSB->SSB-DATA and vice versa) in Morse. This stopped a while ago but I couldn't pin down what caused it.
> 
> I view this as a bug fix for a bug I didn't report.

Weird, I'd have expected there to be other problems as well if it was just a naming issue. I was going to say that maybe it's related to going from 4.6.0 to 4.6.2 but not fully sure.

Tyrbiter:
Well, I have hamlib-4.6.2 installed here via rpms and it didn't affect anything until I tried this build of freedv-gui.

I have no clear idea, but maybe it's just down to talking to hamlib slightly differently.

tmiw:
> Well, I have hamlib-4.6.2 installed here via rpms and it didn't affect anything until I tried this build of freedv-gui.
> 
> I have no clear idea, but maybe it's just down to talking to hamlib slightly differently.

/shrug

Anyway, I just pushed a minor change to make absolutely sure that the rig names are in the same order as the main rig list. Let me know how this goes.

Tyrbiter:
This most recent commit has returned things to silence on TS-890 mode changes. Odd, but not really a bug, in some ways it's less annoying and I can see the mode on the radio display.

Perhaps the mode shown in freedv-gui could have -D added so that the Options->Rig Control settings chosen are indicated in the main window?

tmiw:
> Perhaps the mode shown in freedv-gui could have -D added so that the effect of the Options->Rig Control settings chosen are indicated in the main window?

Done. In the meantime, I'm waiting to hear back from the person who originally reported this bug (the first fix didn't take, unfortunately, probably because the rig name list wasn't in the exact same order as the general rig list before).

Tyrbiter:
The latest commit is working just fine here, hope the other person finds it good for them too.

tmiw:
> The latest commit is working just fine here, hope the other person finds it good for them too.

Indeed it is, so I'll go ahead and merge.

PR #835, opened 2025-02-18
Increase short RX coloring timeout to 5 seconds.

### LINKED ISSUE #833: Reporter app no longer shows continuous blue background during solid receive.
### Platform

Linux

### Platform Version

Mageia 10 (dev)

### FreeDV Version

 freedv-gui version: 2.0.0-dev2-34da

### Steps to Reproduce

Monitor your TX using an SDR and watch the RX line in the SDR system reporter.
At the end of a transmission the line goes blue as the transmit station callsign appears.
Previously the blue continued during the over and remained until another station transmitted.
At the start of a second TX the line remains blue until the first update of the RX line at which point the blue goes off. It then stays off or occasionally flashes as for some reason the updates of the RX stations are now occurring at 1 sec intervals, where they used to be every 5 sec.
In a net this causes all the RX station blue backgrounds to flash on and off continuously and randomly, or stay white most of the time, depending on the syncronisation of the timeouts involved.

### Expected Behavior

During a good RX period the blue background of stations receiving a good signal should remain in blue.
The start of a new transmission that passes the filter settings should then extinguish the blue until the new transmission is correctly decoded when it should go blue and remain blue irrespective of updates unless sync is lost.

### Actual Behavior

Video demo:
https://github.com/user-attachments/assets/eb56623f-cfcb-40fa-ac0e-abd61788ddfc

### Additional Comments

_No response_

tmiw:
Resolves #833 by increasing the "short" RX coloring timeout (i.e. when a callsign hasn't been received yet) to 5 seconds from the 2 seconds that it was before.

### DIFF
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -52,7 +52,7 @@ extern FreeDVInterface freedvInterface;
 #endif // defined(WIN32)
 #define RX_ONLY_STATUS "RX Only"
 #define RX_COLORING_LONG_TIMEOUT_SEC (20)
-#define RX_COLORING_SHORT_TIMEOUT_SEC (2)
+#define RX_COLORING_SHORT_TIMEOUT_SEC (5)
 #define MSG_COLORING_TIMEOUT_SEC (5)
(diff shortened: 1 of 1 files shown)


barjac:
Yes that is back to 'normal' much better! :)

Issue #837, opened 2025-02-23
[Software Issue] Virtual Audio Cable (VAC) Parameters


AristarchosOfSamos:
### Platform

Windows

### Platform Version

Win10

### FreeDV Version

e56e

### Steps to Reproduce

Start FreeDV and press PTT

### Expected Behavior

Audio from microphone without glitch on other stations receiving FreeDV audio

### Actual Behavior

Audio from microphone (reported from others) is with occasional glitches even when they report strong signal eg with 20+dB SNR, glitches appear every about 30sec, sometimes more often. the glitches apear by increasing the Resyncs number.

### Additional Comments

Current setup here is using PowerSDR and Virtual Audio Cable (VAC) v4.60 on a SDR1K.
It seems the audio from FreeDV to PowerSDR, the transmitted audio has some issues occasionally.
Receiving cable, from PowerSDR to FreeDV, looks just fine as it give a proper audio without glitches on receive/listening other stations.
Are there any recommended settings for Muzychenko's VAC for the cables used ? Specially for the cable used for Tx?

tmiw:
Are you using Ethernet to connect to the radio, or is the connection over something like Wi-Fi? The latter tends to drop packets fairly often, at least with my Hermes Lite.

Also, do you see this problem with, say, VB-Cable, or just VAC?

AristarchosOfSamos:
Hi, yes, connection to radio is via Ethernet and it does not show any dropped eth packages.
Do not have tested the VB-Cable, only VAC, have it way too many years now ~15 and never had any issues in digimodes.

AristarchosOfSamos:
It is shown that in the Windows build action, there is a stage where tests are conducted with the use of both VB-Cable and VAC.
As described above, I use the VAC for the Radio to FreeDV cables, whereas in the tests VAC is used for Mic/Spk to FreeDV.
Is it possible in your build action tests to switch them so VAC to be used for Radio to FreeDV and see what happens?
Also, can you please let me know the settings used for those cables (both VB-Audio and VAC).

The used VAC settings in my setup are all the same as this

![Image](https://github.com/user-attachments/assets/44b86e7b-8c40-4b77-9068-c0d10bcf63cf)

tmiw:
[...]

PR #840, opened 2025-03-02
Fix dropouts related to virtual audio cables.

tmiw:
This PR is to resolve the issues reported in #837 surrounding signal dropouts and virtual audio cables (especially on Windows). These are commonly used with various SDR radios and seem to have additional considerations on top of the typical USB audio device created when plugging a radio into a PC.

Current changes:

* General:
    * Fix issue reported in #837 where RADE was not properly selected as the default FreeDV mode.
    * Ensure that PortAudio operations are only allowed to take place on one single thread.
    * Reduce maximum sleep time in main TX/RX threads to 10ms (from 20ms).
    * Add additional (optional) RADE stress test to validate issue/solution.
* Windows:
    * Force PortAudio/WASPI to use a higher priority for the audio devices.
    * Fix issue with test script that prevented GitHub from properly failing builds.
 * macOS:
     * Use higher priority for Core Audio.
     * Use VB-Cable for radio devices in GitHub environment to improve GH action reliability.


tmiw:
Investigation note: [PortAudio is apparently not thread safe](https://github.com/PortAudio/portaudio/issues/808). FreeDV isn't trying to make simultaneous calls into the API from multiple threads but it sounds like problems happen when calls are made from multiple threads _at all_. Need to enforce single thread access to `Pa_*` from `PortAudioEngine` / `PortAudioDevice`.

tmiw:
Adding support for exclusive mode (89bd860) seems to have made the problem _worse_ as a lot more of the runs seem to be losing sync vs. the previous commit. That probably lends more credence to the "FreeDV isn't providing samples fast enough" theory, so I reduced the maximum delay in the TX/RX thread down to 10ms (1ba3991) to see what happens.

Tyrbiter:
> @Tyrbiter, @barjac, there's one change that might have impact on pipewire (https://github.com/drowe67/freedv-gui/pull/840/files#diff-caec9b80cc039b3da6710b77ee109324e3f3a94d82309cdea40ffb8f744465a0). The GitHub actions have been consistently passing on Linux with this change but it might be worth testing and confirming as well.

I have built this git ID and have it running, as far as I can tell there is no obvious difference from the https://github.com/drowe67/freedv-gui/pull/836/commits/a5b05b02fc91374263b12dc97feacccc3db80a99 git ID and no unexpected debug output in the console window.

PR #842, opened 2025-03-08
Update build scripts to use optimal number of parallel builds.

tmiw:
Updates build scripts to use optimal parallel number of builds.

### DIFF
--- a/.github/workflows/cmake-windows.yml
+++ b/.github/workflows/cmake-windows.yml
@@ -73,7 +73,7 @@ jobs:
       working-directory: ${{github.workspace}}/build_windows
       run: |
         export PATH=${{github.workspace}}/llvm-mingw-20230320-ucrt-ubuntu-18.04-x86_64/bin:$PATH
-        make -j6 package
+        make -j$(nproc) package
 
     - name: Rename installer
       shell: bash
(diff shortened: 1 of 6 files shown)


Tyrbiter:
I ran this build locally and it completed quite rapidly, to the extent that the running freedv-gui generated a mass of FIFO full messages while the build was running. That doesn't normally happen.

tmiw:
> I ran this build locally after deleting the build_linux sub-directory and it completed quite rapidly, to the extent that the running freedv-gui generated a mass of RX FIFO full messages while the build was running. That doesn't normally happen, but suggests it used up all the available CPUs.

Looks like just specifying `-j` doesn't actually limit the number of parallel builds. We need `-j$(nproc)` (or the equivalent on macOS) to do so. More discussion at https://github.com/memkind/memkind/issues/33.

Tyrbiter:
Just tried the latest, seemed to be pretty fast building after deleting the build_linux directory, and no RX FIFO full messages this time either.

tmiw:
> Just tried the latest, seemed to be pretty fast building after deleting the build_linux directory, and no RX FIFO full messages this time either.

👍 I'll go ahead and merge.

barjac:
A bit late to the discussion but this appears to cut the complete build time in half using my script, so very worthwhile. It's probably much better than that as my timing includes the downloading portions. On my modest machine (i5 4 core) its now around 5 mins.
I have seen no regressions.

Issue #841, opened 2025-03-08
Can FreeDV audio node names be changed to not include spaces?


barjac:
At present all FreeDV audio node names include "FreeDV HF Digital Voice".
Can these please be changed to include "FreeDV_HF_Digital_Voice".

Reason:
No other device IDs or source/sink IDs on my systems include spaces which makes string manipulations in pipewire scripting straightforward.

This seems to be the normal naming convention but I have not seen this in print.

Having to deal with spaces only for FreeDV causes real headaches and much more complication in scripts that have to deal with a mix of styles.

tmiw:
"FreeDV HF Digital Voice" is what's being set as `PA_PROP_APPLICATION_NAME`, which happens here:

```c++
context_ = pa_context_new(mainloopApi_, "FreeDV HF Digital Voice");
```

Unfortunately the example given for application name in the PulseAudio documentation has spaces in it, which implies that they are indeed allowed: https://freedesktop.org/software/pulseaudio/doxygen/proplist_8h.html#ab94d533153744b934780ce627ef3e74c. Are these scripts from other developers that you're using?

barjac:
No these are scripts that I am writing in order to be able to set up pipewire links as I need them (not pulseaudio).
When linking nodes both names are included as parameters in the command line, space separated.
If a name has spaces this cannot work without quoting which complicates the coding especially .

However more recently I have realized that in the case of links to and from FreeDV itself, that these are handled perfectly by FreeDV, so as long as any attempts by my scripts to re-write these fail quietly then it's not a problem.

In brief I am saving a list of all links related to FreeDV and restoring them on start for things like mic loudness compensation and bandpass filtering the audio tones output.

However I still think the spaces should go ;)

PR #843, opened 2025-03-10
Shorten PulseAudio/pipewire app name.

tmiw:
Per request in #841, this PR shortens the app name given to PulseAudio/pipewire such that spaces aren't required.


Tyrbiter:
This works for me and is a useful change.

barjac:
Yes that's good, it has the added advantage of making the node titles in qpwgraph actually readable, whereas previously the real information e.g. (FreeDV to Radio) was cropped off due to the unnecessary text.

Screenshot_20250310_102520

tmiw:
Merged. Thanks all.

PR #847, opened 2025-03-20
Move to using native audio on macOS and Windows

tmiw:
This PR is to investigate moving away from PortAudio for macOS and Windows and towards the native APIs for the respective OSes. This effort started due to continued GitHub Action failures for those platforms on recent PRs; the idea is that interacting with native APIs directly reduces overhead and makes it less likely for dropouts to occur.

Currently done:

1. Implemented basic macOS audio handling with `AVAudioEngine`, a newer API available on 10.15+.
2. Implement other query functionality from macOS API:
    a. Default audio devices (in and out)
    b. Minimum/maximum audio channels available
3. Implement native Wndows WASAPI handling of audio.

To investigate:

2. Perform testing with current macOS implementation to see if GH pass rates improve.
3. Update build system to provide consistent way of specifying PortAudio or native use (i.e. something like `USE_NATIVE_AUDIO=1 ./build_linux.sh` instead of `./build_linux.sh pulseaudio`.


tmiw:
*sigh* Of course Windows builds start succeeding now. It might have been a fluke though so I'll keep investigating.

Tyrbiter:
FYI I tested this branch building on Fedora 42, the build completed normally and ldd shows the expected shared libraries.

I may test this version over the weekend.

Tyrbiter:
Yeah, I wasn't going to rush into using it. More looking to spot any bugs that crept in by accident, lots of changes!

tmiw:
Unfortunately x86 macOS is still pretty unreliable in GH Actions, but the rest seem to be working okay now. Mostly. Maybe we should just disable the former and only test on ARM?

I'd say that maybe it'd improve once we have a full C implementation of RADE but I've seen failures in like 700D before, for instance.

Tyrbiter:
> Unfortunately x86 macOS is still pretty unreliable in GH Actions, but the rest seem to be working okay now. Mostly. Maybe we should just disable the former and only test on ARM?

Is it just the GH that is problematic or does it fall over running on real hardware? I suppose that is going to matter more, but it would be nice if everything worked.

tmiw:
> > Unfortunately x86 macOS is still pretty unreliable in GH Actions, but the rest seem to be working okay now. Mostly. Maybe we should just disable the former and only test on ARM?
> 
> Is it just the GH that is problematic or does it fall over running on real hardware? I suppose that is going to matter more, but it would be nice if everything worked.

GH only AFAICT. Though I should probably use my loopback USB sound card setup again locally just to be sure.

Tyrbiter:
@tmiw Do you think this is now at the point where some Windows testers could try it out for real world function?

tmiw:
> @tmiw Do you think this is now at the point where some Windows testers could try it out for real world function?

Walter's been using it. Haven't heard anything yet about any problems.

tmiw:
I'm going to go ahead and merge since I haven't heard anything negative in a few days. Worst-case, something will come up during the pre-release testing.

PR #849, opened 2025-03-26
Fix crash on FreeDV Reporter connection failure

tmiw:
Followup fix for #846 to handle issue where FreeDV crashes if unable to reach the FreeDV Reporter server. To test:

1. Go to Tools->Options->Reporting and enter an IP address that's not being used by anything else on your local network (i.e. 10.0.1.2).
2. Note in logs that it's trying to connect to 10.0.1.2 (or whatever IP you configured). Wait until the connection times out.
4. Repeat (1) but use `qso.freedv.org` instead. It should stop the pending (re)connection and immediately reconnect to the regular FreeDV Reporter server.

PR #851, opened 2025-03-26
Report "unk" for mode on Hamlib disconnect

### LINKED ISSUE #850: [Software Issue] Radio mode not changed when stopping modem
### Platform

Linux

### Platform Version

Fedora 42 beta

### FreeDV Version

v2.0-dev-3877 (PR #849

### Steps to Reproduce

If using SSB-D for FreeDV the mode indication in the main window is not changed back to the saved mode when the modem is stopped.

Radio in LSB mode, start modem, radio changed to selected freq and LSB-D mode, stop modem, freq returns to original freq but LSB-D mode remains in main window although radio shows LSB mode.

### Expected Behavior

Main window mode display follows actual radio mode at all times.

### Actual Behavior

See above.

### Additional Comments

_No response_

tmiw:
Resolves #850 by adding missed event handler call on Hamlib disconnect, which was supposed to auto-reset the displayed mode back to "unk". This PR also contains some cleanup of duplicated code between OmniRig and Hamlib.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -901,6 +901,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix issue preventing suppression of the Msg tooltip for non-truncated messages. (PR #829)
     * Preserve Hamlib rig names on startup to guard against changes by Hamlib during execution. (PR #834)
     * Fix dropouts related to virtual audio cables. (PR #840)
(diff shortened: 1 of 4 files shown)


Tyrbiter:
This PR seems to work as expected, note that it was hacked into the fix in PR #849 but I couldn't see any code conflicts so it should be fine.

PR #853, opened 2025-03-28
We should be logging getaddrinfo failures as warn, not debug.

tmiw:
Observed on code inspection due to lack of errors in https://github.com/drowe67/freedv-gui/pull/849#issuecomment-2761173385. This PR uses warn instead of debug so we can actually log DNS failures.


Tyrbiter:
Now running this version, will report if I see anything weird happening with DNS.

tmiw:
I think I ran into the original issue. I'm going to merge this into `ms-native-audio` in case it happens again.

tmiw:
Happened again. It doesn't look like it's DNS failing, but I have an idea as to what's going on. I'll try the proposed change and get back to you.

PR #855, opened 2025-03-31
Improve EOO TX/RX reliability

tmiw:
This PR contains the following changes to improve EOO TX/RX reliability:

* *Create FIFOs with correct sizes to reduce EOO time*: While working on #847 I noticed that EOO was taking much longer than it should theoretically take (~900ms vs ~160ms + a few tens of milliseconds for the remaining mic audio to travel through the pipeline). On further investigation I discovered that the FIFOs were being incorrectly created when TX is enabled. This PR resolves the incorrect FIFO creation problem, reducing EOO time down to approximately 250-300ms.
* *Tweak reporting ctests to better match real-world use*: A mock rigctld was added and rigged to stop recording audio when FreeDV requests to go from TX to RX, simulating behavior once the radio stops transmitting.
* *Add support for retrieving audio device latency*: `IAudioDevice` was extended to provide an interface for determining audio device latency (the amount of time for audio to leave the computer once passed to the OS). FreeDV also now pauses for this amount of time at the end of TX before sending the Hamlib/serial PTT/OmniRig request to go back into RX.
* *Add support for retrieving radio latency*: `IRigPttController` was extended to provide an interface for retrieving PTT switching latency. This is calculated ourselves by taking timestamps before and after the actual PTT operation (depending on what PTT method is in use) and calculating the difference between them. Once calculated, this feeds into the audio latency wait above.


Tyrbiter:
Good catch @tmiw , I wonder if this will improve the reliability of EOO text  in general. It will be great to find out once it hits the streets :)

tmiw:
> Good catch @tmiw , I wonder if this will improve the reliability of EOO text in general. It will be great to find out once it hits the streets :)

👍 

BTW this PR isn't causing any RX FIFO warnings for you, right? I remember that was happening last time we touched this area until the FIFO sizes got set pretty high.

Tyrbiter:
> > Good catch @tmiw , I wonder if this will improve the reliability of EOO text in general. It will be great to find out once it hits the streets :)
> 
> 👍
> 
> BTW this PR isn't causing any RX FIFO warnings for you, right? I remember that was happening last time we touched this area until the FIFO sizes got set pretty high.

I have seen nothing like this today with this branch using the newly optimized FIFOs, and in fact so far not even when the freedv-gui build process is eating all the spare CPU. I have now started another brand new latest freedv-gui git version as I note that you needed to change a few FIFO definitions over again :) Very confusing which one is which :-/

What I have noticed over the last few weeks is that some fairly strong signals that should give a good EOO decode do not always do so; I wonder if this change is going to improve this once it reaches the masses. I don't know if the extended EOO Tx timing is the cause of this, or something in the Rx processing. Hard to tell to be honest.

tmiw:
> I have seen nothing like this today with this branch using the newly optimized FIFOs, and in fact so far not even when the freedv-gui build process is eating all the spare CPU. I have now started another brand new latest freedv-gui git version as I note that you needed to change a few FIFO definitions over again :) Very confusing which one is which :-/

Before we were just doing `make -j` without capping to the number of cores on a user's system, which is what caused RX FIFO warnings last time for you IIRC. The warnings disappeared as soon as that got changed to `make -j$(nproc)`.

> What I have noticed over the last few weeks is that some fairly strong signals that should give a good EOO decode do not always do so; I wonder if this change is going to improve this once it reaches the masses (special reference for @drowe67 :-) ). I don't know if the extended EOO Tx timing is the cause of this, or something in the Rx processing. Hard to tell to be honest.

It's possible that there are subtle timing differences causing EOO to be truncated before the radio can receive the EOO audio from FreeDV and actually transmit it. Currently once the TX FIFO empties, we wait for any configured RX/TX delay (0ms by default) and then tell Hamlib/serial PTT/OmniRig to put the radio back into RX. Unfortunately I don't think there's an easy way to determine how much extra wait after FIFO empty we should have, or whether any wait should be added at all.

Tyrbiter:
> Before we were just doing `make -j` without capping to the number of cores on a user's system, which is what caused RX FIFO warnings last time for you IIRC. The warnings disappeared as soon as that got changed to `make -j$(nproc)`.

> It's possible that there are subtle timing differences causing EOO to be truncated before the radio can receive the EOO audio from FreeDV and actually transmit it. Currently once the TX FIFO empties, we wait for any configured RX/TX delay (0ms by default) and then tell Hamlib/serial PTT/OmniRig to put the radio back into RX. Unfortunately I don't think there's an easy way to determine how much extra wait after FIFO empty we should have, or whether any wait should be added at all.

Mmm, so maybe we should consider adding in a default delay of 20ms or so and monitor in an empirical way if it helps. Perhaps a 2 stage test where we go with the new code and if there are still noticeable EOO misses then add the delay.

tmiw:
> Mmm, so maybe we should consider adding in a default delay of 20ms or so and monitor in an empirical way if it helps. Perhaps a 2 stage test where we go with the new code and if there are still noticeable EOO misses then add the delay.

I think I have a way to simulate behavior when using hamlib (i.e. by cutting off the TX audio as soon as FreeDV puts the "radio" back into receive). With this change, I'm getting consist RADE reporting test failures now locally. I'll investigate more to see if there's a way to calculate the optimal amount of wait time before going from TX to RX.

tmiw:
> Mmm, so maybe we should consider adding in a default delay of 20ms or so and monitor in an empirical way if it helps. Perhaps a 2 stage test where we go with the new code and if there are still noticeable EOO misses then add the delay.

I updated the tests to add a mock rigctld and enable its use in FreeDV. The rigctld is set up to kill the audio recording as soon as it's told to stop transmitting. Based on this, I had to add 80ms of delay before GitHub would pass the tests, but I suspect that it could be significantly more based on unknown factors. (For example, I had to bump it up to 160ms locally for my laptop to pass the tests.)

OTOH, I don't think ending TX is instant even on SDRs, so it's possible that tweaking the mock rigctld to more closely match the real world could allow us to drop the delay down a bit. Any idea how long it usually takes a radio to go from TX->RX? Based on typical relay off times I suspect this is in the range of 20ms or so but I could be wrong.

Tyrbiter:
Tx/Rx time tends to be a bit of a variable feast, so many possible things affect it. I expect the minimum is about 10ms, I know that for CW an RF rise time of faster than 6ms leads to spectral spreading. At the other end of the range I imagine some radios could take 40-50ms especially if they have sequenced changeover and maybe an external PA.

As for pipeline delays in software or SDR hardware again hard to be sure, but you have found an extra 80ms from somewhere and that may be typical. None of the normal OSs have real time support as such, because that is usually defined as reacting to an interrupt in the smallest processor time slice which is often 10ms. To get really tight timing requires a hardware timer of some kind.

I hope that word soup helped :)

tmiw:
> Tx/Rx time tends to be a bit of a variable feast, so many possible things affect it. I expect the minimum is about 10ms, I know that for CW an RF rise time of faster than 6ms leads to spectral spreading. At the other end of the range I imagine some radios could take 40-50ms especially if they have sequenced changeover and maybe an external PA.

I decided to go with 20ms for now as I suspect most radios in current use use physical relays rather than e.g. PIN diodes or transistors for TX/RX control. With this, 80ms causes the tests to pass consistently on my MacBook Pro.

> As for pipeline delays in software or SDR hardware again hard to be sure, but you have found an extra 80ms from somewhere and that may be typical. None of the normal OSs have real time support as such, because that is usually defined as reacting to an interrupt in the smallest processor time slice which is often 10ms. To get really tight timing requires a hardware timer of some kind.

I found [this](https://stackoverflow.com/questions/65600996/avaudioengine-reconcile-sync-input-output-timestamps-on-macos-ios) which along with some diagnostic tools Apple provides makes me think that the delay time should be at least the sum of input latency + output latency. Whether this is sufficient for Linux and Windows is TBD, though; maybe we can do at least some testing OTA with the former and see if the current 80ms delay helps?

> I hope that word soup helped :)

Yep. 👍

tmiw:
Interestingly, I did some tests on Linux and it looks like a delay isn't needed at all, either in the app itself or in the mock rigctld. This is with virtual audio devices, though, so we may have significantly different results with actual hardware.

tmiw:
I'm not sure if we need to add both the TX input (i.e. microphone) and output device latencies together, but doing so at least gets me a total delay time in the ballpark of 80ms on my laptop. That said, I would think that only output latency should matter, so maybe something is still wrong with the test setup. 

Anyway, there should be logic for pipewire/PulseAudio too for grabbing this, so at least we can see if that improves callsign decoding at all. 👍

tmiw:
> Now testing this latest version, so far it seems to be entirely normal, no strange debug output.

How much latency does it report when you stop transmitting, out of curiosity? The latest should only be using TX out latency now and IIRC it's only something like 10-20 ms on my laptop. Weirdly that causes a test failure locally but apparently (mostly) works fine on GH.

Tyrbiter:
With git b41a:

18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:859: Waiting for EOO to be queued
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/TxRxThread.cpp:650: Triggering sending of EOO
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:180: Queueing 1632 EOO samples to output FIFO
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1472)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1312)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1152)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 992)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 832)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 672)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 512)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 352)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 192)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 32)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 32 EOO samples (remaining in FIFO: 0)
18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:865: Detected that EOO has been enqueued
18:43:13 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:881: All TX finished (diff = 305 ms, fifo_empty = 23295, sample = 23294), going out of PTT
18:43:13 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:894: Pausing for a minimum of 18770 microseconds before TX->RX to allow remaining audio to go out

tmiw:
> With git b41a:
> 
> 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:859: Waiting for EOO to be queued 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/TxRxThread.cpp:650: Triggering sending of EOO 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:180: Queueing 1632 EOO samples to output FIFO 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1472) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1312) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 1152) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 992) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 832) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 672) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 512) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 352) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 192) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 160 EOO samples (remaining in FIFO: 32) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/pipeline/RADETransmitStep.cpp:97: Returning 32 EOO samples (remaining in FIFO: 0) 18:43:12 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:865: Detected that EOO has been enqueued 18:43:13 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:881: All TX finished (diff = 305 ms, fifo_empty = 23295, sample = 23294), going out of PTT 18:43:13 INFO /home/bdm/git/freedv-gui-v2.0-ms-eoo-optimization/src/ongui.cpp:894: Pausing for a minimum of 18770 microseconds before TX->RX to allow remaining audio to go out

Thanks! Yeah, it seems that 10-20ms is typical for audio devices. I wonder if there's something else going on that's causing issues on my laptop. 

(BTW VB-Cable is returning the same latency figure on my laptop and the GH environment.)

Tyrbiter:
I suspect that it's going to need a majority of on-air users to have a new dev version before we get a good handle on this.

I was watching what happened with a german station who was at about SNR 18-20dB and with a bit of SSB interference on frequency and some more HF a couple of kHz no decode occurred when dropping Tx, so whatever is happening now it probably needs the new code to be in widespread use.

tmiw:
> I suspect that it's going to need a majority of on-air users to have a new dev version before we get a good handle on this.
> 
> I was watching what happened with a german station who was at about SNR 18-20dB and with a bit of SSB interference on frequency and some more HF a couple of kHz no decode occurred when dropping Tx, so whatever is happening now it probably needs the new code to be in widespread use.

Agreed, since these changes primarily impact TX and not RX.

Also, I did some testing with actual hardware (both an IC-705 and a QMX plugged in via USB and using different instances of FreeDV). On my laptop, I got latencies of 13687us (IC-705) and 13187us (QMX) and FreeDV was able to consistently decode my callsign regardless of which radio was transmitting. Still need to figure out why the ctest is failing locally but passing on GH, but additionally, we probably need to rig something up to be able to run this test on Windows.

tmiw:
The good news is that the new Windows test I added today seems to pass in the GH environment. The bad news is that the returned latency figures don't seem quite right. Hopefully this latest commit does the trick.

tmiw:
I passed the latest Windows and macOS builds that GitHub generated for this PR to Walter and gang to give a try. If the feedback is good, we can probably merge.

PR #858, opened 2025-04-12
Update build dependency for Linux workflow

dforsi:
This change makes the build succeed with Ubuntu 24.04 which is currently used by the workflow.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -24,7 +24,7 @@ jobs:
       shell: bash
       run: |
            sudo apt-get update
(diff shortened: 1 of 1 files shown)


tmiw:
Thanks for this PR. Most of this is already in the v2.0-dev branch. I'll take a closer look to make sure before merging, though.

tmiw:
I'm going to go ahead and merge this to `master` instead of `v2.0-dev`. Reason being that I don't necessarily want to clobber your `master` with a merge from `v2.0-dev` beforehand. I suspect a merge down from the project's `master` to `v2.0-dev` after this merge will basically be a no-op but that's fine.

PR #862, opened 2025-04-16
sioclient-disabled build (cmake -DUSE_SIOCLIENT=0)

jg1uaa:
Debian has [optional-sio-client](https://sources.debian.org/src/freedv/1.8.11-1/debian/patches/optional-sio-client/) patch, to build without sioclient library.

Debian has this patch due to sioclient is not packaged, but it cannot apply to current build.
Here is an alternative.

And, this might be useful for building-sioclient-is-difficult environments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,20 +466,23 @@ endif(HAMLIB_LIBRARY AND HAMLIB_INCLUDE_DIR)
 
 
 # sioclient library
-message(STATUS "Looking for sioclient...")
-find_path(SIOCLIENT_INCLUDE_DIR sio_client.h)
-find_library(SIOCLIENT_LIBRARY sioclient)
-message(STATUS "  Socket.io library: ${SIOCLIENT_LIBRARY}")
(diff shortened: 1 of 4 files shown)


hobbes1069:
This would be welcomed in Fedora as well. I have been unable to update FreeDV in Fedora due to this. I actually have a package review in progress but it's been stalled because FreeDV needs a git checkout and not a released version.

Tyrbiter:
This PR actually removes the need for sioclient-cpp in freedv-gui

https://github.com/drowe67/freedv-gui/pull/836

It was merged a while ago.

Perhaps @hobbes1069 and @tmiw can discuss to ensure that the as yet unreleased freedv version 2.x will be good in Fedora, I am not sure what can be done about the freedv-1.9.x code at this point.

tmiw:
There are some non-RADE related bugfixes and feature enhancements in the current `v2.0-dev` branch, including the sioclient-cpp removal as @Tyrbiter mentioned. We may want to discuss whether some of those should be backported to 1.9.x (especially if we want to go to RADE-only for the final 2.0.0 release as 1.9.x would be the last version to support the previous modes).

PR #866, opened 2025-04-18
Request real-time priority from the operating system for audio pipeline

tmiw:
This PR requests real-time priority for audio pipeline threads to ensure that they execute consistently enough to reduce/prevent dropouts (especially in the GH environment). If so, this will also investigate the feasibility of reducing the audio latency from 40-50ms on macOS/Windows to closer to 10ms (as is standard in testing with pipewire).

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -29,7 +29,7 @@ jobs:
             
     - name: Spellcheck codebase
       shell: bash
-      run: codespell --ignore-words-list=caf,radae,rade,inout,nin,ontop,parm,tthe,ue `find src -name '*.c*' -o -name '*.h' -o -name '*.mm' | grep -v 3rdparty`
+      run: codespell --ignore-words-list=caf,radae,rade,inout,nin,ontop,parm,tthe,ue `find src -name '*.c*' -o -name '*.h' | grep -v 3rdparty`
 
     - name: Install Python required modules
       shell: bash
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -25,7 +25,7 @@ jobs:
       run: |
            sudo apt-get update
            sudo apt-get upgrade -y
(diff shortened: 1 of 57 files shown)


barjac:
In Mageia we have a realtime group which was introduced for jack/ardour IIRC many years ago.
I have not looked at this recently and recall little about it, but would users benefit from belonging to this group or is it handled by freedv?

tmiw:
> In Mageia we have a realtime group which was introduced for jack/ardour IIRC many years ago. I have not looked at this recently and recall little about it, but would users benefit from belonging to this group or is it handled by freedv?

I was going to look into using `rtkit` to request real-time. Though tbh I'm not sure we need to set the pipeline threads to real time on Linux yet.

Tyrbiter:
There was a fairly recent change in the kernel to improve RT activity, https://tuxcare.com/blog/linux-kernel-6-12/ is one source of information.

I don't know exactly where and how to request the RT capability for a given process.

tmiw:
> There was a fairly recent change in the Linux kernel to improve RT activity, https://tuxcare.com/blog/linux-kernel-6-12/ is one source of information.
> 
> I don't know exactly where and how to request the RT capability for a given process.
> 
> Fedora usually has a bleeding edge kernel, so it should work for distros that keep up to date.

My understanding was that without something like `rtkit`, the app would need root or otherwise have permissions to make itself real-time--no matter what kernel was being used.

Tyrbiter:
> > There was a fairly recent change in the Linux kernel to improve RT activity, https://tuxcare.com/blog/linux-kernel-6-12/ is one source of information.
> > I don't know exactly where and how to request the RT capability for a given process.
> > Fedora usually has a bleeding edge kernel, so it should work for distros that keep up to date.
> 
> My understanding was that without something like `rtkit`, the app would need root or otherwise have permissions to make itself real-time--no matter what kernel was being used.

It sounds like something you can find out about @tmiw as I don't know much. The PREEMPT_RT feature is documented I suppose, but it's not something I really understand :)

tmiw:
I managed to get rtkit working on Linux; "FreeDV rxThread" now shows up as priority of -4 when using `top -H`. :)

tmiw:
Merging so we can get more runtime with a wider group of people. It looks like latency has gone down quite a bit in some cases and we might have improved the pass rate on GH (except for macOS x86_64, but maybe we can investigate more in a future PR).

PR #870, opened 2025-04-27
Add octave as a dependency for macOS build

peterbmarks:
When trying to build on macOS it failed due to octave not being installed on my system. It was missing from the dependencies recommended in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ This document describes how to build the FreeDV GUI program for various operatin
   ```
   $ sudo apt install libspeexdsp-dev libsamplerate0-dev sox git \
   libwxgtk3.2-dev libhamlib-dev libasound2-dev libao-dev \
-  libgsm1-dev libsndfile1-dev cmake module-assistant build-essential
+  libgsm1-dev libsndfile1-dev cmake module-assistant build-essential \
+  octave octave-signal
   $ git clone https://github.com/drowe67/freedv-gui.git
   $ cd freedv-gui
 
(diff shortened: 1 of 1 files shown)


tmiw:
Thanks for that. Just to be safe, I also added Octave to the respective Linux instructions.

tmiw:
> > Thanks for that. Just to be safe, I also added Octave to the respective Linux instructions.
> 
> I just saw this and am somewhat concerned.
> 
> In Mageia 10 (dev) (Linux) My main system which is already running and building v2.0-dev without issue, does not have any octave* system packages installed.
> 
> Also octave-signal does not exist in our repos.
> 
> A regular install of octave-devel would use 3GB of disc space by installing a further 119 packages.
> 
> If only the 'octave' package is installed with recommended packages excluded then this drops to 90MB and 19 packages.
> 
> I don't see any attempt to restrict 'recommends' in the above, but I know little about other distro's packaging.
> 
> Is this new dep in preparation for removal of a chunk of python? If so I will need to package octave-signal. :(

It's possible that octave is just a test dependency, though I'm not sure why macOS would have trouble compiling without it if tests are disabled (which IIRC are by default). If the latest `v2.0-dev` is still able to compile for you with tests off I can go ahead and remove from the README.

barjac:
I don't do anything to enable tests in my script and I don't call for octave as a dependency in any distro so for now I don't see a need for it.
I have built the current 2.0.0-dev2-7b7e3b several times today without it :)

peterbmarks:
Sounds like the macOS build script is somehow requiring octave. I guess it’s inadvertently enabling the test suite.

As you say it would be better to be able to build for users without it.

Peter

tmiw:
See https://github.com/drowe67/freedv-gui/pull/874 for updates.

BTW it looks like I have GitHub set to not install Octave when building the universal binary (and that obviously works), so I'm actually not sure why it was needed for you.

peterbmarks:
I'm really sorry about this. I've just gone to a fresh machine, and successfully built without octave. I can't explain what went wrong before. Again, sorry to waste your time.

PR #861, opened 2025-04-14
Enable color emojis in the FreeDV Reporter window

tmiw:
WIP PR to evaluate the use of `wxDataViewCtrl` to help enable the feature request in #827. Only basic display functionality is working, other previous window behavior is not currently working.

PR #871, opened 2025-04-28
macOS: Better handle when audio devices go away

tmiw:
This PR improves FreeDV behavior on macOS when device configuration changes (i.e. active devices go away). We basically need to do the following:

* For devices that go away:
    * Determine whether device was previously a "default" device. If so, automatically change the device ID to the new default device and restart (note: this does not change the user's configuration; they will receive a "device not found" type error on next start as normal).
    * If device was not a default one: show error message to the user so they can stop/start and/or check configuration.
* Restart all other active devices as well as AVAudioEngine will shut down audio to some (?) devices if there's a configuration change.

Additional fixes in this PR:

* Fix deadlock issue in FreeDV Reporter GUI logic introduced in PR #861.
* Improve logging of latency by listing out audio device latency vs. CAT response time.
* (Experimental, TBD) Disable the Python garbage collector (!) in favor of solely reference counting.
* Fix issue in ThreadedObject that can cause the application to hang on stop.
* Fix rendering issue in FreeDV Reporter GUI causing Last Update column to grow larger than contents.
* Fix crash when tracking band/frequency in FreeDV Reporter.
* Fix deadlocks and crashes in Audio Options window.
* Also sort by connect time in FreeDV Reporter to ensure sorting remains stable.
* Defer deselection of rows to after filter changes complete to avoid inadvertent selection.
* Fix issue causing PTT button to appear still active while stopping TX.
* Fix crash while accessing audio latency on stop.
* Defer sending hide command to FreeDV Reporter until connection fully established.
* Fix rendering issue causing random spikes in From Mic plot.


Tyrbiter:
I see that changing the reporter window to show only stations in one band always leads to a row being highlighted, maybe this could be prevented?

tmiw:
> I see that changing the reporter window to show only stations in one band always leads to a row being highlighted, maybe this could be prevented?

I just tried both the tracking feature and the Band drop-down and it doesn't look like either selects any rows here. There's also code in both handlers to deselect rows, so what I'm seeing is likely not just by coincidence. If you can provide exact steps I can try to duplicate what you're seeing that way.

Tyrbiter:
Open reporter window, currently band select is 'All'. Select a band, any band with stations on it seems to do the same thing, one of the stations will be highlighted, not always the top one.

I have noticed that once I have used the band select and cleared the highlighted row, then further band selections or returning to 'All' does not have a highlighted row in all cases, but repeatedly changing bands and then clearing any highlight that appears then often clears any highlight for a cycle. This is not deterministic, it seems semi-random.

Sorry, it's really hard to explain this.

tmiw:
It was really intermittent here but I think I saw what you were talking about. Hopefully the change I just pushed helps?

Tyrbiter:
> It was really intermittent here but I think I saw what you were talking about. Hopefully the change I just pushed helps?

A very quick check shows that this appears to completely prevent the highlighting happening. I will keep an eye on it and report further if needed.

tmiw:
I'm going ahead and merging this so that the PR doesn't grow out of control. There are a couple more issues that likely need further investigating so those will be separate PRs.

Issue #872, opened 2025-05-05
Gtk crashes


g4dya:
### Platform

Linux

### Platform Version

Linux Mint 22.1

### FreeDV Version

v2.0-dev #7b7e3b

### Steps to Reproduce

Linux Mint 22.1 MATE
Kernel: 6.8.0-59-generic #61-Ubuntu SMP PREEMPT_DYNAMIC
libwxgtk 3.2.4

Run the app and after a random period of time it will crash with the message below.

I've seen this bug on recent commits going back to at least April 25 (#d13a718):

ASSERT INFO:
./include/wx/dynarray.h(196): assert ""n != (-1)"" failed in Remove(): removing inexistent element in wxArray::Remove

BACKTRACE:
[1] wxDataViewModel::ItemDeleted(wxDataViewItem const&, wxDataViewItem const&)
[2] wxEvtHandler::TryHereOnly(wxEvent&)
[3] wxEvtHandler::ProcessEventLocally(wxEvent&)
[4] wxEvtHandler::ProcessEvent(wxEvent&)
[5] wxEvtHandler::ProcessPendingEvents()
[6] wxApp::DoIdle()
[7] g_main_loop_run
[8] gtk_main
[9] wxGUIEventLoop::DoRun()
[10] wxEventLoopBase::Run()
[11] wxAppConsoleBase::MainLoop()
[12] wxEntry(int&, wchar_t**)
[13] main
[14] __libc_start_main

### Expected Behavior

Not crashing

### Actual Behavior

Crashing as above

tmiw:
Does https://github.com/drowe67/freedv-gui/pull/871 work any better for you?

g4dya:
Unfortunately not. ( #1b7d1c8f ). Getting exactly the same error.

BTW switching to a different workspace in the workspace switcher seems to provoke the bug.

tmiw:
I believe I was able to duplicate @g4dya's configuration in a VM. Unfortunately workspace switching hasn't duplicated the problem so far. I'll keep it running and see if the original reported issue happens.

g4dya:
@Tyrbiter - no, I'm afraid there's no newer libwxgtk package for my distro AFAICS.

@tmiw - many thanks for attempting to reproduce bug in a VM.

The problem appears pretty randomly. The program may crash on startup, or it may run for an hour or more.

When I get time (which won't be until next week) I will try to triage the exact commit where the problem started.

tmiw:
[...]

PR #873, opened 2025-05-07
Fix additional GTK assertion on startup for FreeDV Reporter.

tmiw:
WIP - investigating FreeDV Reporter related crash reported in #872.


barjac:
I have not seen this issue but this branch is running OK without regressions for me.

Tyrbiter:
Seems perfectly fine with the very latest commit here.

PR #874, opened 2025-05-08
Remove octave/octave-signal from README.

tmiw:
Turns out that Octave isn't actually needed for a build without tests enabled. This PR removes Octave from the README to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,7 @@ This document describes how to build the FreeDV GUI program for various operatin
   ```
   $ sudo apt install libspeexdsp-dev libsamplerate0-dev sox git \
   libwxgtk3.2-dev libhamlib-dev libasound2-dev libao-dev \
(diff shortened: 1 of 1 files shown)


tmiw:
Additionally, I  apparently don't have Octave installed on my MacBook Pro and can still build, so I propose completely backing out #870. Sound reasonable @barjac, @peterbmarks?

barjac:
Yes that sounds fine.

tmiw:
No worries! Always good to have people double checking stuff :)

barjac:
Maybe if you installed your system libsndfile-devel package rather than letting the script try to build it, you may not hit this issue?

For all Linux builds my script [1] does this, however I know nothing about Mac ;)

https://github.com/barjac/freedv-rade-build/blob/main/freedv-rade-build

e.g. line 60

PR #745, opened 2024-09-02
Set timeout for Hamlib comms to avoid GUI getting stuck.

tmiw:
Resolves #742 by setting a timeout when connecting to the radio via Hamlib. Note that this requires hamlib 4.6~git to have full effect (especially with Icom radios).

PR #875, opened 2025-05-08
Hamlib: improve behavior with Icom rigs, serial port PTT

tmiw:
~On some radios (e.g. Flex 6000/8000 series), attempting to set the timeout in Hamlib causes `rig_open()` to fail. This PR causes FreeDV to reattempt connection without setting timeouts, ensuring that rig control can still be used.~

This PR performs the following:

1. Reverts timeout changes from #745.
2. Adapts various config requests from WSJT-X to (a) improve Icom behavior and (b) improve behavior when using serial port PTT.
3. Adds support in Hamlib for CAT PTT via the Data port (instead of Mic)--needed for some older radios.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -140,7 +140,7 @@ your radio:
 + Hamlib: support for many different radios via the Hamlib library and a serial port (or via TCP/IP for some devices, e.g. SDRs or FLrig/rigctld).
 + Serial Port: direct access to the serial port pins
 
-You may also optionally configure a second serial port for PTT input.
+You may also optionally configure an additional serial port for PTT input.
 This can be useful for interfacing devices like foot switches to 
 FreeDV. If configured, FreeDV will switch into transmit mode (including
 sending the needed Hamlib or serial commands to initiate PTT) when it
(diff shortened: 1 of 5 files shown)


barjac:
This has greatly improved the situation where the radio may inadvertently be off when starting the modem.
No error dialogues or having to stop the modem, just turn on radio and everything just works!
Turning off the radio and back on does not cause issues either now - all just seamless :)

Using rigctld from system hamlib package (hamlib-4.6.2) for PTT and CAT.

Nice - thanks!

tmiw:
> This has greatly improved the situation where the radio may inadvertently be off when starting the modem. No error dialogues or having to stop the modem, just turn on radio and everything just works! Turning off the radio and back on does not cause issues either now - all just seamless :)
> 
> Using rigctld from system hamlib package (hamlib-4.6.2) for PTT and CAT.
> 
> Nice - thanks!

Actually, I was going to revert because this didn't fix the original issue I was having. What do you think about just increasing the timeout instead? It's currently at 1 second but maybe something like 5 would be better?

barjac:
Oh! I really like this as it is, as it also avoids the issue where the Stop button refuses to do anything due to the hamlib error dialogue grabbing the focus.
Sometimes after closing that dialogue the rig refuses to respond and only a reboot will sort out the USB connection to the rig. Restarting rigctld does not help. All of those issues seem fixed by this PR.

tmiw:
FWIW disabling the timeout logic with hamlib 4.6.2 does eventually cause an error to appear with my IC-705 when I use the GPS port instead of the CI-V one (the scenario in the original GH issue). However, it took quite a while for the error window to pop up (basically something like 5-10 seconds when toggling PTT or trying to stop FreeDV, not on start), so I'm not fully sure I want to go away from setting any timeouts at all.

OTOH, the issues you're running into aren't great either, especially needing to reboot sometimes. Not sure why that's happening TBH. I'll have to think about this more.

tmiw:
I decided to remove the timeout entirely as (a) hamlib 4.6.2 seems to handle the original bug scenario better now and (b) WSJT-X doesn't set it. I also took the opportunity to add some serial PTT related config options that WSJT-X seems to do; I'll do a quick sanity check of Hamlib CAT + serial PTT and will merge if that works okay.

PR #876, opened 2025-05-09
Further audio dropout improvements

tmiw:
This PR implements various improvements to the audio chain to allow further reductions in dropouts:

1. 10ms latency is now requested from the OS on all supported platforms.
2. Further cleanup to remove locking and dynamic allocations (especially in `ParallelStep` and the audio filters).
3. FIFO sizes are increased for legacy FreeDV modes (700D/E/1600).
4. Audio record and playback has been moved to threads that don't execute in RT context.
5. Ensure libsamplerate is compiled with optimizations on macOS and Windows (to reduce CPU usage and improve GH Action pass rate).

Additionally, a minor bug was fixed that caused the waterfall to clear itself on every mode change.


tmiw:
Going to merge this PR as it's gotten pretty big. We'll watch the GH Action pass rate (as well as feedback from testers) to see if this at least kept things the same, if not hopefully gotten better.

PR #878, opened 2025-05-10
FreeDV Reporter: Force re-sorting on user connect.

### LINKED ISSUE #877: Reporter sort order lost if a station leaves and returns
Background info:
During UK FreeDV nets we use a digit as the first character of the Msg field of the reporter and sort on the Msg columm.

This allows everyone to keep track of the sequence which works well.

Bug:
If a station leaves the net temporarily and then returns he appears at the top of the list and the sort order is ignored, until everyone in the group clicks the header of the Msg column to reset the sort again.

Here the transmitting station had left briefly and then returned:

Image

This was also noticed by G4DYA who was using a different branch of freedv at the time, my (G4MKT) build and his can be seen in the screen shot. The GW stations are using Windows.

tmiw:
Resolves #877 by forcing re-sorting of the FreeDV Reporter list on every timer fire.

### DIFF
--- a/src/config/FreeDVConfiguration.cpp
+++ b/src/config/FreeDVConfiguration.cpp
@@ -50,7 +50,7 @@ FreeDVConfiguration::FreeDVConfiguration()
     , reporterWindowWidth("/Windows/FreeDVReporter/width", -1)
     , reporterWindowHeight("/Windows/FreeDVReporter/height", -1)
     , reporterWindowVisible("/Windows/FreeDVReporter/visible", false)
-    , reporterWindowCurrentSort("/Windows/FreeDVReporter/currentSort", -1)
+    , reporterWindowCurrentSort("/Windows/FreeDVReporter/currentSort", 5) /* FREQUENCY_COL */
     , reporterWindowCurrentSortDirection("/Windows/FreeDVReporter/currentSortDirection", true)
         
     /* Current tab view */
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -792,7 +792,7 @@ void FreeDVReporterDialog::OnTimer(wxTimerEvent& event)
     FreeDVReporterDataModel* model = (FreeDVReporterDataModel*)spotsDataModel_.get();
     model->updateHighlights();
 
-    if (sortRequired_)
+    //if (sortRequired_)
     {
         model->Resort();
         sortRequired_ = false;
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -1852,6 +1852,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::refreshAllRows()
             if (newVisibility)
             {
                 ItemAdded(wxDataViewItem(nullptr), wxDataViewItem(kvp.second));
+                parent_->sortRequired_ = true;
             }
             else
             {
@@ -2014,6 +2015,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::onUserConnectFn_(std::string
         if (temp->isVisible)
         {
             ItemAdded(wxDataViewItem(nullptr), wxDataViewItem(temp));
+            parent_->sortRequired_ = true;
         }
     });
 
@@ -2099,6 +2101,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::onFrequencyChangeFn_(std::st
                 if (newVisibility)
                 {
                     ItemAdded(wxDataViewItem(nullptr), dvi);
+                    parent_->sortRequired_ = true;
                 }
                 else
                 {
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -59,7 +59,6 @@ using namespace std::placeholders;
 FreeDVReporterDialog::FreeDVReporterDialog(wxWindow* parent, wxWindowID id, const wxString& title, const wxPoint& pos, const wxSize& size, long style) 
     : wxFrame(parent, id, title, pos, size, style)
     , tipWindow_(nullptr)
-    , sortRequired_(false)
 {
     if (wxGetApp().customConfigFileName != "")
     {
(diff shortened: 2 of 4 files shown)


barjac:
Thanks! That seems to work OK.
Not sure what is actually triggering the re-sort as in a quick test on 80m just now with no tracking filter set, a 'Z' at the start of the Msg line of a joining station appeared at the top of other text (including a 'A') but below blanks.
It then, after several seconds moved to the bottom to its proper place.
I would not imagine that the re-sort would be that slow if it was triggered when the station joined. Does it wait for an update of that line?

Not an issue just curious ;)

I will get a better test in the News net in the morning.

tmiw:
> Thanks! That seems to work OK. Not sure what is actually triggering the re-sort as in a quick test on 80m just now with no tracking filter set, a 'Z' at the start of the Msg field of a joining station appeared at the top of other text (including a 'A') but below blanks. It then, after several seconds moved to the bottom to its proper place. I would not imagine that the re-sort would be that slow if it was triggered when the station connected. Does it then wait for an update of that row?
> 
> Not an issue just curious ;)
> 
> I will get a better test in the News net in the morning.

I think the re-sort is delayed until the highlights all update, which happens every 100ms IIRC. I wouldn't expect it to take much longer than that but maybe there's something in wxWidgets that delays it further.

barjac:
OK I have just been monitoring with no filter set and find that I can see no way to disable the sort order.
The sort on each column used to be tri-state (up, down and none) so that sorting could be stopped. Now and for a few builds recently it has become bi-state.
If the sorting cannot be turned off this way it may be annoying for someone trying to watch the reporter unfiltered.

tmiw:
> OK I have just been monitoring with no filter set and find that I can see no way to disable the sort order. The sort on each column used to be tri-state (up, down and none) so that sorting could be stopped. Now and for a few builds recently it has become bi-state. If the sorting cannot be turned off this way it may be annoying for someone trying to watch the reporter unfiltered, as stations come and go.

Yeah, I think this is a property of the control that's now being used. Maybe we can just default to distance or something else that's sensible?

barjac:
OK if nothing can be done then probably frequency is a reasonable default for initial config.

Regarding this PR it is working but occasionally fails to move a newly arrived station.

In these shots taken this morning the sort is incorrect after stations arrived/returned:

Screenshot_20250511_092343

Screenshot_20250511_092530

Screenshot_20250511_092737

Clicking the Msg header twice corrected these in all cases.
Flags sort after text and the Welsh flag sorts after the English (George Cross :)
It worked correctly more than not, but the fail rate was quite high :(

tmiw:
Does the latest help at all?

barjac:
> Does the latest help at all?

Only a quick test so far but looks good - will check more in net tomorrow afternoon :)

barjac:
#5bf3
Mousing over any row causes the row background to flicker from white to pale blue as the mouse is moved.

Tyrbiter:
I've just noticed a massive load of repeating debug output, the repeating unit appears to be this:

(freedv:1226389): Gtk-CRITICAL **: 18:12:41.344: ../gtk/gtktreeview.c:6712 (validate_visible_area): assertion `has_next' failed.
There is a disparity between the internal view of the GtkTreeView,
and the GtkTreeModel.  This generally means that the model has changed
without letting the view know.  Any display from now on is likely to
be incorrect.


(freedv:1226389): Gtk-CRITICAL **: 18:12:41.409: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed

(freedv:1226389): Gtk-CRITICAL **: 18:12:41.509: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed

(freedv:1226389): Gtk-CRITICAL **: 18:12:41.609: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed

(freedv:1226389): Gtk-CRITICAL **: 18:12:41.710: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed

(freedv:1226389): Gtk-CRITICAL **: 18:12:41.810: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed

I'm running https://github.com/drowe67/freedv-gui/pull/878/commits/5bf38c878ec7eb2bf6a708a575f3b39c48c5b613 and up to today it had been running for at least 48 hours without any obvious problems.

It didn't seem to affect anything other than the debug, and at the time I saw it I didn't have the reporter window open.

barjac:
I have been testing for the 'flashing background under cursor' issue from different Linux systems and it only seems to happen in systems that are very current as regards package versions.

It does not happen in:
Mageia 9
Linux Mint 22.1 (thanks to G4DYA for testing)

It does happen in:
Mageia10 (dev) (currently 'bleeding edge')
OpenSUSE Tumbleweed (rolling release)

So this appears as though an updated lib/kernel may be to blame.

barjac:
If the mouse is left stationary the flicker happens every time a row updates.

https://github.com/user-attachments/assets/dae051a1-dc2f-40d4-9a5f-9a919fcfe2dd

tmiw:
I'm not seeing the flickering on my end, but I also haven't updated my Fedora 42 VM in a bit. I'll run `dnf update` and try again.

Tyrbiter:
> I'm not seeing the flickering on my end, but I also haven't updated my Fedora 42 VM in a bit. I'll run `dnf update` and try again.

I can't see any flicker with Fedora 42, also worth mentioning that wxWidgets 3.2.8 landed for F42 last week so I have that.

tmiw:
> > I'm not seeing the flickering on my end, but I also haven't updated my Fedora 42 VM in a bit. I'll run `dnf update` and try again.
> 
> I can't see any flicker with Fedora 42, also worth mentioning that wxWidgets 3.2.8 landed for F42 last week so I have that.

Yeah, I just updated, rebooted and recompiled and no flicker here. I'm using Plasma in my F42 environment; not sure if that matters.

Regardless, it sounds like it might be a DE/library issue and not something with FreeDV itself. We can probably merge if the original issue (bad sorting) is resolved.

barjac:
> > > I'm not seeing the flickering on my end, but I also haven't updated my Fedora 42 VM in a bit. I'll run `dnf update` and try again.
> > 
> > 
> > I can't see any flicker with Fedora 42, also worth mentioning that wxWidgets 3.2.8 landed for F42 last week so I have that.
> 
> Yeah, I just updated, rebooted and recompiled and no flicker here. I'm using Plasma in my F42 environment; not sure if that matters.
> 
> Regardless, it sounds like it might be a DE/library issue and not something with FreeDV itself. We can probably merge if the original issue (bad sorting) is resolved.

Yes the sorting is now OK

PR #882, opened 2025-05-14
Avoid crash on stop if hamlib is somehow not able to initialize the radio

tmiw:
Discovered during testing using rig "Kenwood TS-890S" and SmartSDR on Windows. This PR makes sure we don't try to change frequency or mode to something invalid to avoid a crash in hamlib.


tmiw:
Testing confirms that the crash is fixed, but I won't be able to get confirmation that [df25893](https://github.com/drowe67/freedv-gui/pull/882/commits/df25893aa011cceb83fd7b79c32f5520ff5f609e) works until the ham in question gets back from Dayton. I'll merge for now and create another PR if needed.

PR #893, opened 2025-05-25
Improve Hamlib behavior during TX.

### LINKED ISSUE #892: [Software Issue] Slow transision to TX using TS-480 mode
### Platform

Windows

### Platform Version

Windows 10

### FreeDV Version

FreeDV 2.0.0-dev2

### Steps to Reproduce

If possible configure your test hardware to use the TS-480 configuration in the CAT and PTT Config window.  

Press FreeDV PTT.  It  will take up to four (4) seconds for the rig to go into transmit mode.

### Expected Behavior

The rig should go into TX mode in less than a second. 

### Actual Behavior

When the PTT button is pressed,  there is a delay of up to four (4) seconds before the rig is commanded to go into transmit mode.  The attached debug log file shows the delay between when the FreeDV button is pressed and when FreeDV actually commands the rig into transmit mode.  It appears to have something to do with the very long default polling delay.  It appears that FreeDV is waiting until the current 

Image

[freedv.zip](https://github.com/user-attachments/files/20431604/freedv.zip)

poll delay expires before commanding the rig into transmit mode.  

### Additional Comments

_No response_

tmiw:
Per #892, this PR does the following:

1. Imposes a 500ms timeout for Hamlib commands (with zero retries allowed).
2. Removes pre-TX interrogation of the current frequency/mode (original issue raised).
3. (Experimental) Increases Hamlib/OmniRig polling rate to 1s (from 5s).


tmiw:
> This (This PR patch applied to current v2.0-dev) breaks regular net-rigctl usage in Linux.

Any difference if you increase the timeout from 100ms to, say, 200ms? i.e. changing the following line from:

```c++
rig_set_conf(rig_, rig_token_lookup(rig_, "timeout"), "100");
```

to

```c++
rig_set_conf(rig_, rig_token_lookup(rig_, "timeout"), "200");
```

Tyrbiter:
I've taken the latest 2.0-dev branch and merged https://github.com/drowe67/freedv-gui/pull/893/commits/dbda8d04750e4aaaf959f27999c11537d84e11b0 and https://github.com/drowe67/freedv-gui/pull/897/commits/5b14a948274b14122bdab88f1745837288345a47 into it, it's now running here.

So far no problems, it seems to go into Tx almost instantly, but both of these merges probably also need to be blessed by @barjac for completeness as he found one and uses an older radio.

tmiw:
> I've taken the latest 2.0-dev branch and merged [dbda8d0](https://github.com/drowe67/freedv-gui/commit/dbda8d04750e4aaaf959f27999c11537d84e11b0) and [5b14a94](https://github.com/drowe67/freedv-gui/commit/5b14a948274b14122bdab88f1745837288345a47) into it, it's now running here.
> 
> So far no problems, it seems to go into Tx almost instantly, but both of these merges probably also need to be blessed by @barjac for completeness as he found one and uses an older radio.

I think [6713c08](https://github.com/drowe67/freedv-gui/pull/893/commits/6713c0818899f901390d813bd3c788d1bc3b8227) on its own is safe to include in 2.0.0 and will likely solve 80% of @w4mmp-ron's original problem. I'll create a new PR for that one.

barjac:
I don't see any issues now with this branch at dbda8 :)

Also tested with various TX delays set from 0 to 1000 just in case there may have been some interaction.

tmiw:
I'm going to merge this with the remaining changes now that 2.0.0 is out. If we run into problems, we can always revisit in a future PR.

PR #894, opened 2025-05-27
Additional cleanup of RT-unsafe logic.

tmiw:
Fedora 42 has Clang 20 available, which contains something cool called [RTSan](https://github.com/realtime-sanitizer/rtsan?tab=readme-ov-file). Running it on the FreeDV codebase allowed me to find several instances where real-time unsafe constructs were implicitly used. This PR contains all the fixes that we're able to implement in freedv-gui itself (Codec2 and librade can't currently be made fully real-time safe due to extensive use of `malloc()`).

### DIFF
--- a/src/gui/dialogs/dlg_audiooptions.cpp
+++ b/src/gui/dialogs/dlg_audiooptions.cpp
@@ -898,8 +898,11 @@ void AudioOptsDialog::plotDeviceInputForAFewSecs(wxString devName, PlotScalar *p
                         }
                     
                         int n8k = resample(src, in8k_short, in48k_short, 8000, sampleRate, TEST_BUF_SIZE, TEST_BUF_SIZE);
-                        resample_for_plot(fifo, in8k_short, n8k, FS);
-    
+                        short* tmp = new short[n8k];
+                        assert(tmp != nullptr);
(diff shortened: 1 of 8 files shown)


Tyrbiter:
Is this something that could make it into 2.0.0 or should it wait for some real world testing?

tmiw:
> Is this something that could make it into 2.0.0 or should it wait for some real world testing?

Unit tests on GitHub all pass (except on Intel macOS, but that's a known problem) and some testing locally of the modified code doesn't indicate any issues, so I think we can include this in 2.0.0. 👍

tmiw:
> > > Running happily here for the last few hours, no difference from the previous build in behaviour.
> > 
> > 
> > Any change in the fifo or underflow counters at all?
> 
> With 4 million or so outempty1 and 2 I have infull1 and 2 0 with both Audio 1 and 2 outUnderflow around 11 to 14.
> 
> Essentially no difference, but the ratios are high so it's difficult to be accurate.

Yeah, I suppose there's only so much that can be done when the codec itself is still using a lot of RT-unsafe practices. There's some more work that can be done to further clean up RT-unsafe practices in this repo but that's probably better post-2.0.0 release.

PR #896, opened 2025-05-27
Merge v2.0-dev with master

tmiw:
This PR merges v2.0-dev up to master (along with needed changelog/version tagging). After this is merged, official builds of 2.0.0 will be generated and published in appropriate channels. Further development in v2.0-dev will also cease in favor of PRs against master, with a v2.1-dev branch coming at some point in the future.


Tyrbiter:
I have tried to edit the USER_MANUAL.md file a bit. It's probably a bit of a WIP, but it does remove things that definitely should be removed. I had to tar it as a plain patch file was not working:

[USER_MANUAL.tar.gz](https://github.com/user-attachments/files/20542337/USER_MANUAL.tar.gz)

HTH

USER_MANUAL.md
```diff
@@ -509,7 +469,7 @@ This section describes features on Tools-Filter.
 Control | Description
  -------------------------- | ------------------------------------------------------------------------ |
 Noise Suppression | Enable noise suppression, de-reverberation, AGC of mic signal using the Speex pre-processor
-700C/700D Auto EQ | Automatic equalization for FreeDV 700C and FreeDV 700D Codec input audio
+700D Auto EQ | Automatic equalization for FreeDV 700D/E Codec input audio
```
tmiw, line 472: I thought auto EQ was only for 700D but I guess it makes sense that it'd work for E too.

USER_MANUAL.md
```diff
@@ -405,6 +402,8 @@ Additionally, the squelch setting with simultaneous decode enabled is relative t
 difference between the "Min SNR" of 700D and the mode in question; see "FreeDV Modes" below). For example, the squelch for 700E
 when the squelch slider is set to -2.0 becomes 1.0dB. This is designed to reduce undesired pops and clicks due to false decodes.
 
+RADEV1 does not use the squelch and when this mode is enabled the Squelch is disabled and the slider setting is ignored.
```
tmiw, line 405: Related note: Do you think the Squelch settings should be hidden/grayed out for 2.0.1?

Also maybe this should be the following instead:

```
When using RADEV1, the squelch settings in the main window are ignored. Instead, FreeDV only passes decoded audio if it's able to synchronize with the incoming signal.
```

tmiw (commented):
Thanks @Tyrbiter!

PR #915, opened 2025-06-05
Only resort if we actually get new data from the server.

tmiw:
Adds flag to only trigger resorting of the rows in FreeDV Reporter if new data arrives from the server. This reduces CPU usage by ~25% (40% to 30% on macOS for example).

### DIFF
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -790,7 +790,12 @@ void FreeDVReporterDialog::OnTimer(wxTimerEvent& event)
 {
     FreeDVReporterDataModel* model = (FreeDVReporterDataModel*)spotsDataModel_.get();
     model->updateHighlights();
-    model->Resort();
+    
+    if (model->sortOnNextTimerInterval)
+    {
+        model->Resort();
+        model->sortOnNextTimerInterval = false;
+    }
 }
 
 void FreeDVReporterDialog::OnFilterTrackingEnable(wxCommandEvent& event)
@@ -1412,6 +1417,7 @@ FreeDVReporterDialog::FreeDVReporterDataModel::FreeDVReporterDataModel(FreeDVRep
     , currentBandFilter_(FreeDVReporterDialog::BAND_ALL)
     , filterSelfMessageUpdates_(false)
     , filteredFrequency_(0)
+    , sortOnNextTimerInterval(false)
 {
     // empty
 }
@@ -1847,6 +1853,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::refreshAllRows()
             {
                 ItemDeleted(wxDataViewItem(nullptr), wxDataViewItem(kvp.second));
             }
+            sortOnNextTimerInterval = true;
         }
         else if (updated && kvp.second->isVisible)
         {
@@ -2003,6 +2010,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::onUserConnectFn_(std::string
         if (temp->isVisible)
         {
             ItemAdded(wxDataViewItem(nullptr), wxDataViewItem(temp));
+            sortOnNextTimerInterval = true;
         }
     });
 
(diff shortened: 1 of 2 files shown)


tmiw:
@barjac, @Tyrbiter, I know we had problems with sorting recently. Can you check to make sure this didn't break anything?

Tyrbiter:
Running this now, I tried various things in the Reporter window and it all seems to behave as expected.

@barjac may well find something, he is more demanding of the reporter features I'd say.

jaspejavier:
Hi! In case it is useful: before this fix, my cpu jumped from 16-20% cpu usage without reporter to about 40% cpu with reporter open. After applying this pull req, cpu jumps from 16-20% to only 20-25% instead. Tried sorting every column ascending and descending and it works ok for me. (btw I'm now compilling against `libwxgtk3.2-dev` - please tell me if you want it tested against `libwxgtk3.0-gtk3-dev` too) 73 de LU3JIJ

tmiw:
I think I was able to get CPU usage down a bit more. Let me know if the latest in this PR works well for everyone.

tmiw:
What I've noticed is that if someone's transmitting or receiving, CPU usage goes back up to pre-PR levels. In the past I've tried re-sorting only if data corresponding to the current sort column's changed but I definitely remember that causing problems before. Maybe we can try that again plus sort when rows appear/disappear?

Tyrbiter:
It's really difficult to tie this down. I notice that one or two stations running older code are not updating at all and have stuck rows. Not sure if that can make a difference to CPU usage.

tmiw:
OK, I just pushed two commits. https://github.com/drowe67/freedv-gui/pull/915/commits/2c614b12485b20420f50e8973681d069817d81ce should be approximately what we tried during development of the latest iteration of FreeDV Reporter (but always re-sorts on user connect/disconnect). I'll hold off on merging for a few days while we see if this has any issues.

BTW on my 2019 MBP, the latest commit gets CPU usage of the FreeDV Reporter window down to about an average of 10% and maximum of 20% (without the Start button having been pressed). This IMO is good enough to resolve the original issue if it doesn't introduce any problems.

barjac:
This all seems to be behaving OK here. (except for an assert #920 which is unrelated as I have seen it before these changes)

I am seeing an increase of < 10% on starting the reporter but this drops back over a couple of seconds to around 2% over the general load of FreeDV-GUI which now seems higher (now at around 50% to 80%) than I remember it being before I switched to Wayland (~30% IIRC in RX).

I explained how to rebuild this branch to DF6VI which he did while we were in QSO and he reported a big improvement in his Linux system which was breaking previously with the reporter running (IIUC).

Tyrbiter:
Difficult to be sure about load but it doesn't seem to break anything here, yet :) I will report further once I have some hard numbers.

https://github.com/drowe67/freedv-gui/pull/918 has got rid of this:

12:55:59: Debug: 1 threads were not terminated by the application.

So the typo @tmiw fixed was the cause.

tmiw:
Oops, I found some typos and got a crash while I was testing locally. Hopefully the same good results are there with the latest?

jaspejavier:
I built this branch again, but using `libwxgtk3.0-gtk3-dev`. Now I'm getting:

`/usr/include/wx-3.0/wx/datetime.h(876): assert "IsValid() && dt.IsValid()" failed in operator!=(): invalid wxDateTime`
`/usr/include/wx-3.0/wx/datetime.h(1733): assert "IsValid()" failed in GetValue(): invalid wxDateTime`

Several times. 

If I uncheck "Show this dialog next time", FreeDV works OK
If I go back to branch `master`, FreeDV works OK
If I use `libwxgtk3.2-dev`, FreeDV works OK

tmiw:
OK, I think I resolved the wxWidgets 3.0-related assertions. Try now?

jaspejavier:
Yes! Now ok on wx3.0!

tmiw:
CPU usage still seemed a bit high so I took a look using Instruments on macOS. Seems that since now we're requesting frequency updates from Hamlib every second (instead of every five seconds), we're sending server updates every second or so even if the frequency hasn't changed. I went ahead and added some logic to only send updates to the server when the frequency actually updates since I can see CPU usage ramping right back up once 2.0.1 gets released.

tmiw:
I'll go ahead and merge since I haven't heard of any issues with the most recent commits. We can revisit in another PR if needed.

PR #925, opened 2025-06-09
Add missing dependency for macOS builds

relistan:
Super tiny README fix. The build tooling depends on `wget` to build on macOS and it's not part of the base distro from Apple.


tmiw:
Merged. Thanks!

relistan:
@tmiw 🤦 I should have added to Macports. Thanks for catching and merging!

Issue #920, opened 2025-06-07
Assert happening occasionally possibly related to graphic in reporter message.


barjac:
ASSERT INFO:
./src/common/strconv.cpp(1165): assert ""Assert failure"" failed in FromWChar(): trying to encode undefined Unicode character

BACKTRACE:
[1] wxMBConvStrictUTF8::FromWChar(char*, unsigned long, wchar_t const*, unsigned long) const
[2] wxString::AsChar(wxMBConv const&) const
[3] wxDataViewTextRenderer::SetTextValue(wxString const&)
[4] wxDataViewRendererBase::PrepareForItem(wxDataViewModel const*, wxDataViewItem const&, unsigned int)
[5] g_hash_table_foreach
[6] g_signal_emit_valist
[7] g_signal_emit
[8] gtk_cell_area_apply_attributes
[9] g_signal_emit_valist
[10] g_signal_emit
[11] g_main_loop_run
[12] gtk_main
[13] wxGUIEventLoop::DoRun()
[14] wxEventLoopBase::Run()
[15] wxAppConsoleBase::MainLoop()
[16] wxEntry(int&, wchar_t**)
[17] main /home/baz/freedv-rade/freedv-gui/src/main.cpp:203
[18] __libc_start_main
[19] _start

This has happened about four times during testing over the last few days since 2.0.0 release but not certain that it relates to the release as I was using PR branches previously anyway.

The above was from ms-reporter-cpu-usage branch today during a QSO on 20m with a German station while exchanging messages on the reporter where I had a flag displayed which may have contributed.

barjac:
There is also a problem when trying to backspace over a graphic in the Message entry dialogue.

A coloured flag becomes a mono generic flag and the cursor will not backspace over it.

If the cursor is moved ahead (left) of the graphic it can be deleted using the delete key.

Sadly none of this triggers the assert in testing to try to reproduce this bug. :(

tmiw:
I wonder if 5bd401b might have caused it as it was touching code around UTF-8 conversion. Can you see if the official 2.0.0 release has the same problem?

barjac:
Several linux stations using older versions from before 2.0.0 release see the flag issue when trying to backspace over the graphic.

The assert issue is very rare with no known way to reproduce. Maybe best to wait for other reports. I will try running 2.0.0 for a while.

Tyrbiter:
Just checked, I also see the flag colour/b+w thing too, no idea what is happening or even how to debug it.

[...]

PR #929, opened 2025-06-11
Fix GTK+ assertion after FreeDV Reporter has been open for a long time.

tmiw:
After the FreeDV Reporter window has been open for at least a few hours, the following GTK+ assertion appears in the terminal window:

```
(freedv:1307905): Gtk-CRITICAL **: 21:52:53.357: ../gtk/gtktreeview.c:6712 (validate_visible_area): assertion `has_next' failed.
There is a disparity between the internal view of the GtkTreeView,
and the GtkTreeModel. This generally means that the model has changed
without letting the view know. Any display from now on is likely to
be incorrect.

(freedv:1307905): Gtk-CRITICAL **: 21:52:53.600: _gtk_rbtree_reorder: assertion 'tree->root->count == length' failed
```

This does not seem to be producing any other problems, but out of an abundance of caution this PR was created to resolve this warning. One theory is that the `CallAfter()` calls that the FreeDV Reporter window are doing are somehow occurring out of order vs. internal events thrown by the `wxDataViewCtrl` control, so this PR basically ensures that all queued events are processed every time server messages trigger `CallAfter()`.

(Originally reported [here](https://github.com/drowe67/freedv-gui/issues/920#issuecomment-2960576297).)

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -802,6 +802,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix WASAPI errors on some machines by supporting audio mix formats other than 16-bit integer. (PR #919)
     * Reduce CPU usage of FreeDV Reporter window by only re-sorting if we actually get new data from the server. (PR #915)
     * FreeDV Reporter: Fix issue with first column not being aligned properly with other columns. (PR #922)
+    * Fix GTK+ assertion after FreeDV Reporter has been open for a long time. (PR #929)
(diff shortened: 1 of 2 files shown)


Tyrbiter:
I have this running, and will see how it goes.

tmiw:
FWIW I had this running all day today and I've yet to see that assertion. In previous testing I believe it would have appeared by now without this fix, so I think it's safe to merge. We can always revisit if it causes other problems.

PR #931, opened 2025-06-12
Work around Linux bug preventing some flag emojis from being fully deleted on backspace.

tmiw:
This PR works around Linux emoji flag bug reported in https://github.com/drowe67/freedv-gui/issues/920 by detecting malformed emoji flags on text update and deleting them. Per https://en.wikipedia.org/wiki/Regional_indicator_symbol:

> A separate mechanism ([emoji tag sequences](https://en.wikipedia.org/wiki/Tags_(Unicode_block))) is used for regional flags, such as England 🏴󠁧󠁢󠁥󠁮󠁧󠁿, Scotland 🏴󠁧󠁢󠁳󠁣󠁴󠁿, Wales 🏴󠁧󠁢󠁷󠁬󠁳󠁿, Texas 🏴󠁵󠁳󠁴󠁸󠁿 or California 🏴󠁵󠁳󠁣󠁡󠁿.[[12]](https://en.wikipedia.org/wiki/Regional_indicator_symbol#cite_note-12) It uses [U+](https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs)1F3F4 🏴 WAVING BLACK FLAG and formatting [tag characters](https://en.wikipedia.org/wiki/Tags_(Unicode_block)) instead of regional indicator symbols. It is based on [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) regions with hyphen removed and lowercase, e.g. GB-ENG → gbeng, terminating with [U+](https://en.wikipedia.org/wiki/Tags_(Unicode_block))E007F CANCEL TAG. Flag of England is therefore represented by a sequence U+1F3F4, U+E0067, U+E0062, U+E0065, U+E006E, U+E0067, U+E007F. In the tenth revision the Unicode consortium was considering [U+](https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs)1F3F3 🏳 WAVING WHITE FLAG instead,[[13]](https://en.wikipedia.org/wiki/Regional_indicator_symbol#cite_note-13) but from eleventh onwards it is black.[[14]](https://en.wikipedia.org/wiki/Regional_indicator_symbol#cite_note-14)

The logic in this PR simply detects whether U+1F3F4 exists but not U+E007F. If this condition exists, the entire character sequence is deleted. Note that if only U+1F3F4 exists but no other Unicode tag characters do, the logic does not delete the flag (mainly because Linux/GTK/etc. already do the right thing in this scenario).


Tyrbiter:
This is mainly working for me, however with the England flag I need to backspace twice to remove it, I suspect this is probably true for all of the regional flags. I imagine this is fixable, although I have just seen the note in the commit, so perhaps it isn't.

I suppose most people will just hit backspace again thinking that the keyboard glitched.

tmiw:
> This is mainly working for me, however with the England flag I need to backspace twice to remove it, I suspect this is probably true for all of the regional flags. I imagine this is fixable, although I have just seen the note in the commit, so perhaps it isn't.
> 
> I suppose most people will just hit backspace again thinking that the keyboard glitched.

Yeah, it might be possible to get rid of that second backspace but didn't want the higher impact that would cause (plus the possibility that the fix would have to eventually be backed out if the original bug was ever fixed upstream). At least with this solution it should be okay to keep it in the code since it'll only modify the text if there's actually a problem.

PR #933, opened 2025-06-13
Lock real time memory pages in memory to avoid being swapped out.

tmiw:
WIP - DO NOT MERGE

Page faults are another potential source of audio dropouts. This PR uses OS-specific system calls to lock certain areas of memory and prevent them from being swapped out, as being brought back from swap can take significant time.


barjac:
I have been running this branch for several days now, both on Mageia 9 and 10 without any issues. Last night I ran two instances on my desktop with one taking the downlink from QO-100 via the Goonhilly SDR and the other TX/RX on 160m using i5 desktop 4 core 8GB RAM with load average in top around 1.5. We had an excellent cross band QSO with 15dB SNR from the satellite with only about 6W uplink power.
For me this branch seems very stable.

tmiw:
> I have been running this branch for several days now, both on Mageia 9 and 10 without any issues. Last night I ran two instances on my desktop with one taking the downlink from QO-100 via the Goonhilly SDR and the other TX/RX on 160m using i5 desktop 4 core 8GB RAM with load average in top around 1.5. We had an excellent cross band QSO with 15dB SNR from the satellite with only about 6W uplink power. For me this branch seems very stable.

Thanks for the feedback! Unfortunately the primary criteria I was using to determine whether this can be merged (an improvement in the GH Action pass rate for this branch vs `master`, especially for macOS) still hasn't been met and I'm not sure if we want to put much more effort into this line of investigation. I may take a few of the changes from this PR and merge them in separately (as they're fairly low-risk) and otherwise close out this PR.

PR #939, opened 2025-06-18
Legacy FreeDV modes: ensure that sync is obtained in an atomic manner.

tmiw:
Current Codec2 clears the sync flag internally for 700D and E modes at the beginning of `freedv_rx()`. Because of this, there is a small window where sync is not in its final state when other parts of the system retrieve it. This is most evident in macOS tests failing in the GitHub environment but this is theoretically possible on (a) any platform and (b) in user-facing contexts (i.e. brief audio dropouts due to the multi-RX logic occurring while decode is still in progress). 

As most development is focused on RADE, this PR updates the freedv-gui logic to only retrieve the sync state from Codec2 *after* `FreeDVReceiveStep` finishes processing a block of audio, ensuring that other parts of the system always see consistent state for sync.

### DIFF
--- a/src/freedv_interface.cpp
+++ b/src/freedv_interface.cpp
@@ -829,7 +829,8 @@ int FreeDVInterface::preProcessRxFn_(ParallelStep* stepObj)
     std::shared_ptr<ReceivePipelineState> state = std::static_pointer_cast<ReceivePipelineState>(stepObj->getState());
 
     // Set initial state for each step prior to execution.
-    for (auto& step : stepObj->getParallelSteps())
+    auto& parallelSteps = stepObj->getParallelSteps();
+    for (auto& step : parallelSteps)
     {
         assert(step != nullptr);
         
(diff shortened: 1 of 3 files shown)


tmiw:
## Failures on `master` (last 10 builds)

* macOS: 7/10 builds failed
    * x86_64: rade_reporting_clean (1), fullduplex_700E (4), fullduplex_700D (1) 
    * aarch64: rade_reporting_mpp (1), fullduplex_700E (1)
* Linux: 0/10
* Windows: 1/10 (GitHub environment problem--could not install virtual audio devices)

## Failures in this PR (last 10 builds)

* macOS: 3/10 builds failed
    * x86_64: fullduplex_700D (1), fullduplex_1600 (1), rade_reporting_clean (1)
    * aarch64: No failures
* Linux: 0/10
* Windows: 0/10

Based on the above results, this PR cuts the number of macOS failures in the GH environment in half and is worthwhile to merge.

*Note: the fullduplex_700D and fullduplex_1600 failures in this PR appear to occur right as full duplex comms kicks off, so there may be an additional bug that needs to be fixed. This will be tracked separately.*

PR #953, opened 2025-06-29
Be explicit about the use of signed char for reporting.

tmiw:
The reporting interface uses `char` for the SNR. Most other C data types default to `signed` but whether `char` defaults to signed depends on the compiler and platform, causing some systems (for example, Raspberry Pi) to report extremely high SNR figures to FreeDV Reporter instead of negative SNRs as expected. This PR forces used of `signed char` for the SNR field to ensure that reporting behavior is correct regardless of platform.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -823,6 +823,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Legacy FreeDV modes: ensure that sync is obtained in an atomic manner. (PR #939)
     * Fix intermittent crash on Windows when pushing Start. (PR #943)
     * FreeDV Reporter: Defer deallocation of disconnected users. (PR #948)
+    * FreeDV Reporter: Be explicit about the use of signed char for reporting. (PR #953)
 2. Documentation:
     * Add missing dependency for macOS builds to README. (PR #925; thanks @relistan!)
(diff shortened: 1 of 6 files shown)


Tyrbiter:
I didn't expect to see any difference running this on Linux, presumably it should be merged before releasing 2.0.1?

tmiw:
> I didn't expect to see any difference running this on Linux, presumably it should be merged before releasing 2.0.1?

Yeah, ideally this would be merged soon. We might need to test this on a Raspberry Pi (like the original bug report) as I think the code in the PR is equivalent to what's currently in master on x86.

Tyrbiter:
> > I didn't expect to see any difference running this on Linux, presumably it should be merged before releasing 2.0.1?
> 
> Yeah, ideally this would be merged soon. We might need to test this on a Raspberry Pi (like the original bug report) as I think the code in the PR is equivalent to what's currently in master on x86.

I don't recall ever seeing a very high SNR report. Do we know who reported this problem? I can't see anything in the issues list.

tmiw:
One of the people on the Radio Autoencoder list did. I also saw a JA station with like a 250 dB RX SNR the other day too and wasn't sure what that was about, but it makes sense now.

tmiw:
Going to go ahead and merge so we can get another test build out.

PR #957, opened 2025-07-04
Make explicit various codec2 dependencies to avoid compile race condition.

### LINKED ISSUE #956: Parallel builds with many cores cause build failures due to race condition.
I have been chasing an issue seen a few times where apparently freedv_api.h is not found during a build_linux.sh run.

A system in question was unable to build using my script, for no obvious reason.

I then noticed -j16 in the build log and we tested a patched build_linux.sh with  
-j1 which built OK.
https://github.com/barjac/freedv-rade-build/issues/20 has build logs.

He has now tested it with -j4 and that also built OK but faster.
Maybe build_linux.sh could set a max limit of maybe four? I have two 4 core machines which have not had trouble, however something higher may still be OK.

Another approach that is probably not worth the effort may be to allow subsequent builds to start using -j$(nproc) only when previous sections have fully completed.

tmiw:
Resolves #956 by adding explicit `add_dependencies` calls in CMake files to ensure Codec2 is built prior to anything that needs it.


barjac:
This does not cause any regressions for me. I have added a 'race-tests' branch in freedv-rade-build for Dave to test this PR on his 16 core machine, but he has not yet got back about the first patch results. (in add-garuda branch)
I have asked him to try 'race-tests' branch if the first tests fail.

If there is no feedback in a few days then I think this 'boots and braces' patch should be merged ;)

tmiw:
I've been doing a bunch of build tests and I'd say it's pretty safe to merge without hearing back first. We can always do a followup if there's an issue.

PR #962, opened 2025-07-07
Disable controls not supported by RADE.

tmiw:
This PR disables controls and tabs on the main window that aren't yet (?) supported by RADE in order to avoid user confusion. Various stats that are relevant for non-RADE are also changed to report "unk" instead when RADE is the current mode.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -768,6 +768,25 @@ setDefaultMode:
         mode = defaultMode;
         goto setDefaultMode;
     }
+    
+    // Disable controls not supported by RADE.
+    bool isEnabled = mode != FREEDV_MODE_RADE;
+    m_sliderSQ->Enable(isEnabled);
+    m_ckboxSQ->Enable(isEnabled);
+    m_textSQ->Enable(isEnabled);
+    m_btnCenterRx->Enable(isEnabled);
+    
+    if (!isEnabled)
+    {
+        m_textBits->SetLabel("Bits: unk");
+        m_textErrors->SetLabel("Errs: unk");
+        m_textBER->SetLabel("BER: unk");
+        m_textFreqOffset->SetLabel("FrqOff: unk");
+        m_textSyncMetric->SetLabel("Sync: unk");
+        m_textCodec2Var->SetLabel("Var: unk");
+        m_textClockOffset->SetLabel("ClkOff: unk");
+    }
+    
     pConfig->SetPath(wxT("/"));
     
     m_togBtnAnalog->Disable();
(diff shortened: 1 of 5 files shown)


Tyrbiter:
This seems to work, the only issue is that the main window tabs that are now unselectable don't look any different as you mentioned via email. This should be explained somehow, but I don't know if a mode-related tooltip is possible so some words in the manual should do for now.

Just noticed there is debug output that tells me when I try to select a forbidden tab.

barjac:
Is the frequency adjustment above the waterfall and spectrum supposed to work in RADE?
It seems to work in all modes graphically, however the 'Center RX' button is greyed in RADE but active in old modes.
The centre mouse click works in all modes.

barjac:
The tab order under the waterfall is not being saved between runs.

Tyrbiter:
> The tab order under the waterfall is not being saved between runs.

It might make sense to have a different tab order for RADE vs legacy modes, but not if it's a difficult request.

tmiw:
> The tab order under the waterfall is not being saved between runs.

This is something that has to be explicitly enabled (Tools->Options->Debug->Enable Experimental Features) due to how it's currently being implemented. 

> Is the frequency adjustment above the waterfall and spectrum supposed to work in RADE? It seems to work in all modes graphically, however the 'Center RX' button is greyed in RADE but active in old modes. The centre mouse click works in all modes.

This isn't supposed to work in RADE. I'll check in a change shortly to make those lines not appear when RADE's selected.

> This seems to work, the only issue is that the main window tabs that are now unselectable don't look any different as you mentioned via email. This should be explained somehow, but I don't know if a mode-related tooltip is possible so some words in the manual should do for now.

Would a message box popping up work too, or would that be too annoying for people?

Tyrbiter:
> > The tab order under the waterfall is not being saved between runs.
> 
> This is something that has to be explicitly enabled (Tools->Options->Debug->Enable Experimental Features) due to how it's currently being implemented.

Ah, I had forgotten about this setting's purpose.

> > This seems to work, the only issue is that the main window tabs that are now unselectable don't look any different as you mentioned via email. This should be explained somehow, but I don't know if a mode-related tooltip is possible so some words in the manual should do for now.
> 
> Would a message box popping up work too, or would that be too annoying for people?

Would this need closing, or could it just close when the mouse click is processed and the box has been there for a second or so?

Tyrbiter:
https://github.com/drowe67/freedv-gui/pull/962/commits/61b463fda529c305f9fdf34b451a7ce5f823dedd

The missing line is a bit of a problem because you can't see that the double click has moved the centre frequency, or that the middle click has re-centred the Rx.

barjac:
> [61b463f](https://github.com/drowe67/freedv-gui/commit/61b463fda529c305f9fdf34b451a7ce5f823dedd)
> 
> The missing line is a bit of a problem because you can't see that the double click has moved the centre frequency, or that the middle click has re-centred the Rx.

If the frequency slider does not work in RADE1 then does that matter? :)

In RADE when trying to slide the scatter tab out of the way it still does show until another tab is clicked.

barjac:
Just tried experimental mode and scatter position is not remembered. I'm referring to the TAB position not a window position. It makes sense to move the unused tabs to the far right which is what I was trying to do, but even with experimental mode it does not survive a re-start of the program.

tmiw:
> Would this need closing, or could it just close when the mouse click is processed and the box has been there for a second or so?

I was thinking something that needed to be closed after it appeared.

> Just tried experimental mode and scatter position is not remembered. I'm referring to the TAB position not a window position. It makes sense to move the unused tabs to the far right which is what I was trying to do, but even with experimental mode it does not survive a re-start of the program.

Is this something that used to work prior to this PR? Admittedly I don't really try to rearrange the tabs all that often so I'm not sure.

Tyrbiter:
I suspect this is because the tabs are within a single window 'frame' but when you drag a tab out into its own 'frame' it's treated differently. This is a guess but perhaps the underlying structure that wxWindows uses to specify the layout is unable to recognise the difference with tabs within a frame.

tmiw:
> I suspect this is because the tabs are within a single window 'frame' but when you drag a tab out into its own 'frame' it's treated differently. This is a guess but perhaps the underlying structure that wxWindows uses to specify the layout is unable to recognise the difference with tabs within a frame.

The tabs are initially added in a particular order when the window loads, and then rearranged as appropriate (assuming "Enable Experimental Features" is turned on). It sounds like from what @barjac is saying, pushing the Tab key cycles through according to what the window looked like on startup and not based on where the tabs ultimately ended up.

Tyrbiter:
> > I suspect this is because the tabs are within a single window 'frame' but when you drag a tab out into its own 'frame' it's treated differently. This is a guess but perhaps the underlying structure that wxWindows uses to specify the layout is unable to recognise the difference with tabs within a frame.
> 
> The tabs are initially added in a particular order when the window loads, and then rearranged as appropriate (assuming "Enable Experimental Features" is turned on). It sounds like from what @barjac is saying, pushing the Tab key cycles through according to what the window looked like on startup and not based on where the tabs ultimately ended up.

I don't see any effect from the Tab key. I think he means drag a tab out of the standard order, close freedv-gui, reopen and the tabs are back to the original positions.

barjac:
> I don't see any effect from the Tab key. I think he means drag a tab out of the standard order, close freedv-gui, reopen and the tabs are back to the original positions.

Correct - nothing to do with the keyboard at all.

The Window 'tabs' under the waterfall etc.
Click and drag one into a new position, (e.g. 'Scatter' to the right hand end of the tabs), close the program and re-start it and the tab reverts to where it was before it was re-positioned.

However I just tested this in 1.9.2 and it is the same.(With experimental on)

Why then is the order different on another system of mine running 2.0.0, if it cannot be changed?

tmiw:
Behavior is the same in current master as well. Can you submit another issue/feature request for this? We'll keep this PR just about the disabling of unusable tabs.

Tyrbiter:
The latest works, although it's not the nicest fix ;-)

Will you create a issue for the tab order @barjac as you spotted the problem.

barjac:
> The latest works, although it's not the nicest fix ;-)
> 
> Will you create a issue for the tab order @barjac as you spotted the problem.

Yes it works OK but I don't like it either ;)

Could not all the useless ones for RADE just be disabled (as in not shown at all) if RADE is selected?

This would probably make the need for re-positioning the tabs redundant.

Tyrbiter:
> > The latest works, although it's not the nicest fix ;-)
> > Will you create a issue for the tab order @barjac as you spotted the problem.
> 
> Yes it works OK but I don't like it either ;)
> 
> Could not all the useless ones for RADE just be disabled (as in not shown at all) if RADE is selected?
> 
> This would probably make the need for re-positioning the tabs redundant.

I think that this is to do with the creation of the window, when you change mode it doesn't recreate the window so the tabs are always present until you close the program completely.

barjac:
> I think that this is to do with the creation of the window, when you change mode it doesn't recreate the window so the tabs are always present until you close the program completely.

Hmm... At present the pop-up dialogue stops the unused tabs from being moved at all even just for the current session which is worse than it was! :(

Since use of the old modes is now minimal, the semi redundant tabs could be moved to the end of the row, out of sight :)

Agggh! 'To Spkr/Hdphns' should not have been removed!

Tyrbiter:
> Agggh! 'To Spkr/Hdphns' should not have been removed!

It hasn't, still works fine here.

tmiw:
I got rid of the veto logic and just moved the unsupportred tabs as far right as I can go. Hopefully that works better.

barjac:
As I commented previously the order of the tabs is different on my laptop. On this for some reason the waterfall is far right.
Looking at .freedv I notice that I still have:
```
FirstTimeUse=0
ExperimentalFeatures=1
[FreeDV2020]
Allowed=0
[MainFrame]
left=0
top=15
width=1358
height=363
rxNbookCtrl=0
TabLayout=014882e06532d38d00034dfb00000002=2,3,4,6,5,7,8,9,1,*0@layout2|name=dummy;caption=;state=2098174;dir=3;layer=0;row=0;pos=0;prop=100000;bestw=258;besth=133;minw=258;minh=133;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1|name=014882e06532d38d00034dfb00000002;caption=;state=2098172;dir=5;layer=0;row=0;pos=0;prop=100000;bestw=0;besth=0;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1|dock_size(5,0,0)=10|
[Audio]
```

The above snippet from [FeeDV2020] is obviously still controlling the tab positions such that on this machine using this PR only two tabs actually work.

I was not it seems imagining that these tab positions were once being saved :)

Maybe this existing mechanism could be re-used to advantage?

Also this PR as it stands will break systems with old config files.

barjac:
> I got rid of the veto logic and just moved the unsupportred tabs as far right as I can go. Hopefully that works better.

Oops! I missed that comment as I was busy typing!

I'm a bit confused since 'TabLayout=' in a fresh .freedv config file with this branch is empty and scatter is still at #2 (3rd position).

Tyrbiter:
Works for me, scatter tab is now the 6th from the left and while selectable, it's out of the way.

barjac:
> Works for me, scatter tab is now the 6th from the left and while selectable, it's out of the way.

OK yes this works except when .freedv already contains a 'TabLayout=...' when it follows that instead, and since that cannot now be edited by dragging tabs there is chicken and egg situation. Maybe remove that line on closedown if it exists? First run would be broken but OK afterwards OR patch to ignore it for now?

Bit busy with other stuff but I will add an issue to restore the tab positioning maybe tonight.

tmiw:
Yeah, I'm able to move tabs around with a fresh `~/.freedv` but of course the positioning isn't saved. Honestly, that part of the UI has always been pretty glitchy, but I get that people do move tabs around to rearrange them so I'm not sure removing that ability is a good idea either.

Anyway, besides the tab positioning not being saved, any other issues preventing this from being merged?

Tyrbiter:
> > Works for me, scatter tab is now the 6th from the left and while selectable, it's out of the way.
> 
> OK yes this works except when .freedv already contains a 'TabLayout=...' when it follows that instead, and since that cannot now be edited by dragging tabs there is chicken and egg situation. Maybe remove that line on closedown if it exists? First run would be broken but OK afterwards OR patch to ignore it for now?

Whether that's to do with swapping .freedv files locally or in the coding of the settings save feature some time in the past I don't know. I think there was some special handling for Windows as well, maybe a wxWindows feature.

I think I can drag tabs now, but not sure if they are remembered yet. That is with https://github.com/drowe67/freedv-gui/pull/962/commits/4fde5ea943bc9fee1aa6f74435f1062d00d82eff

Tyrbiter:
> Yeah, I'm able to move tabs around with a fresh `~/.freedv` but of course the positioning isn't saved. Honestly, that part of the UI has always been pretty glitchy, but I get that people do move tabs around to rearrange them so I'm not sure removing that ability is a good idea either.
> 
> Anyway, besides the tab positioning not being saved, any other issues preventing this from being merged?

I think it's good enough for 2.0.1 and maybe then consider a better strategy for layout changes and saving.

tmiw:
OK, I'll merge for now. Once @barjac creates that followup issue we'll investigate further.

PR #966, opened 2025-07-09
Force left-to-right rendering of UI elements.

### LINKED ISSUE #965: [Software Issue] Window orientation is affected by the operating system's regional settings.
### Platform

Windows

### Platform Version

11

### FreeDV Version

2.0.0

### Steps to Reproduce

1. Set the regional format to Hebrew
2. Execute FreeDV
Image

This will cause the application to display in RTL which feels weird (especially if the text is still in english)

Image

### Expected Behavior

Just the normal LTR display...

### Actual Behavior

Image

### Additional Comments

_No response_

tmiw:
This PR resolves #965 by forcing window rendering to occur left-to-right rather than right-to-left (on Windows and Linux, at least). By doing this, we avoid having to make wider changes around locale handling (i.e. forcing rendering of numbers/dates to en_US-style).

### DIFF
--- a/src/gui/controls/plot.cpp
+++ b/src/gui/controls/plot.cpp
@@ -46,6 +46,10 @@ END_EVENT_TABLE()
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
 PlotPanel::PlotPanel(wxWindow* parent, const char* plotName) : wxPanel(parent, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxTAB_TRAVERSAL, plotName)
 {
+    // XXX - FreeDV only supports English but makes a best effort to at least use regional formatting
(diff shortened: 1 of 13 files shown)


Tyrbiter:
Have this on test now, is there anything in particular that needs looking at.

tmiw:
You probably won't notice any changes unless your DE's language is set to something that reads right to left (like Arabic or Hebrew). Ideally if your environment is set to use such a language FreeDV will look the same with this PR as if it was set to English.

Tyrbiter:
This commit has not broken anything for me here, has the person who reported it with a Hebrew DE setting reported any change or have they not reported back yet?

tmiw:
I set `LANGUAGE=ar_JO.UTF-8` and ran this PR. So that's cool (title):

Screenshot 2025-07-11 at 9 11 01 AM

Anyway, I think this PR is working. Here's a comparison with `master`:

Screenshot 2025-07-11 at 9 13 30 AM

Anyway, I think we can merge and do a followup one if the OP reports a problem.

PR #968, opened 2025-07-10
Add notes for linux users having trouble with hamlib

peterbmarks:
A common problem for Linux users is pre-installed software that grabs serial ports and interferes with hamlib talking to the radio. I've added a note to common problems on how to disable ModemManager and brltty.


Tyrbiter:
This is a fairly complex issue as you can see in this page:

https://modemmanager.org/docs/modemmanager/port-and-device-detection/

If people need ModemManager for some devices I think it should be possible to prevent access to tty ports used for radio control, but it's probably not that simple.

For reference, I have always had ModemManager running here and it has never caused me a problem. Currently Fedora 42 has version 1.22.0 which is the most recent I am aware of.

Problems with this could be due to older versions that don't have the same ability to avoid probing problematic ports.

tmiw:
Yeah, I wonder if it would be better to provide info about blacklisting ports from ModemManager control rather than disabling it entirely. I don't really know much about it, though, so that may very well be difficult to impossible.

Tyrbiter:
It's certainly not the easiest thing to configure, but then anything involving udev rules is like that if there are no simple examples to follow.

peterbmarks:
I'm not sure how much of a problem ModemManager might cause. Do many people still use dial-up modems? QRP-Labs gives the same advice in their user guides. Certainly brltty is a problem and comes installed and enabled on Ubuntu and Rasparian. My proposed wording is just a draft but I think it fits in the "common problems" area.

Tyrbiter:
I agree, however I think that it's cellular modems that might be the issue these days, many of them still use AT commands as those were the standard when GSM first appeared 30 odd years ago.

Perhaps adding something to indicate that ModemManager configuration is likely to need additional measures if problems still occur once brltty has been disabled?

tmiw:
As mentioned in today's meeting, go ahead and push a new commit with the suggested changes and I'll go ahead and merge.

PR #971, opened 2025-07-12
Work around issues when using Bluetooth devices on macOS.

tmiw:
This PR fixes various issues when using Bluetooth devices on macOS:

1. Audio device ordering was updated to start TX devices before RX devices (to allow the RX devices to change sample rates as needed, since Bluetooth devices downgrade to a lower sample rate for the speaker device when the mic device is present).
2. Buffer size calculation was fixed to actually use the next power of two. (The previous implementation was incorrect in certain cases).
3. Additional logging was added to help diagnose future problems with macOS audio.
4. RADE pipeline steps are returned directly by `FreeDVInterface` instead of going through `ParallelStep` (workaround as a `ParallelStep` bug is likely contributing).

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -28,7 +28,7 @@ jobs:
 
     - name: Spellcheck codebase
       shell: bash
-      run: codespell --ignore-words-list=caf,radae,rade,inout,nin,ontop,parm,tthe,ue `find src -name '*.c*' -o -name '*.h' | grep -v 3rdparty`
+      run: codespell --ignore-words-list=numer,caf,radae,rade,inout,nin,ontop,parm,tthe,ue `find src -name '*.c*' -o -name '*.h' | grep -v 3rdparty`
 
   dist:
     runs-on: ubuntu-22.04
--- a/.github/workflows/cmake-macos.yml
+++ b/.github/workflows/cmake-macos.yml
@@ -52,7 +52,7 @@ jobs:
           sqlite3 $HOME/Library/Application\ Support/com.apple.TCC/TCC.db "INSERT OR IGNORE INTO access VALUES ('kTCCServiceMicrophone','/opt/off/opt/runner/provisioner/provisioner',1,2,4,1,NULL,NULL,0,'UNUSED',NULL,0,1687786159);"
 
     - name: Workaround macOS permission issues
-      if: ${{ matrix.os == 'macos-latest' }}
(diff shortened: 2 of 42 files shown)


tmiw:
This does have impact outside of macOS so it would be nice if @Tyrbiter / @barjac could test. I'll watch the automated tests and see if those cause problems too.

Tyrbiter:
Running https://github.com/drowe67/freedv-gui/pull/971/commits/6dcf561717e600bbf0f0aa53cb2eb2c4e789fcbc now, if I see or hear anything odd I will report back, it certainly starts and appears to run normally, audio is going where it normally does.

barjac:
Likewise looks good here, no problems with audio interconnections on start-up all looks correct.
Will test live later today.

/OT
I have recently noticed that records in the reporter appear to randomly become selected when not clicked. I can't reproduce reliably yet, it's one of those 'how did that happen?' things, but it is definitely related to an unrelated mouse action.

Tyrbiter:
Still working here, decoding and reporting, can't see any problems from my end.

barjac:
> Still working here, decoding and reporting, can't see any problems from my end.

Yes same here, no issues at all. Busy net on 60m today with 8+ stations, usage is picking up!

tmiw:
One more quick sanity check before I generate new builds?

Tyrbiter:
I have this latest running now, but it seems GH doesn't like something being checked.

tmiw:
> I have this latest running now, but it seems GH doesn't like something being checked.

Yeah, I forgot to update a test file. Hopefully GH goes through this time.

Tyrbiter:
I have https://github.com/drowe67/freedv-gui/pull/971/commits/cb07fd3a7feb0465b8d708023af9a75c2a998742 running now, it seems GH checks have all passed.

Or maybe not now...

tmiw:
I just pushed some stuff to try to help with the test reliability a bit more. For instance, `rade_loss` was consistently failing on macOS when tests did fail, and I suspect it was because the way the test was implemented broke real-time safety.

Tyrbiter:
I have https://github.com/drowe67/freedv-gui/pull/971/commits/42e28f6c201c2e310fdfcdae8102a4c44f403a65 running, it seems to function normally, recent changes aimed at GH testing unsurprisingly don't seem to affect operation.

tmiw:
I'm going to go ahead and merge since tests are passing now. Going to verify a few things on Windows once GH finishes building master and then I'll send out another release candidate build.

PR #980, opened 2025-07-21
Add Mic/Speaker volume control to main window.

### LINKED ISSUE #952: Main Window with audio gain control
### Contact Details

DL2GRF, DJ2MO, DF6VI

### Feature for this project

Hello.
Would it be possible to make the filter window more adjustable?
It takes up a lot of screen space by aranging other tools (i.e. main/reporter) at one screen.

The easiest solution would be to add a simple audio gain control to the main window?

Thanks,

Stefan

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Non-standard compiler

None

### Implementation assistance

Testing

### Support assistance

supporting

tmiw:
Resolves #952 by adding a volume control to the main window for microphone/speaker volume, tied to the equivalent controls in the Filter dialog. These controls were also made so that they're always enabled (no need to enable the equalizer in the Filter dialog to use).

Note: an alternate solution (hide the graph on the Filter dialog to make it smaller) was originally agreed to by the PLT, but given additional user feedback, was switched to adding the additional control instead.

### DIFF
--- a/src/eq.cpp
+++ b/src/eq.cpp
@@ -63,6 +63,12 @@ std::shared_ptr<void> MainFrame::designAnEQFilter(const char filterType[], float
 
 void  MainFrame::designEQFilters(paCallBackData *cb, int rxSampleRate, int txSampleRate)
 {
+    // Volume can be adjusted via main window without enabling filters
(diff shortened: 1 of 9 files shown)


barjac:
Works perfectly here! Nice!

PR #990, opened 2025-07-30
Fix UI bugs when using wxWidgets 3.3.

tmiw:
This PR implements the following in order to allow FreeDV to work with wxWidgets 3.3:

* Cleans up startup/initial display warnings on console for various windows and dialogs.
* Fixes sizing issues with boxes on main window.

### DIFF
--- a/cmake/BuildWxWidgets.cmake
+++ b/cmake/BuildWxWidgets.cmake
@@ -1,4 +1,4 @@
-set(WXWIDGETS_VERSION "3.2.8")
+set(WXWIDGETS_VERSION "3.3.1")
 
 # Ensure that the wxWidgets library is staticly built.
 set(wxBUILD_SHARED OFF CACHE BOOL "Build wx libraries as shared libs")
--- a/cmake/wxWidgets-Direct2D-color-font.patch
+++ b/cmake/wxWidgets-Direct2D-color-font.patch
@@ -1,9 +1,9 @@
 diff --git a/include/wx/msw/setup.h b/include/wx/msw/setup.h
-index b1d20d927..4f834c941 100644
+index a1c003d00..6be2b0e4e 100644
 --- a/include/wx/msw/setup.h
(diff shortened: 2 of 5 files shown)


tmiw:
@barjac, the latest change should finally fix the rendering issues with the main window on 3.3. Let me know how that goes.

barjac:
Super! UI back to normal!

tmiw:
Hmm, how's performance for the FreeDV Reporter window with 3.3? I ask because there's apparently [an open issue for the control that window uses](https://github.com/wxWidgets/wxWidgets/issues/25025), albeit only on Windows. Maybe we can just have the UI fixes for Linux users in here and punt on making the macOS and Windows builds use it until it's a bit more stable?

barjac:
> Hmm, how's performance for the FreeDV Reporter window with 3.3?

I am going to do more testing of that as I think there are still issues.

I have tested this branch in Mageia9 (wxGTK-3.2) and I don't see any regressions there, neither do I see any of the issues to follow below on this Mageia9 X11 machine.

The problems I have been seeing with 3.3 in Mageia10 are in the reporter window, which I mentioned previously but could not capture in a screen shot when I tried. Mageia10 is on Wayland with xwayland which may also be a factor.

One issue is the random selecting of rows. Also its not easy to remove a selection if the window width is reduced such that there is no free space on the footer to click in.

The second is ghost rows appearing below all the real ones as the mouse is moved downwards over actual rows. Only the callsigns of existing real stations appear in the ghost rows under the correct column.

tmiw:
I created https://github.com/drowe67/freedv-gui/pull/993 for the FreeDV Reporter specific bugs that we're fixing, as I suspect they're still a problem in wxWidgets < 3.3 as well.

tmiw:
I did some checking on Windows and macOS with 3.3 and I didn't see anything weird going on, so I'm going to go ahead and merge. We can address the FreeDV Reporter window bugs in #993.

Issue #967, opened 2025-07-09
Tab positions below the waterfall are no longer remembered between re-starts of FreeDV


barjac:
In the past with FreeDV-1.x at some point the tab positions were saved between program runs.
Now that some tabs are not applicable to RADE mode it would be good if this feature was restored so the unused tabs could be moved out of view.

tmiw:
What are your thoughts about just removing the non-RADE tabs altogether? I'm not sure how often various tabs are used (other than From Mic and the waterfall, as the former is switched to for TX and the latter is basically the default on RX), so I'm not sure if anyone will miss those.

Also, the current save/load code is *very* hacky since it's basically redefining internal wxWidgets classes (hence why it's disabled by default). We can see about fixing this issue, but I don't think we'll be able to promote it out of "experimental" any time soon.

tmiw:
Hmm, I was able to get tab positioning to stick by doing the following:

1. Check Tools->Debugging->Enable Experimental Features, then restart FreeDV.
2. Move tab into its own "pane" (in this example, "From Mic":

<img width="997" height="951" alt="Image" src="https://github.com/user-attachments/assets/bf45f048-ed13-4272-ba3c-67c05ab24a08" />

3. Drag "From Mic" to where you want it to go (i.e. next to "Waterfall"):

<img width="997" height="951" alt="Image" src="https://github.com/user-attachments/assets/04e69d2d-8875-4a47-8612-741e3b709b9f" />

4. Restart FreeDV again. Note the preserved position of the "From Mic" tab and FreeDV starting with "Waterfall" selected:

<img width="997" height="951" alt="Image" src="https://github.com/user-attachments/assets/62918d8a-e75d-4080-bf85-00408f5dd55d" />

It seems like (2) is the critical step. Without doing that, the tab (re-)positioning doesn't stick across restarts. Unfortunately, I suspect it's a problem in internal wxWidgets code, specifically the `m_mgr.SavePerspective()` and `m_mgr.LoadPerspective()` calls we're doing (but would need to investigate more to know for sure).

barjac:
Ah OK, that is probably how I ended up with a strange order in my .freedv, from testing the original enhancement request to save the pane positions.
I will test that tonight as will be haymaking all day today.

Yes I think removing the unused tabs would be great if you feel that they will never be needed in future. Doing that would allow all the tabs to be visible at once, even in a squashed main window. That would make this 'issue' redundant. :)

\OT
[...]

PR #996, opened 2025-08-09
Move Spectrum plot averaging setting to Tools->Options.

tmiw:
Based on comments in #967, this PR moves the Spectrum plot averaging control to Tools->Options to free up some space. This was done as this control is not used often.


barjac:
Works fine here with no regressions seen. Much better spectrum display now :)

PR #975, opened 2025-07-19
Second pass at pipeline optimization.

tmiw:
This PR is a followup to #1002 to further improve performance/reduce audio dropouts and thus reduce automated test failures:

* `std::shared_ptr<>` references were removed in favor of `std::unique_ptr<>` in `src/pipeline` (to reduce the overhead when dereferencing memory during audio processing).
* Linux/macOS only: Use [mimalloc](https://github.com/microsoft/mimalloc?tab=readme-ov-file) for memory allocation to indirectly improve performance at the `librade` level without modifications to the latter.
* Use `-O3` for `libsamplerate` instead of `-O2`.
* Update/add `std::atomic` semantics:
    * Use `std::memory_order_acquire` / `std::memory_order_release` where appropriate.
    * Add additional use of `std::atomic` for some memory accesses (to avoid data races).
* Reduce context switching in Unit Test thread during test execution (by increasing delays to 100ms from 10-20ms).
* Set `PYTHONOPTIMIZE` environment variable to improve the performance of Python RADE code.
* Eliminate unneeded floating-point computations during audio pipeline execution.
* Use `GenericFIFO` instead of `codec2_fifo_*`. 
* FIFO size tuning:
    * Increase incoming audio FIFO lengths to avoid dropouts on Linux (discovered during FIFO size tuning).
    * Decrease some FIFO sizes due to low usage in testing (to improve cache utilization/thrashing).
* Add logic to delay `TxRxThread` instances until fully ready to execute.
* `TcpConnectionHandler` / FreeDV Reporter: reduce context switches by writing inbound socket data into FIFO.
* `rade_loss` test: 
    * Pre-process input files to avoid unnecessary resampling within FreeDV.
    * Use custom configuration to disable FreeDV Reporter reporting.
* macOS only: 
    * Use various OS-specific API calls to allow scheduler to properly schedule the FreeDV process.
    * Use PLT-approved minimum OS version when compiling virtual audio devices (used for GitHub testing).
    * Update `MacAudioDevice` to increment the internal semaphore by the number of "real-time" threads in system.
* GitHub Actions: Kill macOS processes/daemons known to use significant resources prior to test execution.

PR #997, opened 2025-08-10
Enable AGC functionality in Speex.

### LINKED ISSUE #988: [Feature Request] Automatic Gain Control on the microphone
### Contact Details

peterbmarks@gmail.com

### Feature for this project

I have difficulty controlling my microphone level and I sometimes hear other stations where mic level is too low or high.
A nice feature would be some sort of Automatic Gain Control or at least a peak limiter that would reduce gain if it is too high.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

I see some open source AGC libraries such as https://github.com/cpuimage/WebRTC_AGC/blob/master/agc.c 

### Support assistance

Testing

tmiw:
This PR resolves #988 by doing the following:

1. Adds AGC setting to Tools->Filter. This is enabled by default and will become disabled unless Speex is also enabled (as the latter is a prerequisite). The updated Filter dialog appears as follows:

image

2. Adds logic inside `SpeexStep` to follow changes to (1) in real-time and enable/disable AGC in the Speex library as appropriate.

### DIFF
--- a/src/config/FilterConfiguration.cpp
+++ b/src/config/FilterConfiguration.cpp
@@ -32,6 +32,7 @@ FilterConfiguration::FilterConfiguration()
     , codec2LPCPostFilterGamma("/Filter/codec2LPCPostFilter/Gamma", CODEC2_LPC_PF_GAMMA*100)
     , codec2LPCPostFilterBeta("/Filter/codec2LPCPostFilter/Beta", CODEC2_LPC_PF_BETA*100)
     , speexppEnable("/Filter/speexpp_enable", true)
+    , agcEnabled("/Filter/agcEnable", true)
     , enable700CEqualizer("/Filter/700C_EQ", true)
 {
     std::function<float(float)> gammaBetaSaveProcessor = [](float val) {
@@ -75,6 +76,7 @@ void FilterConfiguration::load(wxConfigBase* config)
     
     load_(config, speexppEnable);
     load_(config, enable700CEqualizer);
+    load_(config, agcEnabled);
 }
 
 void FilterConfiguration::save(wxConfigBase* config)
@@ -89,4 +91,5 @@ void FilterConfiguration::save(wxConfigBase* config)
     
     save_(config, speexppEnable);
     save_(config, enable700CEqualizer);
+    save_(config, agcEnabled);
 }
--- a/src/config/FilterConfiguration.h
+++ b/src/config/FilterConfiguration.h
@@ -89,6 +89,7 @@ public:
     ConfigurationDataElement<float> codec2LPCPostFilterBeta;
     
     ConfigurationDataElement<bool> speexppEnable;
+    ConfigurationDataElement<bool> agcEnabled;
     ConfigurationDataElement<bool> enable700CEqualizer;
     
     virtual void load(wxConfigBase* config) override;
--- a/src/gui/dialogs/dlg_filter.cpp
+++ b/src/gui/dialogs/dlg_filter.cpp
@@ -19,6 +19,8 @@
 //
 //==========================================================================
 
+#include <atomic>
+
 #include "dlg_filter.h"
 #include "gui/util/LabelOverrideAccessible.h"
 
@@ -52,7 +54,8 @@
 extern FreeDVInterface freedvInterface;
 extern wxConfigBase *pConfig;
 extern wxMutex g_mutexProtectingCallbackData;
-    
+extern std::atomic<bool> g_agcEnabled;
+
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
 // Class FilterDlg
 //-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=--=-=-=-=
@@ -107,7 +110,11 @@ FilterDlg::FilterDlg(wxWindow* parent, bool running, bool *newMicInFilter, bool 
     m_ckboxSpeexpp = new wxCheckBox(sb_speexpp, wxID_ANY, _("Speex Noise Suppression"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
     sbSizer_speexpp->Add(m_ckboxSpeexpp, 0, wxALL | wxALIGN_LEFT, 5);
     m_ckboxSpeexpp->SetToolTip(_("Enable noise suppression, dereverberation, AGC of mic signal"));
-
+    
+    m_ckboxAgcEnabled = new wxCheckBox(sb_speexpp, wxID_ANY, _("AGC"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
+    sbSizer_speexpp->Add(m_ckboxAgcEnabled, 0, wxALL | wxALIGN_LEFT, 5);
+    m_ckboxAgcEnabled->SetToolTip(_("Automatic gain control for microphone"));
+    
     m_ckbox700C_EQ = new wxCheckBox(sb_speexpp, wxID_ANY, _("700D/700E Auto EQ"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
     sbSizer_speexpp->Add(m_ckbox700C_EQ, 0, wxALL | wxALIGN_LEFT, 5);
     m_ckbox700C_EQ->SetToolTip(_("Automatic equalisation for FreeDV 700D/700E Codec input audio"));
@@ -248,6 +255,7 @@ FilterDlg::FilterDlg(wxWindow* parent, bool running, bool *newMicInFilter, bool 
     m_LPCPostFilterDefault->Connect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FilterDlg::OnLPCPostFilterDefault), NULL, this);
 
     m_ckboxSpeexpp->Connect(wxEVT_COMMAND_CHECKBOX_CLICKED, wxScrollEventHandler(FilterDlg::OnSpeexppEnable), NULL, this);
+    m_ckboxAgcEnabled->Connect(wxEVT_COMMAND_CHECKBOX_CLICKED, wxScrollEventHandler(FilterDlg::OnAgcEnable), NULL, this);
     m_ckbox700C_EQ->Connect(wxEVT_COMMAND_CHECKBOX_CLICKED, wxScrollEventHandler(FilterDlg::On700C_EQ), NULL, this);
 
     int events[] = {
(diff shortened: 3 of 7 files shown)


tmiw:
@peterbmarks  / @barjac / @Tyrbiter, it would be useful to get some testing here to see how well the Speex default settings work for FreeDV. 👍

w4mmp-ron:
Hi,
My initial observations.  
1) Before this AGC feature,  my Gain value was about -5.  Now the gain needs to be set to approximately 13.0 to make 0.4 peak to peak reading (0.4 to -0.4).  My understanding at this time that is the acceptable range, but this is based on K0PFX (Mel) has recommended. 
2) It would appear to me that the AGC is just a bit too aggressive based on how much I needed to increase the gain.  But not knowing how the AGC works internally,  this is simply a guess.

OK, this was transmitting into a dummy load to get some initial observations.  I will see what other stations have to say about my audio.  Lot's of SSB on 20M at the moment so I will need to wait until the preferred 14.2360MHz frequency is clear.  
Cheers,
Ron

tmiw:
There are a few settings in Speex for AGC, but I admit I don't really know how they work. I figure I'll play with them and see what's needed to get +/- 0.4 using gain = 0 and the microphone built into my laptop.

peterbmarks:
Just tested on Linux. I also had to increase the gain but I can see it is working and made it easier to keep mic level in range. 

Screenshot from 2025-08-11 09-34-16

tmiw:
I adjusted the "AGC level" parameter in speex and got it to show peaks +/- 0.4 with my MacBook Pro's built-in microphone without needing additional gain. Any better for you guys?

barjac:
Interesting.
I just spent an hour playing with this. :)

I think it has potential but at present the automatic gain range is too high for our use case.

The max. gain that it produces with no input is far too high such that the first word spoken is always very loud along with high background noise, after which the gain drops dramatically to a nice normal level.
We don't need such a large gain swing. The same detection and control with far less potential gain could work well.

I find that the GUI mic level control needs ~+4.5dB to produce a nice audio level after setting the mic sound card gain such that it produces no clipping with loud (much too high) audio.

I wonder if for testing/setting-up purposes a GUI window with controls for all the AGC block parameters could be made or 'borrowed' from some project? Experimental features?

Is the Speex noise reduction before or after the AGC? Just curious. ;)

g4dya:
I just tried out #4bd8b1d and first impression is that it's working quite well. It would be nice if the AGC and noise reduction were separately switchable as I prefer not to have anything in the signal path I don't need. 

@barjac one of the aims was to match loudness between stations. Enabling stations to twiddle agc parameters might defeat that.

w4mmp-ron:
Hi,
Just had a QSO with a station (W1FC).  He reported the audio is fine.  
This is when running 2.0.2-dev-2596.   I needed to reduce the microphone gain down to 4.8 from 13 which was expected with the changes in this release.  I will do more on air testing.
73,
Ron

tmiw:
Taking a look at what can be adjusted re: AGC:

```c
/** Set preprocessor Automatic Gain Control level (float) */
#define SPEEX_PREPROCESS_SET_AGC_LEVEL 6

/** Set maximal gain increase in dB/second (int32) */
#define SPEEX_PREPROCESS_SET_AGC_INCREMENT 26

/** Set maximal gain decrease in dB/second (int32) */
#define SPEEX_PREPROCESS_SET_AGC_DECREMENT 28

/** Set maximal gain in dB (int32) */
#define SPEEX_PREPROCESS_SET_AGC_MAX_GAIN 30

/** Set preprocessor Automatic Gain Control level (int32) */
#define SPEEX_PREPROCESS_SET_AGC_TARGET 46
```

So far, I've been adjusting `SPEEX_PREPROCESS_SET_AGC_LEVEL` (which sounds like it might be equivalent to `SPEEX_PREPROCESS_SET_AGC_TARGET`). Based on looking at speex source code, it looks like the defaults are:

* AGC Level: 8000 (EDIT: this was what was increased to get +/- 0.4 on my end. Currently 21000, the maximum that speexdsp apparently allows is 32767.)
* AGC Increment: 12 dB/sec
* AGC Decrement: 40 dB/sec
* Max gain: 30 dB

@barjac, do you think that AGC Decrement and/or Increment should be adjusted based on the feedback you previously provided? Regardless, there's going to be at least a little bit of delay before AGC actually kicks in.

@g4dya, I actually think noise reduction should happen along with AGC. Otherwise, noise ends up getting taken into account, which means we very well could not be getting the most benefit from AGC.

w4mmp-ron:
Hi,
Just had another QSO (K0PFX).  The good news is I can move the microphone away from my mouth and the audio level remains good.  This is until the distance is so far that the audio almost drops out.  This is much better than before the AGC  implementation.  Prior to the change only a small change in the distance would make a large change in the audio level.  He reports the audio quality is fine but the audio volume spikes just a bit and I see that on the Frm Mic display. 
73,
Ron

tmiw:
I bumped up AGC Level to 24000 and decreased max gain to 20 dB as of the latest commit. Let me know how this goes? If this makes things worse we can just merge this PR with the previous settings.

peterbmarks:
I would add that I've been using it on the Australia net and reports of my audio have been excellent.

g4dya:
I've been using 72edd47f and my impression is that the first word spoken is too loud but that it quickly settles down to a good level - i.e. 90% of peaks within about ±0.4 on the 'from mic' graph.  Would it be possible to make the agc use unity gain as its starting point?

Tyrbiter:
Would it make sense to merge this into https://github.com/drowe67/freedv-gui/pull/975 just so that we can all-up test the current bleeding edge?

tmiw:
> I've been using [72edd47](https://github.com/drowe67/freedv-gui/commit/72edd47f53747130f54a2bf52d5ca3cb91a13524) and my impression is that the first word spoken is too loud but that it quickly settles down to a good level - i.e. 90% of peaks within about ±0.4 on the 'from mic' graph. Would it be possible to make the agc use unity gain as its starting point?

I've been looking at speexdsp and I wonder if we need to bump `AGC_LEVEL` back down a bit. The downside of course being that people would then need to apply some level of positive mic gain to get back up to +/- 0.4. 

Question: is 72edd47 any better or worse than 4bd8b1d in terms of the first word issue?

> Would it make sense to merge this into #975 just so that we can all-up test the current bleeding edge?

This might be better to do by merging this PR. Let's get the levels dialed in first, I think.

tmiw:
> > I've been using [72edd47](https://github.com/drowe67/freedv-gui/commit/72edd47f53747130f54a2bf52d5ca3cb91a13524) and my impression is that the first word spoken is too loud but that it quickly settles down to a good level - i.e. 90% of peaks within about ±0.4 on the 'from mic' graph. Would it be possible to make the agc use unity gain as its starting point?
> 
> I've been looking at speexdsp and I wonder if we need to bump `AGC_LEVEL` back down a bit. The downside of course being that people would then need to apply some level of positive mic gain to get back up to +/- 0.4.
> 
> Question: is [72edd47](https://github.com/drowe67/freedv-gui/commit/72edd47f53747130f54a2bf52d5ca3cb91a13524) any better or worse than [4bd8b1d](https://github.com/drowe67/freedv-gui/commit/4bd8b1d392f6c37863f89eab1b9ac200680828a6) in terms of the first word issue?

> Would it make sense to merge this into #975 just so that we can all-up test the current bleeding edge?

This might be better to do by merging this PR. Let's get the levels dialed in first, I think.

g4dya:
I don't know which of the two commits is better or worse in terms of the first word issue. The difference, if any, is quite small. In both cases I think the first word was 6 dB or more louder than what followed.

Would it be possible to reset the AGC_LEVEL to 8000 and, only when AGC is enabled, put in some fixed gain after the AGC so the user doesn't have to?

tmiw:
> I don't know which of the two commits is better or worse in terms of the first word issue. The difference, if any, is quite small. In both cases I think the first word was 6 dB or more louder than what followed.
> 
> Would it be possible to reset the AGC_LEVEL to 8000 and, only when AGC is enabled, put in some fixed gain after the AGC so the user doesn't have to?

Sure, but I'm not sure we can find a fixed gain level that would be appropriate for all setups. That said, so far people seemed to be needing to add ~10 dB with `AGC_LEVEL` at 8000 (and at the current level the amount seemed to be closer to 4 or so).

g4dya:
AIUI 8000 is the libspeexdsp AGC_LEVEL default. Assuming it's a linear relationship, 24000/8000 = 9.5 dB, so 10 dB make-up gain makes sense.

If people are sending low microphone level into freedv-gui then a higher AGC_LEVEL would make sense. But is that special case the default we should optimise for? If so, it will be sub-optimal for users sending a 'correct' level let alone excessive level, leading to the loud first word issue. When I did my tests I carefully adjusted the audio level for -20 dBFS RMS and -8 dBFS true peak, corresponding to 40% peak amplitude (and good audio practice). (On linux there are several very good Jack metering tools which give chapter and verse. In this case I used jkmeter.)

Would about 10 dB of make-up gain after the AGC be a problem?

tmiw:
> AIUI 8000 is the libspeexdsp AGC_LEVEL default. Assuming it's a linear relationship, 24000/8000 = 9.5 dB, so 10 dB make-up gain makes sense.
> 
> If people are sending low microphone level into freedv-gui then a higher AGC_LEVEL would make sense. But is that special case the default we should optimise for? If so, it will be sub-optimal for users sending a 'correct' level let alone excessive level, leading to the loud first word issue. When I did my tests I carefully adjusted the audio level for -20 dBFS RMS and -8 dBFS true peak, corresponding to 40% peak amplitude (and good audio practice). (On linux there are several very good Jack metering tools which give chapter and verse. In this case I used jkmeter.)
> 
> Would about 10 dB of make-up gain after the AGC be a problem?

Not sure, but another option is to reset `AGC_LEVEL` back to 8000 and increase `AGC_MAX_GAIN` by 10 dB instead. That should effectively add the additional 10 dB for those setups that need it (even if it takes a word or two). I'll see how well that works locally before pushing that change.

tmiw:
Unfortunately adding a fixed gain immediately after the AGC produces a similar initial spike as just having `AGC_LEVEL` set to a high level. I went ahead and adjusted the level down to 12000 (and max gain to 40 dB) since that seemed to produce okay results when a microphone is properly adjusted using the OS volume controls.

g4dya:
Just tried 4bc2f35 on this afternoon's UK 5MHz net, which I recorded and listened to afterwards. With the AGC_LEVEL reduced to 12000 the first word problem seems to have lessened. I needed 8 dB of gain to reach ±0.4, and then the playback audio output was close to -20 LUFS ( = -20dBFS RMS loudness weighted as per Rec. ITU-R BS.1770) which I consider about optimum.

Given that we're seeing different results, it might be time to merge this into master for wider testing.

tmiw:
> Just tried [4bc2f35](https://github.com/drowe67/freedv-gui/commit/4bc2f355bdf93631bfe05558d102d980755a4a11) on this afternoon's UK 5MHz net, which I recorded and listened to afterwards. With the AGC_LEVEL reduced to 12000 the first word problem seems to have lessened. I needed 8 dB of gain to reach ±0.4, and then the playback audio output was close to -20 LUFS ( = -20dBFS RMS loudness weighted as per Rec. ITU-R BS.1770) which I consider about optimum.
> 
> Given that we're seeing different results, it might be time to merge this into master for wider testing.

Fair enough. We can always have a followup PR if needed.

w4mmp-ron:
Hello,
Here are my findings when running 2.0.2-dev-d55e.   I need to increase the FreeDV Mic gain to 10 to achieve a +/-4 on the Frm Mic display.  The AGC does appear to hold the audio level to a consistent level when changing the position of the microphone.   I need to perform real world testing to get feed back on the quality of the audio.
73,
Ron

PR #999, opened 2025-08-10
Linux: Search for and list serial devices from /dev/serial for PTT config.

tmiw:
Based on feedback from the mailing list, this PR adds serial devices from /dev/serial to the drop-down list in both the Easy Setup and CAT & PTT Config dialogs. Example:

Screenshot 2025-08-10 at 4 04 49 PM


tmiw:
@barjac / @Tyrbiter, think `/dev/tty*` should be removed from the above list in favor of just `/dev/serial` entries? Or is there still value in showing the former?

Tyrbiter:
I'd say leave the /dev/tty* for now, a lot of radios don't provide useful info in the /dev/serial/* tree anyway although it is good for those that do.

barjac:
What causes that list to be displayed? I see nothing like that in #ad22b

tmiw:
> What causes that list to be displayed? I see nothing like that in #ad22b

You can get to that list by going to Tools->CAT and PTT Config or Tools->Easy Setup. Either way, this is when configuring the serial port to use for Hamlib.

> I'd say leave the /dev/tty* for now, a lot of radios don't provide useful info in the /dev/serial/* tree anyway although it is good for those that do.

Fair enough. What about `/dev/serial/by-path`? Any value in that vs. `by-id`?

Tyrbiter:
> Fair enough. What about `/dev/serial/by-path`? Any value in that vs. `by-id`?

by-path is less useful here than by-id so perhaps just leave the by-path entries out for now and see if people complain.

tmiw:
> > Fair enough. What about `/dev/serial/by-path`? Any value in that vs. `by-id`?
> 
> by-path is less useful here than by-id so perhaps just leave the by-path entries out for now and see if people complain.

OK, removed. I'll make sure it builds then merge. Thanks!

PR #1002, opened 2025-08-17
Various GUI and FreeDV Reporter optimizations

tmiw:
This PR ports over the GUI and FreeDV Reporter optimizations done during #975 development for individual review. This is being done as (a) these changes have value individually but do not appear to impact ctest pass rates (the primary goal of the previously mentioned PR) and (b) to reduce the scope of the previous PR.

Changes made:

* `PlotScalar` (used for "Mic In" plot during TX): Use different algorithm for drawing lines to reduce the amount of drawing required.
* `PlotWaterfall`: Cache previously drawn waterfall slices to reduce the amount of drawing required.
* FreeDV Reporter:
    * Only trigger redraw if content of a row (excluding "Last Update") has changed *and* if the window is visible.
    * Remove inheritance from `ThreadedObject` in favor of using `std::mutex` to enforce access control. (Reduces overhead from unnecessary threads.)
* `TcpConnectionHandler`: 
    * Remove unnecessary timer in favor of long-running thread to handle data RX.
    * Reduce number of events enqueued from RX thread to main object thread to reduce context switching.
* `ThreadedObject`:
    * (macOS-only) Use Grand Central Dispatch instead of `std::thread` to take advantage of existing threads that are auto-created.
    * Feature enhancement: Allow objects to run in the context of parent `ThreadedObject` instances. (Currently unused.)
* `ThreadedTimer`: add `restart()` method to prevent repeatedly destroying and recreating threads.

### DIFF
--- a/src/gui/controls/plot_scalar.cpp
+++ b/src/gui/controls/plot_scalar.cpp
@@ -23,6 +23,8 @@
 #include <wx/wx.h>
 #include <wx/graphics.h>
 
+#include <map>
+
 #include "plot_scalar.h"
 
 BEGIN_EVENT_TABLE(PlotScalar, PlotPanel)
@@ -57,6 +59,7 @@ PlotScalar::PlotScalar(wxWindow* parent,
 
     m_rCtrl = GetClientRect();
 
+    lineMap_ = nullptr;
     m_channels = channels;
     m_t_secs = t_secs;
     m_sample_period_secs = sample_period_secs;
@@ -87,6 +90,7 @@ PlotScalar::PlotScalar(wxWindow* parent,
 PlotScalar::~PlotScalar()
 {
     delete[] m_mem;
+    delete[] lineMap_;
 }
 
 //----------------------------------------------------------------
@@ -148,7 +152,6 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
     float index_to_px;
     float a_to_py;
     int   i;
-    int   prev_x, prev_y;
     float a;
 
     m_rCtrl = GetClientRect();
@@ -166,7 +169,9 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
 
     plotWidth = m_rGrid.GetWidth();
     plotHeight = m_rGrid.GetHeight();
-        
+
+    if (plotWidth <= 0 || plotHeight <= 0) return;
+ 
     wxBrush ltGraphBkgBrush = wxBrush(BLACK_COLOR);
     ctx->SetBrush(ltGraphBkgBrush);
     ctx->SetPen(wxPen(BLACK_COLOR, 0));
@@ -180,11 +185,21 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
     pen.SetWidth(1);
     ctx->SetPen(pen);
     
-    // draw all samples
-
-    prev_x = prev_y = 0; // stop warning
-
     // plot each channel     
+
+    // x -> (y1, y2)
+    if (lineMap_ != nullptr)
+    {
+        lineMap_ = new MinMaxPoints[plotWidth + PLOT_BORDER + XLEFT_OFFSET];
+        assert(lineMap_ != nullptr);
+    }
+
+    for (int index = 0; index < plotWidth + PLOT_BORDER + XLEFT_OFFSET; index++)
+    {
+        lineMap_[index].y1 = INT_MAX;
+        lineMap_[index].y2 = INT_MIN;
+    }
+
     int offset, x, y;
     for(offset=0; offset<m_channels*m_samples; offset+=m_samples) {
 
@@ -243,16 +258,25 @@ void PlotScalar::draw(wxGraphicsContext* ctx)
             else {
                 if (i)
                 {
-                    wxGraphicsPath path = ctx->CreatePath();
-                    path.MoveToPoint(x, y);
-                    path.AddLineToPoint(prev_x, prev_y);
-                    ctx->StrokePath(path);
+                    auto item = &lineMap_[x];
+                    item->y1 = std::min(item->y1, y);
+                    item->y2 = std::max(item->y2, y);
                 }
-                prev_x = x; prev_y = y;
             }
         }
     }
-    
+   
+    if (!m_bar_graph)
+    {
+        for (int index = 0; index < plotWidth + PLOT_BORDER + XLEFT_OFFSET; index++)
+        {
+            auto item = &lineMap_[index];
+            if (item->y1 == INT_MAX || item->y2 == INT_MIN) continue;
+
+            ctx->StrokeLine(index, item->y1, index, item->y2);
+        }
+    } 
+
     drawGraticule(ctx);
 }
 
@@ -273,10 +297,6 @@ void PlotScalar::drawGraticule(wxGraphicsContext* ctx)
     int plotWidth = m_rGrid.GetWidth();
     int plotHeight = m_rGrid.GetHeight();
 
-    wxBrush ltGraphBkgBrush;
-    ltGraphBkgBrush.SetStyle(wxBRUSHSTYLE_TRANSPARENT);
-    ltGraphBkgBrush.SetColour(*wxBLACK);
-    ctx->SetBrush(ltGraphBkgBrush);
     ctx->SetPen(wxPen(BLACK_COLOR, 1));
 
     wxGraphicsFont tmpFont = ctx->CreateFont(GetFont(), GetForegroundColour());
@@ -356,6 +376,16 @@ void PlotScalar::clearSamples()
 //----------------------------------------------------------------
 void PlotScalar::OnSize(wxSizeEvent& event)
 {
+    m_rCtrl = GetClientRect();
+    m_rGrid = m_rCtrl;
+    if (!m_mini)
+        m_rGrid = m_rGrid.Deflate(PLOT_BORDER + (XLEFT_OFFSET/2), (PLOT_BORDER + (YBOTTOM_OFFSET/2)));
+
+    int plotWidth = m_rGrid.GetWidth();
+    delete[] lineMap_;
+
+    lineMap_ = new MinMaxPoints[plotWidth];
+    assert(lineMap_ != nullptr);
 }
 
 //----------------------------------------------------------------
--- a/src/gui/controls/plot_scalar.h
+++ b/src/gui/controls/plot_scalar.h
@@ -53,7 +53,16 @@ class PlotScalar: public PlotPanel
          void setLogY(int logy) { m_logy = logy; }
 
          void clearSamples();
-         
+        
+    private:
+        struct MinMaxPoints
+        {
+            int y1;
+            int y2;
+        };
+
+        MinMaxPoints* lineMap_;
+ 
     protected:
 
          int      m_channels;
--- a/src/gui/controls/plot_waterfall.cpp
+++ b/src/gui/controls/plot_waterfall.cpp
@@ -79,7 +79,6 @@ PlotWaterfall::PlotWaterfall(wxWindow* parent, bool graticule, int colour): Plot
 
     m_max_mag = MAX_MAG_DB;
     m_min_mag = MIN_MAG_DB;
-    m_fullBmp = NULL;
     sync_ = false;
 }
 
@@ -87,8 +86,6 @@ PlotWaterfall::PlotWaterfall(wxWindow* parent, bool graticule, int colour): Plot
 // we plot in and allocate a bit map of the correct size
 void PlotWaterfall::OnSize(wxSizeEvent& event) 
 {
-    delete m_fullBmp;
-    
     // resize bit map
 
     m_rCtrl  = GetClientRect();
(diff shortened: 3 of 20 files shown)


tmiw:
I tried to get some performance numbers on Linux to see how much this actually helped but unfortunately since CPU usage in `top` for the `freedv` process has a habit of spiking it's hard to tell. Hopefully @barjac and @Tyrbiter can tell this actually improved stuff on that platform.

Tyrbiter:
I've been using PR #975 in its various guises over a week or more, and it definitely improves the gui aspect and the plotting is smoother too.

I now have this new build running, so will report what I find in a few days. It would be nice if the build checks all passed, but somehow macOS seems cursed.

tmiw:
> I've been using PR #975 in its various guises over a week or more, and it definitely improves the gui aspect and the plotting is smoother too.

~~There's one more GUI related change that I could bring over from the other PR but that's not a simple `git checkout [branch] [file]` (and I'm not sure if it's relevant). I guess we'll find out.~~

EDIT: now brought in.

> I now have this new build running, so will report what I find in a few days. It would be nice if the build checks all passed, but somehow macOS seems cursed.

I'm beginning to suspect that the only real way to fix the macOS failures is to have a full C port of `librade` (and make sure we're not making calls like `malloc` during encode/decode to ensure real-time safety). Though `fullduplex_RADEV1` seems mostly solid on macOS, so I'm really not sure why the `rade_loss` test causes the macOS version to intermittently lose sync.

tmiw:
~~Note to self: waterfall RADE signal seems off-center compared to `master`. Need to investigate/fix before merge.~~

EDIT: fixed.

barjac:
Re. the Mic/Spkr gain slider.
Middle mouse click to reset whichever is in use to 0dB would be useful.

Tyrbiter:
> ~Note to self: waterfall RADE signal seems off-center compared to `master`. Need to investigate/fix before merge.~
> 
> EDIT: fixed.

I don't know if this is relevant, but note that due to the sub-carrier number and spacing when the RADE code starts up this is in stdout:

Input BPF bandwidth: 1740.000244 centre: 1475.000000

There is a 25Hz offset relative to the nominal 1500Hz centre frequency, I don't know if this might have affected the look of the waterfall relative to the legacy modes.

tmiw:
> > ~Note to self: waterfall RADE signal seems off-center compared to `master`. Need to investigate/fix before merge.~
> > EDIT: fixed.
> 
> I don't know if this is relevant, but note that due to the sub-carrier number and spacing when the RADE code starts up this is in stdout:
> 
> Input BPF bandwidth: 1740.000244 centre: 1475.000000
> 
> There is a 25Hz offset relative to the nominal 1500Hz centre frequency, I don't know if this might have affected the look of the waterfall relative to the legacy modes.

I think this is a holdover from the legacy modes. The main issue was that the width used for calculating the waterfall slice was different than the viewport, if that makes sense. Needed to resize the generated slice so that the width is the same.

tmiw:
> Re. the Mic/Spkr gain slider. Middle mouse click to reset whichever is in use to 0dB would be useful.

Try the latest? I'm not able to test at the moment (no middle button on laptop) but the changes *should* work.

Tyrbiter:
> > Re. the Mic/Spkr gain slider. Middle mouse click to reset whichever is in use to 0dB would be useful.
> 
> Try the latest? I'm not able to test at the moment (no middle button on laptop) but the changes _should_ work.

Could you merge this into PR #975 as well please.

tmiw:
> > > Re. the Mic/Spkr gain slider. Middle mouse click to reset whichever is in use to 0dB would be useful.
> > 
> > 
> > Try the latest? I'm not able to test at the moment (no middle button on laptop) but the changes _should_ work.
> 
> Could you merge this into PR #975 as well please.

Done.

barjac:
> > Re. the Mic/Spkr gain slider. Middle mouse click to reset whichever is in use to 0dB would be useful.
> 
> Try the latest? I'm not able to test at the moment (no middle button on laptop) but the changes _should_ work.

No its not working as expected but it is doing something ;)
In RX or TX placing the mouse pointer to the right of the current marker and clicking the middle button increments the gain by 4dB and to the left it decrements by 4dB irrespective of the current setting, always 4dB.

tmiw:
> > > Re. the Mic/Spkr gain slider. Middle mouse click to reset whichever is in use to 0dB would be useful.
> > 
> > 
> > Try the latest? I'm not able to test at the moment (no middle button on laptop) but the changes _should_ work.
> 
> No its not working as expected but it is doing something ;) In RX or TX placing the mouse pointer to the right of the current marker and clicking the middle button increments the gain by 4dB and to the left it decrements by 4dB irrespective of the current setting, always 4dB.

How about now?

Tyrbiter:
Running https://github.com/drowe67/freedv-gui/pull/1002/commits/1120ee6e864a924b08f88f545e1fbcbde6e149aa now, the mix/spkr slider seems to work as expected now.

barjac:
Running #1120 it's 'sort of' working but far too picky on where to middle click in the widget.

It just took me 10 mins to find the _really_ tiny area where this works to centre the slider and it is just above the slider knob circle but below the bottom of the p in Spkr.

That equates to a horizontal band around 1/16" high (red in the image below) on my screen! Users will not easily find it.

Screenshot_20250820_110021

EDIT: I just found another area the same size but equi-spaced below the slider, but still rather difficult to find. Maybe centre click on the digits would be easier?

Tyrbiter:
> Running #1120 it's 'sort of' working but far too picky on where to middle click in the widget.
> 
> It just took me 10 mins to find the _really_ tiny area where this works to centre the slider and it is just above the slider knob circle but below the bottom of the p in Spkr.
> 
> That equates to a horizontal band around 1/16" high (red in the image below) on my screen! Users will not easily find it.
> Screenshot_20250820_110021

Odd, here I can middle click anywhere on the slider or the slider button and in much of the area above and below it.

No idea why this is so different between one DE and another, although I don't have wxWindows 3.3.x in Fedora 42 yet.

barjac:
If I middle click on the slider (circle) it does nothing. If I middle click on the slider track it moves 4dB towards where I click. Only the red area above the slider and similar area below causes the re-centre.
This is in Mageia 10 with Wayland/xwayland and wxGTK-3.3.1.

I just tested this in Mageia 9 with X11 and wxGTK-3.2.8.1 and its the same as @Tyrbiter reports which is fine.

tmiw:
I moved the middle-click handler to the parent box instead of the slider, hopefully that works better for @barjac.

Tyrbiter:
> I moved the middle-click handler to the parent box instead of the slider, hopefully that works better for @barjac.

This makes things a bit odd for Fedora 42 now, I get the 4dB step behaviour @barjac described when I mouse over the slider and scroll the wheel, the only way of getting the 0dB reset is to middle click over the box holding the centred dB value below the slider.

Although different, this isn't necessarily bad behaviour as long as it is made clear in the manual, however it's odd that this should change so differently between Fedora and Mageia with different wxWidget versions and perhaps different Wayland/Xwayland packages.

Perhaps trying to work out what relevant libraries each of us has will help. There are so many \*ayland\* packages on my system that I don't know where to start really.

tmiw:
> This makes things a bit odd for Fedora 42 now, I get the 4dB step behaviour @barjac described when I mouse over the slider and scroll the wheel, the only way of getting the 0dB reset is to middle click over the box holding the centred dB value below the slider.

Confirmed on my end in Fedora too. I also noticed that spinning the scroll wheel increases/decreases by 6 dB for me (and 4 when just pushing down on the wheel). This behavior is pretty similar in sliders in other Linux apps, right?

Tyrbiter:
> > This makes things a bit odd for Fedora 42 now, I get the 4dB step behaviour @barjac described when I mouse over the slider and scroll the wheel, the only way of getting the 0dB reset is to middle click over the box holding the centred dB value below the slider.
> 
> Confirmed on my end in Fedora too. I also noticed that spinning the scroll wheel increases/decreases by 6 dB for me (and 4 when just pushing down on the wheel). This behavior is pretty similar in sliders in other Linux apps, right?

I get 4dB steps with wheel pressed or not, so that's a difference for some unknown reason. As for other Linux apps, I've never noticed this sort of behaviour but that's not to say it doesn't exist.

barjac:
Trying the latest @ #6f63 in Mageia 10 it's very flakey, in that it sometimes takes anything up to ~8 clicks in different places to get the re-centre to work, other times its first time. Seems to sometimes prefer the top and bottom edges of the box rather than inside it but nothing seems to be reliably repeatable now. I will try in Mageia 9 tomorrow. Previously the narrow strips where it worked were reliable, but just too narrow. If a specific place like over the xx.xdB text could be used, it may be less unpredictable.

barjac:
In Mageia 9 #6f63 seems perfect now as long as you know that middle click works anywhere in the widget box **below** the slider.

Tyrbiter:
> In Mageia 9 #6f63 seems perfect now as long as you know that middle click works anywhere in the widget box **below** the slider.

This smells like a wxWidgets 3.3.x bug to be honest, I'm quite pleased it hasn't appeared for Fedora yet.

barjac:
There is a minor issue with the reporter Msg column display in the table.
When a message is longer than the box by up to ~3 characters the elipsis is shown but the tooltip is not generated.
Maybe the elipsis presence should enable the tooltip on mouseover rather than whatever does it now?

barjac:
> > In Mageia 9 #6f63 seems perfect now as long as you know that middle click works anywhere in the widget box **below** the slider.
> 
> This smells like a wxWidgets 3.3.x bug to be honest, I'm quite pleased it hasn't appeared for Fedora yet.

No this is in in Mageia 9 with X11 and wxGTK-3.2.8.1, not Cauldron with 3.3.x.

Tyrbiter:
> > > In Mageia 9 #6f63 seems perfect now as long as you know that middle click works anywhere in the widget box **below** the slider.
> > 
> > 
> > This smells like a wxWidgets 3.3.x bug to be honest, I'm quite pleased it hasn't appeared for Fedora yet.
> 
> No this is in in Mageia 9 with X11 and wxGTK-3.2.8.1, not Cauldron with 3.3.x.

It was the problems in Cauldron you commented on that I meant @barjac , like Fedora 42 with wxGTK-3.2.8 your Mageia 9 machine seems to be behaving well.

tmiw:
Middle-click is now set to key off of the dB figure under the slider to preserve OS/DE behavior of the slider when using the mouse wheel.

> There is a minor issue with the reporter Msg column display in the table. When a message is longer than the box by up to ~3 characters the elipsis is shown but the tooltip is not generated. Maybe the elipsis presence should enable the tooltip on mouseover rather than whatever does it now?

We're comparing current column size vs. the calculated size of the text in the column, which isn't perfect since there are cases where the column only clips like 1-2 characters (but the ... is enough to make the comparison fail). I'll need to think about this some more.

Tyrbiter:
Sorry @tmiw this https://github.com/drowe67/freedv-gui/pull/1002/commits/705083f9e83c36d1a032d36b0b6a22e9fc48582c doesn't give me any middle-click action at all.

tmiw:
> Sorry @tmiw this [705083f](https://github.com/drowe67/freedv-gui/commit/705083f9e83c36d1a032d36b0b6a22e9fc48582c) doesn't give me any middle-click action at all.

Yeah, I just went ahead and reverted back to having the handler be on the slider for now. I'm not sure why there's differences between each environment but maybe we need to consider some sort of other way of enabling easy reset to 0. (Or maybe see if #1009 eliminates the need for this once it merges.)

barjac:
> > Sorry @tmiw this [705083f](https://github.com/drowe67/freedv-gui/commit/705083f9e83c36d1a032d36b0b6a22e9fc48582c) doesn't give me any middle-click action at all.
> 
> Yeah, I just went ahead and reverted back to having the handler be on the slider for now. I'm not sure why there's differences between each environment but maybe we need to consider some sort of other way of enabling easy reset to 0. (Or maybe see if #1009 eliminates the need for this once it merges.)

I have just had another play with Mageia 10 and I have discovered where the reset works reliably with #6f63. Its in an area of each corner of the widget about 3/16" square - not bad to use once you know, but obviously different to wxGTK-3.2.
I'm not keen on #1009 forcing 1dB steps, I think 0.5dB would be more reasonable. I'm finding the AGC quite useful, now I have other settings appropriate to it. Most reports seem in favour that I have heard.

barjac:
> > There is a minor issue with the reporter Msg column display in the table. When a message is longer than the box by up to ~3 characters the elipsis is shown but the tooltip is not generated. Maybe the elipsis presence should enable the tooltip on mouseover rather than whatever does it now?
> 
> We're comparing current column size vs. the calculated size of the text in the column, which isn't perfect since there are cases where the column only clips like 1-2 characters (but the ... is enough to make the comparison fail). I'll need to think about this some more.

If the elipsis is there then create the tooltip OR simply always create it?

tmiw:
> > > There is a minor issue with the reporter Msg column display in the table. When a message is longer than the box by up to ~3 characters the elipsis is shown but the tooltip is not generated. Maybe the elipsis presence should enable the tooltip on mouseover rather than whatever does it now?
> > 
> > 
> > We're comparing current column size vs. the calculated size of the text in the column, which isn't perfect since there are cases where the column only clips like 1-2 characters (but the ... is enough to make the comparison fail). I'll need to think about this some more.
> 
> If the elipsis is there then create the tooltip OR simply always create it?

The ellipsis is actually added internally by wxWidgets during rendering and we don't have access to information indicating that this was done. :( Probably the best we can do is start snipping characters from the end and measuring string lengths including the ellipsis until we get a length ~equal to the column length.

EDIT: I just went ahead and updated the logic to always show the tooltip.

tmiw:
Going to go ahead and merge for now; we can create a followup PR if needed.

PR #1008, opened 2025-08-19
change stepsizes gain and frequency

dl1bz:
- change steps in gain sliders to 1db (was 0.1db)
- change frequency steps in EQ to 100Hz (was 1Hz)

0.1db in AF level make not much sense, the same with 1Hz frequency steps in EQ
Result:
- better GUI experience with useful vales and steps
- make repeatable values possible (no mouse lottery)

The second problem are Locales:
If macOS (German version here) I get an error windows if I start FreeDV GUI (no locale for Country Germany)
If Linux (locale set to de_DE@UTF8) the hamlib sends now the frequency with 7,xxxx instead of 7.xxxx

So I removed the extra wxWidget settings for locales and use POSIX instead. That fix both problems. If you not translate the app then locales make no sense (except the date/time display in the reporter windows). But anywhere, better we use POSIX globally.

ps: sorry for my PR correction, there were two other PR included, but they need droped here and had nothing to do with my two changes. I forgot to change to other branch before I commit.

73 Heiko, DL1BZ


barjac:
> * change steps in gain sliders to 1db (was 0.1db)
> 
>     * change frequency steps in EQ to 100Hz (was 1Hz)
> 
> 
> 0.1db in AF level make not much sense, the same with 1Hz frequency steps in EQ Result:
> 
>     * better GUI experience with useful vales and steps
> 
>     * make repeatable values possible (no mouse lottery)
> 
> 
> The second problem are Locales: If macOS (German version here) I get an error windows if I start FreeDV GUI (no locale for Country Germany) If Linux (locale set to de_DE@UTF8) the hamlib sends now the frequency with 7,xxxx instead of 7.xxxx
> 
> So I removed the extra wxWidget settings for locales and use POSIX instead. That fix both problems. If you not translate the app then locales make no sense (except the date/time display in the reporter windows). But anywhere, better we use POSIX globally.
> 
> ps: sorry for my PR correction, there were two other PR included, but they need droped here and had nothing to do with my two changes. I forgot to change to other branch before I commit.
> 
> 73 Heiko, DL1BZ

tmiw:
Oops, I didn't see that comment above. Thanks @barjac! 

That said, I seem to remember there being logic in Hamlib to temporarily switch to POSIX locale, so that still doesn't seem necessary to do here. Maybe there's a bug in that library? Might need further investigation.

dl1bz:
I had a few problems with this PR - it's not complete, sri.
The problem with the locales:
With macOS I got a error msg if FreeDV starts, which means, no locales for Germany (using macOS with German language).
With Linux I had the problem, if I select a frequency in FreeDV (using hamlib) the freq will send as 7,xxx and not 7.xxx. So the TRX CAT produces wrong results. All my Linux systems has de_DE@UTF-8 as Locales per default !

So I need to resend this PR. The code was incomplete, not all commits were included.
Sorry for that, my mistake.

73 Heiko, DL1BZ

PR #1009, opened 2025-08-19
change stepsize gain and frequency sliders, add 3db more TX Mod Level, add a Limiter for Audio Playback

dl1bz:
Here again my PR - now with all changes. Sorry again for my mistakes with the PR I made before.

- change steps in gain sliders to 1db (was 0.1db)
- change frequency steps in EQ to 100Hz (was 1Hz)
- add 20db more for TX Mod Level (and rename TX Attenuation to Mod Level TX), because there are devices where 0db is not enough AF level for input -> add 20db more now

0.1db in AF level make not much sense, the same with 1Hz frequency steps in EQ
Result:

- better GUI experience with useful vales and steps
- make repeatable values possible (no mouse lottery)

The second problem are Locales:
If macOS (German version here) I get an error windows if I start FreeDV GUI (no locale for Country Germany)
If Linux (locale set to de_DE@UTF8) the hamlib sends now the frequency with 7,xxxx instead of 7.xxxx

So I removed the extra wxWidget settings for locales and use POSIX instead. That fix both problems. If you not translate the app then locales make no sense (except the date/time display in the reporter windows). But anywhere, better we use POSIX globally.

73 Heiko, DL1BZ


tmiw:
@barjac / @Tyrbiter: how do you guys feel about localized date/time formatting?

> If macOS (German version here) I get an error windows if I start FreeDV GUI (no locale for Country Germany)

I'm not sure about this. Might need to test some more locally (I have a few Macs I can try).

> If Linux (locale set to de_DE@UTF8) the hamlib sends now the frequency with 7,xxxx instead of 7.xxxx

BTW Hamlib in theory is supposed to temporarily change to "C" locale before switching back to whatever the user's using (https://github.com/Hamlib/Hamlib/blob/master/include/num_stdio.h). Perhaps it's a radio-specific issue? I'm not sure if all radios Hamlib supports use the code in the mentioned file or if some do their own thing.

> add 20db more for TX Mod Level (and rename TX Attenuation to Mod Level TX), because there are devices where 0db is not enough AF level for input -> add 20db more now

So the range for this slider is now -30 to +20, correct? My concern would be around people accidentally overdriving ALC by amplifying the TX signal when it shouldn't be. WSJT-X, for example, still has 0 dB as the maximum IIRC.

dl1bz:
> @barjac / @Tyrbiter: how do you guys feel about localized date/time formatting?
> 
> > If macOS (German version here) I get an error windows if I start FreeDV GUI (no locale for Country Germany)
> 
> I'm not sure about this. Might need to test some more locally (I have a few Macs I can try).

Yes, you use sure an English environment. But I use German, and there I get an error message from wxWidget-API if I start FreeDV "No locale for Country Germany".

> 
> > If Linux (locale set to de_DE@UTF8) the hamlib sends now the frequency with 7,xxxx instead of 7.xxxx
> 
> BTW Hamlib in theory is supposed to temporarily change to "C" locale before switching back to whatever the user's using (https://github.com/Hamlib/Hamlib/blob/master/include/num_stdio.h). Perhaps it's a radio-specific issue? I'm not sure if all radios Hamlib supports use the code in the mentioned file or if some do their own thing.

My SDR app has an rigctld running, I connect FreeDV with hamlib net at 127.0.0.1 as hamlib client. In Linux my Locales are de_DE@UTF-8. In German the decimal marker is , and not . like in English environments. Looks like the locale settings change this - and it is correct for German. But in this case, if I select the frequency in FreeDV, I see a freq set request with e.g. 7,xxx instead of 7.xxxx. Result: Wrong freq will be set. After my patch Englisch (or better "C" = POSIX) is set global  - then all works perfect, no error msg with macOS and Linux sends the correct freq now.

> 
> > add 20db more for TX Mod Level (and rename TX Attenuation to Mod Level TX), because there are devices where 0db is not enough AF level for input -> add 20db more now
> 
> So the range for this slider is now -30 to +20, correct? My concern would be around people accidentally overdriving ALC by amplifying the TX signal when it shouldn't be. WSJT-X, for example, still has 0 dB as the maximum IIRC.

I study the code, I think there is no problem to increase the level up to +20db. There are some target devices, where max. 0db generate not enough AF level for set enough output for the TX. 0db is the best choice, no question, but a kind of reserve is also sometimes helpful.

73 Heiko, DL1BZ

dl1bz:
One note: The comparision with FT8 has limits. FreeDV is NOT "radio in the noise" like FT8. We need much more SNR for use it. Similar is the setup in the TRX. We need normally a flat audio input and output for transport the TX signal with high linearity. That represent the Digimodes DIGL or DIGU, maybe USB-DATA or LSB-DATA in modern TRX. Older TRX havn't this and maybe they have pre- and deemphasis fixed wired in the input line, which is not good for transport the AF signal with high linearity. We generate a TX signal characteristic over a bandwidth with 1.5kHz if using RADEV1, that needs to transmit with the possible highest linearity. For decoding we need a specific SNR, which is never so low as in FT8. Means, we need to transmit with much more TX power as in FT8. The marker, if no TX spektrum is visible like with the SDR TRX, is the ALC. It shows, how much is the TX level out. Modern MOSFET PA have not a problem with a ALC between 50-75%. So we need a litttle headroom for the transmit AF audio, thats why I increase the TX Mod level up to +20db. The best setting is - no question - a range between -3db and 0db for the TX AF signal.

The other things is - the definition of 0db in audio. There are 2(!) specifications. 0db in professional audio like broadcasting means 1Vss. In home audio this value is only 0.775 Vss. But both are marked with 0db. That means, if you talk from 0db you need to say in addition, whats exact the reference value or better voltage.

The newest RADE is a Quantum Leap in DV at shortwave. Thats the only mode, I'm sure, that will be accepted from the masses. Means, the other mode can be kicked out in the near future - the audio quality is simply not acceptable.
Only RADE mode has the power for an everyone acception. That's my clear meaning. 

73 Heiko, DL1BZ

tmiw:
I was able to duplicate the macOS error on my English-only laptop using the instructions outlined in https://github.com/drowe67/freedv-gui/pull/1010. I wasn't able to duplicate the other Linux-related issue, though, in that frequency changes still seemed to work even with a comma instead of a period in the frequency. (This may be a result of the macOS fix, though.) Let's back out the locale changes from here, get it solid in the other PR and then come back to this PR (mainly so we can stay focused on just the slider scaling changes here).

> One note: The comparision with FT8 has limits. FreeDV is NOT "radio in the noise" like FT8. We need much more SNR for use it. Similar is the setup in the TRX. We need normally a flat audio input and output for transport the TX signal with high linearity. That represent the Digimodes DIGL or DIGU, maybe USB-DATA or LSB-DATA in modern TRX. Older TRX havn't this and maybe they have pre- and deemphasis fixed wired in the input line, which is not good for transport the AF signal with high linearity. We generate a TX signal characteristic over a bandwidth with 1.5kHz if using RADEV1, that needs to transmit with the possible highest linearity. For decoding we need a specific SNR, which is never so low as in FT8. Means, we need to transmit with much more TX power as in FT8. The marker, if no TX spektrum is visible like with the SDR TRX, is the ALC. It shows, how much is the TX level out. Modern MOSFET PA have not a problem with a ALC between 50-75%. So we need a litttle headroom for the transmit AF audio, thats why I increase the TX Mod level up to +20db. The best setting is - no question - a range between -3db and 0db for the TX AF signal.

Yep, you definitely need more SNR on the other end than, say, FT8. However, FreeDV (no matter which mode) behaves similarly to other digital modes in that it degrades the more ALC is needed, so the recommendation has always been to back off the levels a bit if ALC starts indicating on your radio (usually 1 bar flickering *at most*, though this depends on the radio).

Re: RADE--FreeDV uses int16 internally, meaning the audio samples sent to the configured audio device peak around +/- 16384. Thus you do *not* want to go much above +1-2 dB of amplification on the FreeDV side to avoid clipping. Recommend backing out the TX Attenuation changes, or if you feel that users still need to apply gain > 0 dB, to limit the upper end of the slider appropriately.

dl1bz:
> Re: RADE--FreeDV uses int16 internally, meaning the audio samples sent to the configured audio device peak around +/- 16384. Thus you do _not_ want to go much above +1-2 dB of amplification on the FreeDV side to avoid clipping. Recommend backing out the TX Attenuation changes, or if you feel that users still need to apply gain > 0 dB, to limit the upper end of the slider appropriately.

int16_t is +/- 32767 and not +/- 16384. I made some checks today with the result, +6db was enough. I add a soft clipper in LevelAdjustStep to not overdrive all with a wrong ScaleFactor. You should always do that if you proceed samples.

Suggestion: We stop now at +10db ?

Heiko, DL1BZ
Maintainer Project deskHPSDR

Tyrbiter:
May I just add that it is very important to compare the SNRs needed in the actual bandwidth used for demodulation, the numbers for FT8 are scaled from a 2.4kHz or 3kHz overall bandwidth. When considered in the actual 50Hz bandwith of the FT8 signal the SNR is actually between 16 and 18dB higher than the figures WSJT-X shows, RADE is capable of producing intelligible speech at negative SNR that is not much higher than the true SNR for FT8.

Also note that the ALC issue on transmit is very dependent on the PAPR of the RADE signal, currently in RADEV1 this is higher than desired but work is underway to reduce this for RADEV2 which will help to remove the effect of Tx chain compression and ALC control.

dl1bz:
> Also note that the ALC issue on transmit is very dependent on the PAPR of the RADE signal, currently in RADEV1 this is higher than desired but work is underway to reduce this for RADEV2 which will help to remove the effect of Tx chain compression and ALC control.

As I understand the theory behind RADEV1 correct, except the other mode like 700 or 1600 we have a PAPR < 1db in the case of RADEV1. This is very low and allows the use of almost the full average power without headroom reserves, significantly reducing the risk of clipping and ALC intervention in the power amplifier transmitter. This differs drastically from classical modes such as OFDM or SSB, where PAPR is usually in the range 6–12 dB.

73 Heiko, DL1BZ

Tyrbiter:
The issue is that the PAPR is being affected by filtering within the transmit chain, so that 1dB is actually significantly more.

A better training process is being worked on for RADEV2 which allows a better waveform to be created. It's described in David Rowe's latest blog articles at https://freedv.org

dl1bz:
Ok I read the latest news, the problem ist not the "base signal", it's the bandpass filter after, which produce results at the end with a increased PAPR to around 4.5db. Let's see if it can be make better with the development of RADEV2. 

But anywhere - back to my part for this PR here - there are devices, where I can't get a usable TX level if I stop at 0db. I made some field tests, my max. raise was +6db. That produced no problems, but enough TX level for the TRX to get an acceptable output power, which I can't get with only 0db. Decoding was fine and clear without any artefacts at a SNR around -1..+3db. So for now I see no significant reasons for not increasing the TX level  - ever inside a moderate range. So these are results from practice - not from pure theory. The TX spectrum was clean, looks ok and was not overdriven.

If this part of my PR will be included in the master branch is only a decision by the maintainers. Maybe yes, maybe not. In my own patched version I use this and it stays in. I decided so.

73 Heiko, DL1BZ

Tyrbiter:
It's really appreciated that you are working on this Heiko, assistance from someone with technical knowledge and coding ability helps FreeDV a lot, if your changes work for everyone I'm sure they will end up being merged. The developers don't have every variation of hardware to test with, so please carry on doing what you're doing.

dl1bz:
I will do that with great pleasure. I had done the same before I start with my own project deskHPSDR. I start with a fork of pihpsdr, because there were a lot things I needed, but not availible. So step by step I extend deskHPSDR, today it is a independent application. If I start I had done the same, sent my ideas to the developer. But the most were dropped for reasons that only he knows. After that I start deskHPSDR and begun to implemented my own ideas. But here I try to share my ideas, because I think, FreeDV can be a great, wide accepted App for Digital Voice in shortwave. RADEV1 is a quantum leap in this direction. V2 maybe much better. Many heads make such a app better, step by step. That's I also learned during my development on deskHPSDR.

There are two layer - the GUI/frontend, which uses the user the whole time. And the backend, which have all the knowledge included for what the app is made. Important is - the GUI must be easy to use, easy to understand and clearly structured.
I still see some potential there and this is what I try to improve now or at least to help to improve that.

73 Heiko, DL1BZ

src/gui/dialogs/dlg_filter.cpp
```diff
@@ -849,6 +854,8 @@ void FilterDlg::sliderToGain(EQ *eq, bool micIn)
     float range = MAX_GAIN-MIN_GAIN;
 
     eq->gaindB = MIN_GAIN + range*((float)eq->sliderGain->GetValue()/SLIDER_MAX_GAIN);
+    // auf ganze dB rasten
+    eq->gaindB = std::round(eq->gaindB);
```
tmiw, line 858: Should this be `lround` as well just like in `setGain()`?

src/pipeline/LevelAdjustStep.cpp
```diff
@@ -57,7 +58,9 @@ std::shared_ptr<short> LevelAdjustStep::execute(std::shared_ptr<short> inputSamp
 
     for (int index = 0; index < numInputSamples; index++)
     {
-        outputSamples_.get()[index] = inputSamples.get()[index] * scaleFactor;
+        double y = inputSamples.get()[index] * scaleFactor;
+        y = 32767.0 * std::tanh(y / 32767.0);  // Soft-Clip statt hartem Clipping
```
tmiw, line 62: I always get concerned when we start to add non-linearity, especially if is intended to prevent clipping a digital waveform. Can we disable this for the TX level slider but leave it for the mic/speaker volume one?

tmiw (commented):
See below. I'll give this PR a try on my end soon and see how it goes.

tmiw:
> > Re: RADE--FreeDV uses int16 internally, meaning the audio samples sent to the configured audio device peak around +/- 16384. Thus you do _not_ want to go much above +1-2 dB of amplification on the FreeDV side to avoid clipping. Recommend backing out the TX Attenuation changes, or if you feel that users still need to apply gain > 0 dB, to limit the upper end of the slider appropriately.
> 
> int16_t is +/- 32767 and not +/- 16384. I made some checks today with the result, +6db was enough. I add a soft clipper in LevelAdjustStep to not overdrive all with a wrong ScaleFactor. You should always do that if you proceed samples.

I should clarify--RADE (and the FARGAN voice codec that it uses) deals with floating-point values from -1.0 to 1.0 and to convert them to int16 for feeding to the user's audio device, FreeDV multiplies by 16384. As this results in values for audio samples that can range between `0.5 * INT_MIN` and `0.5 * INT_MAX` (-16,385 - 16,384), that means we can only multiply by 2 (aka 3 dB) before we max out the range of int16 and start inducing clipping.

> Suggestion: We stop now at +10db ?

As mentioned above, 3 dB is probably the safest maximum we can go with for the TX Level slider as digital signals *really* don't like non-linearity. 10 might be okay for the mic/speaker level (with soft clipping).

tmiw:
> But anywhere - back to my part for this PR here - there are devices, where I can't get a usable TX level if I stop at 0db. I made some field tests, my max. raise was +6db. That produced no problems, but enough TX level for the TRX to get an acceptable output power, which I can't get with only 0db. Decoding was fine and clear without any artefacts at a SNR around -1..+3db. So for now I see no significant reasons for not increasing the TX level - ever inside a moderate range. So these are results from practice - not from pure theory. The TX spectrum was clean, looks ok and was not overdriven.

Out of curiosity, what devices are exhibiting this behavior?

tmiw:
I did some local testing here where I pushed the Record button, started transmitting and slowly increased the TX level slider from 0 dB to 10 dB.  I then played back this file and noticed a few things:

1. SNR stayed pretty solidly at 34 dB even at 10 dB gain, so the soft clipping does seem to be working.
2. Analyzing the last few seconds of the transmission in Audacity seemed to still indicate a linear waveform:

image

I also tried transmitting this signal over the air. The TX waterfall on my Flex 6300 looked okay (even at 10 dB) but that's not to say there won't be issues with other radios (and as mentioned above, I don't really have the resources to test on every radio out there). That's why I'd still recommend capping the TX level at 3 dB.

g4dya:
In integer digital audio, there is a natural built-in 'reference level' - the clipping point. This is 0 dBFS (Full Scale).

With the TX attenuation slider at maximum, the output level from freedv-gui (master branch) reaches -4 dBFS true peak as measured on jkmeter on linux. In other words it's only 4 dB below clipping. It's not a good idea to go above this because the audio chips typically used in USB audio interfaces, whether external or inside a rig, start to get non-linear when you get close to 0 dBFS.

If there's a loss of gain later in the TX chain, then it needs to be fixed later in the TX chain. 

I concur that SNR maxes out at 34 dB. That's what I've seen when playing back my own transmission from a file.

dl1bz:
> In integer digital audio, there is a natural built-in 'reference level' - the clipping point. This is 0 dBFS (Full Scale).
> 
> With the TX attenuation slider at maximum, the output level from freedv-gui (master branch) reaches -4 dBFS true peak as measured on jkmeter on linux. In other words it's only 4 dB below clipping. It's not a good idea to go above this because the audio chips typically used in USB audio interfaces, whether external or inside a rig, start to get non-linear when you get close to 0 dBFS.
> 
> If there's a loss of gain later in the TX chain, then it needs to be fixed later in the TX chain.
> 
Thats all correct, no question. I know this all too, 0db is the reference. But "fixed later in the TX chain" is pure theory and far from reality and can't be done in every case. You can fix e.g. firmware in commercial TRX by yourself ? I can't. Look in professional audio tools like Voicemeeter, there is ever an additional gain option above 0db. How much above 0db is another question. I know very well, out TX AF signal is a ready-to-run digital modulated signal for transmit - not pure AF.
The opinions, _where_ you need to correct, at the source side or at the target side are different. A the end you need to avoid signal clipping.

But anywhere, and that confirm all of my tests, +3db is possible. I agree with @tmiw, more can be critical. So my last change is a range between -30db up to +3db. I repeat my tests with the problematic device and say, I can live with +3db more and see not great problems in practise with it.

73 Heiko, DL1BZ

src/gui/dialogs/dlg_filter.cpp
dl1bz: eq->gaindB = std::round(eq->gaindB);   // stay double
eq->gaindB = std::lround(eq->gaindB);  // will be long

In this case we want to keep GaindB internally as a floating point value, but lock it to integer dB.
So round instead of lround is correct in this case.

g4dya:
Sample-rate conversion of the RADE signal after it leaves freedv-gui may increase the true-peak level by a few dB. It's worth keeping a little headroom to allow for this. If we don't then users may inadvertently encounter clipping. For this reason -4 dBFS is a safe maximum level. IMHO it should not be increased.

dl1bz:
I am making a PR here and therefore _only a suggestion or recommendation_. I personally work with my own fork of FreeDV, where I do what I think is right or necessary. Therefore, the decision whether to adopt it or not into the offical version is up to the maintainers and is not up to me. There are reasons _for_ and _against_ this change.
On the other hand – nobody HAS to raise. I add a tooltip if you use this slider and explain the correct procedure.
The reference point is 0db or less, I recommend -3db. Thats my value is I use my deskHPSDR in combination with my Hermes Lite 2 SDR-TRX. For digimode like DIGL or DIGU I set internally in deskHPSDR all strict to 0db and all audio tools are deactivated for transmit a pure, unprocessed AF input signal like RADEV1 or FT8 or whatever. The ALC shows, thats all correct. But my problem was not my Hermes Lite 2 and deskHPSDR, I have the level issue with another, older TRX.

All of my practical tests have shown that a _slight increase_ does not cause any serious problems and solved my problem.
Theory is always one thing, practice another. And sometimes practice contradicts theory :-)
If the problems are fixed with the bandpass and the resulting higher PAPR at the moment, means we get really a PAPR <1db, then at the latest we will have more flexibility.

73 Heiko, DL1BZ

dl1bz:
Add a simple limiter in the audio output chain. Make sense if the QSO partner not using the AGC.
Chain is Demod → EQ/Filter → per-Station-Gain (Software-Volume) → Limiter → Output-FIFO → OS/Device
Too loud stations will be limited before Output-FIFO at −1 dBFS.
Look-ahead 5 ms, Attack 1 ms, Release 50 ms, Threshold −1 dBFS.

73 Heiko, DL1BZ

barjac:
Tried to test but this PR patch does not apply to current master, several hunks fail.

tmiw:
@dl1bz, can you confirm that you synced up with the latest master? It looks like you did do a merge at some point but perhaps not too recently.

dl1bz:
> @dl1bz, can you confirm that you synced up with the latest master? It looks like you did do a merge at some point but perhaps not too recently.

At the moment I'm in holidays  for two weeks. I can do it first if I'm back.

73 Heiko, DL1BZ

dl1bz:
Close my PR now and pull it back and I will forego further PR. You need too much time for decision if add the PR or not. This waste my time to make permantly forward and backward patches against the changed master. Further work I will do now only in my own fork.

73 Heiko, DL1BZ

Tyrbiter:
We appreciate your effort Heiko, and perhaps when everything has settled down you can open another PR for inclusion into the official version. We are very resource limited and need approval from the project leadership team for significant changes, being open source there is no downside to you building your own version.

dl1bz:
My time is also very limited. I cannot invest the whole time for only backporting my PR because a lot changes in the master branch. I understand very well, this is an active project and a lot of code changed in short time, but three weeks waiting for a decision is not acceptable for me and the reason, why I will not be submitting any more PRs until further notice. Stop and go is not my way of working. In the end, you have to decide whether you want PRs at all or whether you would rather do without them. If so, you need to find ways to process them more quickly, I would say within a week at the most.
And one notice from me after one year development of deskHPSDR, I learn this too and very quickly: Make _one_ commit for _one_ problem, not one commit for 10 problems. Squash the commits in logical order for a specific problem solution. This keeps the changes clear and problem-specific for developers. That would be a great help in the case of backporting PRs if needed. You know, what you had changed as Maintainer - other need get a new overview about all changes, which cost a lot of time, especially if 20-30 files were changed.

PR #1010, opened 2025-08-20
Fix various locale related bugs.

tmiw:
Per #1009, this PR fixes various bugs related to locales:

* macOS: Fix spurious error on startup when changing locales (tested with `src/FreeDV.app/Contents/MacOS/FreeDV -AppleLanguages “en-US” -AppleLocale “de-DE”` per https://forum.xojo.com/t/start-app-in-a-different-language-with-different-locale-settings/25219)
* Linux: TBD

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -32,6 +32,7 @@
 #include <wx/cmdline.h>
 #include <wx/stdpaths.h>
 #include <wx/uiaction.h>
+#include <wx/uilocale.h>
 
 #include "git_version.h"
 #include "main.h"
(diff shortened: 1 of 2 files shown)


tmiw:
@dl1bz, are you able to try this PR branch and let me know how it goes re: the issues you discovered? On Linux with this PR, I didn't have to do anything additional to make the frequency selection behave properly even with the comma. Tested by doing the following:

1. Start FreeDV from the `build_linux` folder with `LANG=de_DE.UTF-8 PYTHONPATH=$(pwd)/rade_src src/freedv`
2. In another terminal window (from the `freedv-gui` folder): `python3 test/hamlibserver.py 999999999` (Optional; you should be able to use your existing config.)
3. If doing (2), use "Hamlib NET rigctl" for the rig type and "localhost:4575" as the "serial port" in Tools->CAT and PTT Config.
4. Press Start inside FreeDV and select various frequencies from the drop-down list to confirm the radio is actually changing. If doing (2) there will be debug output in the second terminal window indicating what commands it's receiving and what responses it's sending back.

PR #1011, opened 2025-08-22
Use Speex resampler in the pipeline instead of libsamplerate.

tmiw:
This PR ports the changeover from libsamplerate to Speex resampler from #975. In testing on Linux using `perf`, this reduced the percentage execution time of `ResampleStep` from ~11% to ~1%.


barjac:
Used this today and did not notice any regressions. The audio level using AGC seems rather low so I ended up with around +3dB in the TX GUI to achieve an adequate mic level of around +/-0.3 in the from mic display.

tmiw:
Yeah, I didn't hear anything out of the ordinary when operating OTA the other day but this might be causing regressions in the automated tests. Might revisit once #975 merges.

tmiw:
Unfortunately there are still intermittent failures no matter what quality level we use, so I'm just going to close this out for now.

tmiw:
Reopening so I can see how much impact on CPU usage this will have with the Flex waveform.

tmiw:
Per testing, it seems that this only saves around ~1% CPU on the Pi 4 when running the Flex waveform. Closing again.

PR #1016, opened 2025-08-25
Prevent lockup/crash when testing Hamlib in PTT Config window.

tmiw:
Per reports from off GitHub, this PR fixes an issue where FreeDV crashes if a user attempts to test PTT on the PTT Config window with an invalid configuration.


tmiw:
Note: I still need to test the PTT config on the Easy Setup window. I think that one works a bit better but need to make sure.

EDIT: Easy Setup window seems to work properly, no changes needed there.

peterbmarks:
Thank you!

PR #1020, opened 2025-09-02
Use rigerror2() for friendlier Hamlib errors.

tmiw:
Resolves #1015 by using `rigerror2()` instead of `rigerror()` for any Hamlib error messages that users see.

Note: `rigerror2()` is undocumented in the Hamlib API but has been suggested by developers on the Hamlib mailing list.

PR #1021, opened 2025-09-03
Linux: fix rendering bug for mic/speaker slider when transitioning from TX to RX.

### LINKED ISSUE #1019: Mic/Spkr level slider position loses sync with numeric value in the same widget.
I regularly see the slider off centre when the text says '0.0 dB' (which should really say '0.0dB' without a space).
I suspect that this happens when I end a transmission where I have used a slider offset and return to RX where I don't.
The numeric value is updated but the slider position is not until some time later.

Image

tmiw:
Resolves #1019 by forcing a redraw of the mic/speaker level slider when transitioning from TX to RX. This PR also removes the space between the numeric value and "dB".

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -742,12 +742,12 @@ void MainFrame::loadConfiguration_()
 
     char fmt[15];
     m_sliderTxLevel->SetValue(g_txLevel);
-    snprintf(fmt, 15, "%0.1f dB", (double)g_txLevel / 10.0);
+    snprintf(fmt, 15, "%0.1f%s", (double)g_txLevel / 10.0, "dB");
     wxString fmtString(fmt);
     m_txtTxLevelNum->SetLabel(fmtString);
     
     m_sliderMicSpkrLevel->SetValue(wxGetApp().appConfiguration.filterConfiguration.spkOutChannel.volInDB * 10);
-    snprintf(fmt, 15, "%0.1f dB", (double)wxGetApp().appConfiguration.filterConfiguration.spkOutChannel.volInDB);
+    snprintf(fmt, 15, "%0.1f%s", (double)wxGetApp().appConfiguration.filterConfiguration.spkOutChannel.volInDB, "dB");
     fmtString = fmt;
     m_txtMicSpkrLevelNum->SetLabel(fmtString);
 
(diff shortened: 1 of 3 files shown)


barjac:
Nope, still stays at TX slider position until the mouse travels over widget, then it corrects. I see the space has gone though :)

Tyrbiter:
A bit of testing with https://github.com/drowe67/freedv-gui/pull/1021/commits/ff4486c1d848d8bc07f2081899d8f008035d048c here using this sequence:

Press PTT to start Tx
Change Mic level slider to +3.0dB
Press PTT to return to Rx
Mix/spkr level slider returns to 0dB after the EOO time and immediately after PTT on the radio drops.

I think that's an "it works for me".

Also checked that this sticks after restarting freedv-gui, which it does.

FYI I have also applied PR #1020 to this build, which I think cannot possibly affect the mic/spkr slider.

barjac:
Here is a short video to demonstrate. The message below FreeDV in the video was not sent as I decided to use a video instead, although most of what it says is relevant.

https://github.com/user-attachments/assets/57ce2d1a-9c24-4699-9943-dce3e66ff04f

tmiw:
I tried this on my Fedora VM (running Plasma 6.4.4 directly on Wayland and with `GDK_BACKEND=x11`) and didn't see the issue @barjac is seeing. Maybe it's a Mageia package thing?

barjac:
Was that stable Fedora or Rawhide?

Tyrbiter:
> Was that stable Fedora or Rawhide?

In my case it's stable Fedora 42, all updates applied (including some in pending awaiting being pushed to testing) and I am using Wayland.

Could this be more wxWidgets trouble in Mageia?

tmiw:
> Was that stable Fedora or Rawhide?

Fedora 42 for me too.

barjac:
> > Was that stable Fedora or Rawhide?
> 
> In my case it's stable Fedora 42, all updates applied (including some in pending awaiting being pushed to testing) and I am using Wayland.
> 
> Could this be more wxWidgets trouble in Mageia?

I don't think so.
I have just tested in stable Mageia 9 and it is the same.
I then re-built FreeDV at the commit that first introduced the Mic/Spkr slider (#dd2932b) also in Mageia 9 and it is the same. I guess I just never noticed it.
Mageia 9 for testing this is on my laptop with X11 and lib64wx_gtk3u_* are all 3.2

Also tested in OpenSUSE Tumbleweed and it is just the same.

I have just confirmed your findings in Fedora 42 (in a VM).

Very odd and it's not only Mageia with the issue.

tmiw:
How about with [7e0c512](https://github.com/drowe67/freedv-gui/pull/1021/commits/7e0c512af3bf45712dfbf880ecb15ba7ffa3e99e)?

Tyrbiter:
> How about with [7e0c512](https://github.com/drowe67/freedv-gui/pull/1021/commits/7e0c512af3bf45712dfbf880ecb15ba7ffa3e99e)?

I built https://github.com/drowe67/freedv-gui/pull/1021/commits/d73e8330621dbc1167c010885d9b08a20cb7b83e and it is still working as before here.

barjac:
\o/ Yay! Looks good.
7e0c512 is tested in Mageia 9 and it's OK,
d73e833 is tested in Mageia 10 and it's OK,
d73e833 is tested in Mageia 9 and it's OK,
d73e833 is tested in OpenSUSE Tumbleweed and it's OK.

Push it quick! :)

PR #1023, opened 2025-09-08
Debug and fixing of unit tests

tmiw:
This PR contains the following fixes:

1. macOS: Adds retry logic due to known issues with hdiutil during DMG generation.
2. Adds `std::atomic` for playback-related variables as a possible fix for tests hanging until GitHub timeout of 6 hours.
3. EOO is now enqueued as a single unit to hopefully improve reliability of `rade_reporting_*` tests.


Tyrbiter:
Well, this looks good.

tmiw:
> Well, this looks good.

Maybe. The original issue is intermittent, though, so we might need to do a few more builds.

Tyrbiter:
> > Well, this looks good.
> 
> Maybe. The original issue is intermittent, though, so we might need to do a few more builds.

Fine with me, it's just that I haven't seen all the tests just work for some time :)

tmiw:
Of course, now I can't duplicate the original issue (FreeDV hanging during tests until GH times out). Or well, not when I'm actively looking at this anyway.

Issue #1022, opened 2025-09-07
Should the current AGC implementation be removed from master branch before any new release?


barjac:
In principle the idea of AGC is good, in order to avoid large level variations between stations, however in practice it is our opinion that the current implementation is not adequately achieving this goal.

In some cases it appears to have the reverse effect in that it becomes more difficult to set an acceptable audio level with it enabled than without it.

We, (the stations listed below) have discussed this at length and feel that the current implementation should be removed from master before any new general release is made.

We welcome further development in this area but want to avoid any 'bad press' about FreeDV that could arise from keeping this feature in its present form.

Having all tested it over many QSOs, we have now turned it off in favour of using manual settings.

73
G4MKT, G4DYA, G6WPJ, GW7SAB, 2D0DRM

Tyrbiter:
An interesting suggestion @barjac 

Can you describe what happens that makes things more difficult with AGC than without, it's worth knowing the details if possible as I'm not clear exactly why you have all come to this conclusion.

tmiw:
Alternatively, the default for AGC can just be changed to disabled instead of enabled. That way, the behavior doesn't change for people unless they opt in. Though as @Tyrbiter said it would be a good idea to know how the current AGC implementation makes things harder in some cases.

g4dya:
Without AGC, the only thing to adjust is the overall gain. This can be done either in the audio chain before freedv-gui or (in 2.0.2-dev) in freedv-gui itself. With AGC, in order to solve the 'first word' problem, it's necessary to adjust a combination of both by trial and error. Having done this the level does not stay consistent from one transmission to the next.

Looking at libspeexdsp, it seems the documentation for preprocess.c isn't up to date with the code.

Also, using Grok AI, I looked at the amplitude/frequency weighting curve used in the preprocess.c AGC and there's a huge peak in the response at 3.8 kHz which makes no sense at all. (As a sanity check I asked it to also calculate the ITU-R BS.1770 curve which it has done correctly.) 

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/ce6571a2-93c1-4c41-bf6b-358445a2548c" />

Tyrbiter:
Well that is odd, there is a 9dB flat level reduction up to about 1.5kHz, and as you say the peak makes no sense.

Also:

$ cat /usr/share/doc/speexdsp/TODO 
For 1.2:
Major points:
[...]

PR #1024, opened 2025-09-09
Use EBU R 128 based loudness measurement to implement AGC.

tmiw:
Per feedback in #1022, this PR goes away from libspeexdsp AGC and implements an AGC based on [EBU R 128 loudness measurement](https://en.wikipedia.org/wiki/EBU_R_128).

TODO:

* ~~Update Filter dialog to allow AGC to be enabled/disabled without Speex being enabled first.~~

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -868,7 +868,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Various GUI and FreeDV Reporter performance improvements. (PR #1002, #1026)
     * Use rigerror2() for friendlier Hamlib errors. (PR #1020)
 3. Build system:
-    * Update Hamlib to v4.6.4. (PR #989)
+    * Update Hamlib to v4.6.5. (PR #1030)
     * macOS: Upgrade PyTorch to 2.7 (and NumPy to 2.3). (PR #1003)
 4. Documentation:
     * Add pkgconf dependency for macOS instructions. (PR #1013)
--- a/cmake/BuildHamlib.cmake
+++ b/cmake/BuildHamlib.cmake
@@ -16,7 +16,7 @@ endif(MINGW AND CMAKE_CROSSCOMPILING)
 
 include(ExternalProject)
 ExternalProject_Add(build_hamlib
-    URL https://github.com/Hamlib/Hamlib/archive/refs/tags/4.6.4.zip
+    URL https://github.com/Hamlib/Hamlib/archive/refs/tags/4.6.5.zip
     BUILD_IN_SOURCE 1
     INSTALL_DIR external/dist
     PATCH_COMMAND ${HAMLIB_PATCH_CMD}


g4dya:
I tried de2af3f and I'm not sure whether AGC is working or not. I get this repeated on STDERR when I press PTT:

19:13:50 [50301] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)
19:13:50 [50302] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)
19:13:50 [50303] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)
19:13:50 [50304] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)
19:13:50 [50305] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)

barjac:
It is working fine here.
No problems at all so far in testing. Much improved.

Interesting that you have used webrtcvad as I have been using the same over the last few days with some success after throwing G4DYA's spec at chatGPT which went as follows:-

Me (thanks to G4DYA :):
"I am looking for an automatic gain control to place ahead of freedv.
The ideal kind of AGC would be a slow-acting loudness normalization, gated so that the gain only changes when the user is speaking. During pauses in speech, or when in receive, the mic gain would not change. Loudness would be measured by the method in Rec. ITU-R BS.1770. Having measured the loudness of several stations a target loudness of -20 LUFS would probably be about right. Can you suggest a way of achieving this with either GnuRadio or other audio building blocks."

//-----------------
ChatGPT:
Nice, good spec — that’s totally doable. Below I give a practical plan, design notes, and a ready-to-run Python sketch you can drop into a GNU Radio or plain-JACK / PulseAudio chain. The idea:

measure loudness via BS.1770 (use an existing library: pyloudnorm or libebur128)

detect speech activity via a lightweight VAD (e.g. webrtcvad) and only update gain while VAD==true (your “gated” requirement)

compute required gain to reach target = −20 LUFS and apply a slow attack / slower release smoothing (exponential) so it’s “loudness normalization” rather than fast AGC

always run a ceiling/limiter after gain so peaks don’t clip the codec/ADC

Below: block diagram, component choices, parameter suggestions, and a Python prototype you can adapt into GNU Radio as a custom block (or run as a JACK client feeding FreeDV).

Block diagram

Audio input (mic)
→ Tap A: VAD (webrtcvad) → Speech activity flag
→ Tap B: BS.1770 loudness meter (libebur128 or pyloudnorm) → Measured LUFS
→ Gain controller (only updates when Speech activity == true) → Smooth (attack/release) → Limiter → Encoder / FreeDV input

Component recommendations

Loudness measurement (BS.1770): use pyloudnorm (python) or libebur128 (C). These implement the BS.1770 / EBU R128 algorithms (use them rather than hand-rolling BS.1770 unless you want deep DSP).

Voice activity detection (gate): webrtcvad (Python package) — very small, low-latency and robust for speech vs silence.

Audio I/O:

If you want GNU Radio native: implement the controller in a custom Python block (audio stream floats) and place it before the FreeDV block.

If you prefer standalone: use Jack (recommended for low-latency routing) and a small Python JACK client (jack-client or sounddevice with JACK backend) that multiplies the audio stream by the computed gain and outputs to the FreeDV input.

Limiter: a simple brickwall or soft clipper (e.g., one-pole peak limiter or a small lookahead compressor) to prevent clipping. FreeDV’s encoder will be unhappy with heavy clipping.

Smoothing (time constants): pick slow attack/release so the normalization is natural:

attack_time ≈ 0.5–1.0 s (how quickly gain rises when loudness falls below target)

release_time ≈ 6–12 s (how slowly gain reduces when loudness goes above target)

hold_time after speech→silence: 0.5–2 s (do not change gain during brief pauses)

Gain limits: min_gain = −20 dB, max_gain = +12 dB (protect mic/ADC and avoid excessive noise boost)

Behavior details to match your spec

Gate: only update the controller when VAD reports the user is speaking. During receive or speech pauses, the gain is held (no changes).

Loudness target: −20 LUFS (your chosen target).

Loudness measurement window: BS.1770 specifies integrated/short-term/momentary. For realtime control use a short-term or sliding integrated estimate (e.g., 3 s sliding window) but rely on pyloudnorm’s momentary/short-term routine or libebur128 to compute a continuously-updating loudness value.

Latency: choose buffer sizes small enough for FreeDV but remember smoothing introduces seconds of response by design.
//-----------------------------------------

That last comment proved to be an issue, however after several hours of testing and adapting and sampling over a much shorter period (0.5sec) the attached python script was evolved.

My thoughts after testing both this PR and the attached python lean towards this PR as being perfectly acceptable without the gating feature.

I am including the python since some aspects of this implementation may prove useful/inspiring to Mooneer :)

[freedv-agc.py](https://github.com/user-attachments/files/22241754/freedv-agc.py)

Tyrbiter:
> I tried [de2af3f](https://github.com/drowe67/freedv-gui/commit/de2af3fb0a9f08ef4daff80f277fd9ecf5a82445) and I'm not sure whether AGC is working or not. I get this repeated on STDERR when I press PTT:
> 
> 19:13:50 [50301] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1) 19:13:50 [50302] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1) 19:13:50 [50303] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1) 19:13:50 [50304] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1) 19:13:50 [50305] ERROR /home/richard/src/freedv-gui/src/pipeline/AgcStep.cpp:127: Failed processing AGC (err = -1)

Me too, this is with https://github.com/drowe67/freedv-gui/pull/1024/commits/de2af3fb0a9f08ef4daff80f277fd9ecf5a82445

tmiw:
Try 3ee9042? It seems to be working better with the GitHub tests, anyway.

Tyrbiter:
> Try [3ee9042](https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449)? It seems to be working better with the GitHub tests, anyway.

The Failed processing AGC messages on Tx have gone.

Along with the Failed checks...

tmiw:
> > Try [3ee9042](https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449)? It seems to be working better with the GitHub tests, anyway.
> 
> The Failed processing AGC messages on Tx have gone.
> 
> Along with the Failed checks...

Excellent. There's still the TODO above but additional feedback about how this AGC implementation behaves would be good.

Tyrbiter:
> > > Try [3ee9042](https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449)? It seems to be working better with the GitHub tests, anyway.
> > 
> > 
> > The Failed processing AGC messages on Tx have gone.
> > Along with the Failed checks...
> 
> Excellent. There's still the TODO above but additional feedback about how this AGC implementation behaves would be good.

@barjac commented on it above his huge chatGPT output entry.

tmiw:
> > > > Try [3ee9042](https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449)? It seems to be working better with the GitHub tests, anyway.
> > > 
> > > 
> > > The Failed processing AGC messages on Tx have gone.
> > > Along with the Failed checks...
> > 
> > 
> > Excellent. There's still the TODO above but additional feedback about how this AGC implementation behaves would be good.
> 
> @barjac commented on it above his huge chatGPT output entry.

Yep. Was thinking from @g4dya too. :)

barjac:
#3ee9 Working well, no sign of first word issues and loud explosives don't over drive or flat top.
Not yet worked anyone else using this (too many Win users). I will hopefully have a chat with G4DYA tomorrow.

g4dya:
3ee9042 has fixed the error messages for me too. No sign of first word problem but not much sign of AGC either. Too loud or too quiet stays as such. Is the AGC button on the filter menu supposed to do anything in this build?

tmiw:
> [3ee9042](https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449) has fixed the error messages for me too. No sign of first word problem but not much sign of AGC either. Too loud or too quiet stays as such. Is the AGC button on the filter menu supposed to do anything in this build?

Yep, the checkbox is supposed to still work. I did some testing again with https://github.com/drowe67/freedv-gui/commit/3ee9042f16c30dc5b439e0aef4c4ff2a9f84a449 and it looks like AGC is working, but I'm only seeing peaks of +/- 0.2. If you increase the microphone volume slider, does it make any sort of difference? If so, we might be able to add a 2-3 dB offset to that slider internally whenever AGC's enabled.

barjac:
No noticeable change with 2443df6
Levels out seem OK here with Mic/Spkr slider at 0.0dB and mic gain set so audio level meter in system shows that it does not clip on loudest likely input.
I think I see what Richard means, very low input levels do not seem to promote much increase in AGC gain to compensate, however I don't really see that as an issue since it's probably better that way than pulling up hiss and background noises in silent periods.

Tyrbiter:
> No noticeable change with [2443df6](https://github.com/drowe67/freedv-gui/commit/2443df6d7b03a5ce47920a43bcc0b287c5b0dca6) Levels out seem OK here with Mic/Spkr slider at 0.0dB and mic gain set so audio level meter in system shows that it does not clip on loudest likely input. I think I see what Richard means, very low input levels do not seem to promote much increase in AGC gain to compensate, however I don't really see that as an issue since it's probably better that way than pulling up hiss and background noises in silent periods.

It might be because the AGC decay time constant is fairly long, whereas the attack time constant needs to be fast to deal with the first word situation.

g4dya:
Testing 2443df6 with a voice keyer test file in which the same audio sample (me reciting the phonetic alphabet) is repeated at -30, -25, -20, -15 and -10 LUFS (the last two with limiting (1ms attack, 200ms release) to prevent clipping). The difference of 20 LU between the first and last input samples resulted in a 13.5 LU difference in the loudness of the decoded RADE files. (For this purpose LU can be taken as dB.) So there appears to be some AGC action but not very much.

-30 LUFS in   -25.0 LUFS out
-25 LUFS in   -19.4 LUFS out
-20 LUFS in   -14.9 LUFS out
-15 LUFS in   -12.3 LUFS out
-10 LUFS in   -11.5 LUFS out

tmiw:
How about now with [2cf887f](https://github.com/drowe67/freedv-gui/pull/1024/commits/2cf887f3795b2791c57e9bdba15c696836c07280)?

g4dya:
2cf8887f gave these results:

-30 LUFS in  -15.5 LUFS out
-25 LUFS in  -12.3 LUFS out
-20 LUFS in  -11.2 LUFS out
-15 LUFS in  -10.9 LUFS out
-10 LUFS in  -10.8 LUFS out

So a 20 LU change in input loudness was reduced to 4.7 LU which is much better. 

Will try the next commit in a few minutes.

@barjac - re. our discussion on air earlier today - voice keyer files do go through the AGC. The levels are the same whether played through the voice keyer or streamed in via the "mic socket".

g4dya:
299a1ed gave these results:

-30 LUFS in  -19.2 LUFS out
-25 LUFS in  -17.1 LUFS out
-20 LUFS in  -16.0 LUFS out
-15 LUFS in  -15.5 LUFS out
-10 LUFS in  -15.2 LUFS out

So a 20 LU change in input level was reduced to 4 LU, and a target output level of around -16 LUFS is OK. (Broadcasters use -24 LUFS (US) or -23 LUFS (EBU), streamers like Spotify and Youtube use -14 LUFS.)

g4dya:
After trying 299a1ed I had a comment that it sounded a bit loud. At -16 LUFS it might be, given that I've previously observed -20 LUFS to be typical without AGC. The waveform on 'From Mic' also looks quite flat-topped, so I guess there might be quite a bit of compression involved. Are there any parameters we can tweak, such as lengthening the recovery time (gain increase) ?

barjac:
> After trying [299a1ed](https://github.com/drowe67/freedv-gui/commit/299a1ed5d56c5552d08cfb26478e944d398fbd6d) I had a comment that it sounded a bit loud. At -16 LUFS it might be, given that I've previously observed -20 LUFS to be typical without AGC. The waveform on 'From Mic' also looks quite flat-topped, so I guess there might be quite a bit of compression involved. Are there any parameters we can tweak, such as lengthening the recovery time (gain increase) ?

I agree its a bit high now. Running with -2dB on the Mic slider looks about right to me.

I tried tweaking it a bit and this seems about optimum:

```
diff --git a/src/pipeline/AgcStep.cpp b/src/pipeline/AgcStep.cpp
index e2c00159..4c6ba4ff 100644
--- a/src/pipeline/AgcStep.cpp
+++ b/src/pipeline/AgcStep.cpp
@@ -56,7 +56,7 @@ AgcStep::AgcStep(int sampleRate)
     // Set AGC configuration
     agcConfig_.compressionGaindB = 21; // default 9 dB
     agcConfig_.limiterEnable = 1; // default kAgcTrue (on)
-    agcConfig_.targetLevelDbfs = 7; // default 3 (-3 dBOv)
+    agcConfig_.targetLevelDbfs = 6; // default 3 (-3 dBOv)
     status = WebRtcAgc_set_config(agcState_, agcConfig_);
     if (status != 0)
     {
```

tmiw:
Oops, I just pushed a change to bump `targetLevelDbfs` to effectively -9 dB. Does Barry's patch work better for you guys? If so, I can push the commit to apply it.

g4dya:
I haven't tried @barjac's patch, as I can't easily build it.

6282992 gave these results:

-30 LUFS in -21.2 LUFS out
-25 LUFS in -17.7 LUFS out
-20 LUFS in -16.3 LUFS out
-15 LUFS in -16.2 LUFS out
-10 LUFS in -15.8 LUFS out

This result is not as good as the previous one 299a1ed.

I think that the change is to the peak level in dBFS rather than loudness in LUFS, and it's reaching that by applying more compression/limiting, which is not what's needed.

Could we please go back to the setting in 299a1ed but put a 4dB 'pad' (i.e. x 0.63 amplitude) after the AGC when it's enabled? Then the loudness with -20 LUFS in would be -20 LUFS out, matching the typical loudness with AGC off and level set to  ±0.4.

tmiw:
> Could we please go back to the setting in [299a1ed](https://github.com/drowe67/freedv-gui/commit/299a1ed5d56c5552d08cfb26478e944d398fbd6d) but put a 4dB 'pad' (i.e. x 0.63 amplitude) after the AGC when it's enabled? Then the loudness with -20 LUFS in would be -20 LUFS out, matching the typical loudness with AGC off and level set to ±0.4?

Done.

barjac:
> > Could we please go back to the setting in [299a1ed](https://github.com/drowe67/freedv-gui/commit/299a1ed5d56c5552d08cfb26478e944d398fbd6d) but put a 4dB 'pad' (i.e. x 0.63 amplitude) after the AGC when it's enabled? Then the loudness with -20 LUFS in would be -20 LUFS out, matching the typical loudness with AGC off and level set to ±0.4?
> 
> Done.

Sorry but that is not good.
Its now hard clipping at +/-0.2 in the 'from mic' display!

Re-testing With my very minor patch to 299a1ed the peak level sits at +/-0.4 but it is never allowed to exceed it - it appears to be clipped at just over that level, can that clipping level be raised?

@g4dya
FYI Using my build script followed by the included update script you can select a branch, a specific commit and add patches:
```
Do you want to (U)pdate, (F)ull re-build, (B)ackup, (R)estore from backup, (S)witch backup/current, Create (D)esktop launcher or (Q)uit? [U/F/B/R/S/D/Q] 
U
Use a specific branch or tag? [N/y] y
y
Enter branch/tag name:ms-webrtc-agc
Add patch(es)? [N/y] y
y
 NOTE: Patch files must be in /home/baz/freedv-rade/build/ folder
Enter a patch filename (or press Enter to finish): lufs6.patch
Enter a patch filename (or press Enter to finish): 

Use a specific commit? [N/y] y
y
Enter the commit git#:299a1ed

You already have a backup of a previous FreeDV build.
Replace the current backup with a new one of the current version? [N/y] 
N
Summary of selected tasks:
###########################
FreeDV will be built at ms-webrtc-agc and patched with lufs6.patch at commit 299a1ed

Perform the selected tasks? [Y/n] 
Y
```

g4dya:
With f6fd535 I'm seeing this:

-30 LUFS in -23.2 LUFS out
-25 LUFS in -19.7 LUFS out
-20 LUFS in -18.2 LUFS out
-15 LUFS in -18.1 LUFS out
-10 LUFS in -17.8 LUFS out

This is puzzling because these levels are 2.2 to 4 LU lower than before but I was expecting 4 LU across the board.

At this stage it would help to understand what WebRTC_AGC is doing, audio-wise, which currently I don't. It seems to be acting like a compressor/limiter, possibly gated, rather than a slow-acting loudness normalizer. A fast compressor/limiter will inevitably increase the 'density' (i.e. reduce the PAPR) and the resulting flat top on the 'From mic' display might look like clipping but probably isn't. I also don't understand how the Fargan coding-decoding process might affect loudness levels.

Tyrbiter:
> At this stage it would help to understand what WebRTC_AGC is doing, audio-wise, which currently I don't. It seems to be acting like a compressor/limiter, possibly gated, rather than a slow-acting loudness normalizer. A fast compressor/limiter will inevitably increase the 'density' (i.e. reduce the PAPR) and the resulting flat on the 'From mic' display might look like clipping but probably isn't. I also don't understand how the Fargan coding-decoding process might affect loudness levels.

That is something that David should consider in his RADEV2 dev code as it could negatively impact his carefully crafted waveform.

barjac:
> At this stage it would help to understand what WebRTC_AGC is doing, audio-wise, which currently I don't. It seems to be acting like a compressor/limiter, possibly gated, rather than a slow-acting loudness normalizer. A fast compressor/limiter will inevitably increase the 'density' (i.e. reduce the PAPR) and the resulting flat on the 'From mic' display might look like clipping but probably isn't. I also don't understand how the Fargan coding-decoding process might affect loudness levels.

My thought is that this testing should initially be done using the AGC logic in isolation without involving the rade encode/decode in the mix. Could it be split out into a stand-alone for testing?

Tyrbiter:
> > At this stage it would help to understand what WebRTC_AGC is doing, audio-wise, which currently I don't. It seems to be acting like a compressor/limiter, possibly gated, rather than a slow-acting loudness normalizer. A fast compressor/limiter will inevitably increase the 'density' (i.e. reduce the PAPR) and the resulting flat on the 'From mic' display might look like clipping but probably isn't. I also don't understand how the Fargan coding-decoding process might affect loudness levels.
> 
> My thought is that this testing should initially be done using the AGC logic in isolation without involving the rade encode/decode in the mix. Could it be split out into a stand-alone for testing?

See my previous comment, it's all part of a chain and I don't know the interactions (which also vary depending on the radio Tx chain at the end).

barjac:
Yes I did read your comment, but to better understand the operation of the current method it would help to play with it in isolation first. That is what I was meaning ;)

Tyrbiter:
> Yes I did read your comment, but to better understand the operation of the current method it would help to play with it in isolation first. That is what I was meaning ;)

I have not yet tried to understand the webrtc AGC documentation, it may be that what we want is actually only detailed in the code rather than actually being written down in simple terms.

tmiw:
Unfortunately the documentation I've seen is basically StackOverflow posts and the Git repo I mentioned in the PR description. :(

I just integrated Barry's changes and backed out the 4dB pad. I also reduced `compressionGaindB` to 17 as this seems to be the maximum compression applied to the signal.

Tyrbiter:
> Unfortunately the documentation I've seen is basically StackOverflow posts and the Git repo I mentioned in the PR description. :(
>

That was what was worrying me, the repo has essentially nothing other than the base code.
 
> I just integrated Barry's changes and backed out the 4dB pad. I also reduced `compressionGaindB` to 17 as this seems to be the maximum compression applied to the signal.

Well perhaps this will provide a good result, I expect it will get tested very soon.

Can you merge in master again so we have the reporter message changes in too.

tmiw:
> Can you merge in master again so we have the reporter message changes in too.

Done.

g4dya:
It looks as though WebRTC_AGC is just a compressor/limiter. In order to match loudness between stations using it and those that set manually to ±0.4, a pad after the AGC will need to be used. If the resulting low level on the 'From Mic' display is unacceptable then this approach probably won't work. It will need a loudness leveler, possibly based on libebur128, not a compressor/limiter.

Tyrbiter:
> It looks as though WebRTC_AGC is just a compressor/limiter. In order to match loudness between stations using it and those that set manually to ±0.4, a pad after the AGC will need to be used. If the resulting low level on the 'From Mic' display is unacceptable then this approach probably won't work. It will need a loudness leveler, possibly based on libebur128, not a compressor/limiter.

So many different ways of getting to almost the right solution ;-)

At least libebur128 is available for Fedora (and probably many other distros) and I have it installed.

barjac:
> It looks as though WebRTC_AGC is just a compressor/limiter. In order to match loudness between stations using it and those that set manually to ±0.4, a pad after the AGC will need to be used. If the resulting low level on the 'From Mic' display is unacceptable then this approach probably won't work. It will need a loudness leveler, possibly based on libebur128, not a compressor/limiter.


This sounds similar to what I mentioned back in my first comment #3 at the start of this PR discussion after spending some time talking to chatGPT.

Suggested approach:

Audio input (mic)
→ Tap A: VAD (webrtcvad) → Speech activity flag
→ Tap B: BS.1770 loudness meter (libebur128 or pyloudnorm) → Measured LUFS
→ Gain controller (only updates when Speech activity == true) → Smooth (attack/release)
→ Limiter

I have not had time to test this again fully but it was basically working with FreeDV.

tmiw:
If we were to go with `libebur128`, would that be in addition to WebRTC_AGC or instead of it?

barjac:
> If we were to go with `libebur128`, would that be in addition to WebRTC_AGC or instead of it?

We doubled ;)
See above.

g4dya:
libebur128 would do loudness normalization very well, but relatively slowly. I think we would only need WebRTC_AGC if limiting is needed to prevent clipping. With a target loudness of -20 LUFS there would hopefully be sufficient headroom to keep the risk of clipping low enough to ignore.

Note libebur128 is only a measurement algorithm that will report the loudness in LUFS - this information will then need to be fed to a separate gain control that will gradually ramp the input gain up or down to bring the measured LUFS towards the target level. That said, it implements ITU-R BS.1770 which is the gold standard, de facto standard and state of the art in loudness measurement.

barjac:
Hi Richard, I have a working version of what I described earlier using libebur128 for measurement and a simple CLI live monitor to show its operation.

You could maybe take some measurements on it in isolation and tweak it as required as proof of concept.

I will email it to you shortly.

g4dya:
ebumeter is essentially a gui front end for libebur128. It's a JACK client so you might need to install pipewire-jack to provide the interface.

For a 1 kHz sine wave, the loudness in LUFS is equal to the level in dBFS.

IIRR the  BS.1770 measurement process passes the audio through the K-filter weighting and then measures the RMS level in 100 ms chunks. It then takes a rolling average of the last 4 valid chunks. Any chunk which is more than 10 dB below this rolling average is considered invalid. Thus during silences the measured loudness is frozen. That's how the gating works and it's built into libebur128. See Annex 1 in this:

https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.1770-5-202311-I!!PDF-E.pdf

For mono there's no channel summing.

tmiw:
I just pushed a first pass at an AGC using libebur128. I'm sure there are still problems with it, but very brief testing seems to indicate that things are mostly working okay.

BTW in the top of AgcStep.cpp I have the following configuration based on previous comments in this PR:

```c++
constexpr float AGC_LOUDNESS_TARGET_LUFS = -23.0;
constexpr float AGC_MAX_GAIN_DB = 12.0;
constexpr float AGC_MIN_GAIN_DB = -20;
constexpr float AGC_ATTACK_TIME_SEC = 6.0;
constexpr float AGC_RELEASE_TIME_SEC = 0.5;
constexpr float SILENCE_THRESHOLD_LUFS = -70.0; // from https://en.wikipedia.org/wiki/EBU_R_128
```

Feel free to suggest values that work better :)

barjac:
This is looking really promising, however there is still a bug in the Mic/Spkr slider that has been causing some confusion.
To reproduce:
Set the slider manually using the mouse to zero
Transmit and note the output level approx in 'from mic'.
Move the slider up to some crazy high value while stil transmitting
Use the mouse and middle button click  to centre the slider
The graphic re-centres correctly as do the digits in the widget, but the gain does not change!
Only dragging the slider actually affects the gain and level shown in the 'from'mic'.

This was tested in this branch @ #a90e783 I will now check in master and create an issue if it is there too.

Opened issue #1027

g4dya:
A quick test indicates that this works well and certainly a vast improvement on the earlier libraries tried.

constexpr float AGC_ATTACK_TIME_SEC = 6.0;
constexpr float AGC_RELEASE_TIME_SEC = 0.5;

I'm puzzled by the terminology here, which seems to be the wrong way round. In audio circles, the attack time is for gain reduction when the input is too high, and the release time is for gain increase when the input is too low.

For the UI, I suggest that the 'Level' bargraph should show the level before AGC rather than after it.

g4dya:
I also suggest that the mic gain control should be applied before the AGC rather than after it. This is necessary to match the loudness between stations. If necessary any tweak should be made to AGC_LOUDNESS_TARGET_LUFS rather than being user-tweakable. Initial tests suggest -23 LUFS is a good figure.

Tyrbiter:
> A quick test indicates that this works well and certainly a vast improvement on the earlier libraries tried.
> 
> constexpr float AGC_ATTACK_TIME_SEC = 6.0; constexpr float AGC_RELEASE_TIME_SEC = 0.5;
> 
> I'm puzzled by the terminology here, which seems to be the wrong way round. In audio circles, the attack time is for gain reduction when the input is too high, and the release time is for gain increase when the input is too low.
> 

The latest code does this:

if (targetGainDb_ < currentGainDb_)
                {
                    agcInterval = AGC_RELEASE_TIME_SEC;
                }
                else
                {
                    agcInterval = AGC_ATTACK_TIME_SEC;
                }

Now that seems backwards in terms of the description of the variables, but it should function correctly because the variable definitions are also reversed.

I think that if we make constexpr float AGC_ATTACK_TIME_SEC = 0.5; constexpr float AGC_RELEASE_TIME_SEC = 6.0;

and then change the code above to (targetGainDb_ > currentGainDb_)

then the code should both read and function correctly.

> For the UI, I suggest that the 'Level' bargraph should show the level before AGC rather than after it.

Makes sense, so the user can see their level adjustment working.

tmiw:
Definition names have been moved around accordingly, as well as volume control/level display happening based on pre-AGC audio. See [c7fc751](https://github.com/drowe67/freedv-gui/pull/1024/commits/c7fc7517afdd9bf0e7df49e35f76e8daf28fb556).

g4dya:
@tmiw thanks for all the recent improvements.

Now that we have libebur128-based AGC and are only using WebRTC_AGC as a limiter, it's probably safe to relax the limiting from -3 dBFS to -1 dBFS, i.e:

-agcConfig_.targetLevelDbfs = 3; // default 3 (-3 dBOv)
+agcConfig_.targetLevelDbfs = 1; // default 3 (-3 dBOv)

Then the waveform won't look "clipped" to about ±0.7 on the 'Frm mic' display.

As this display is post-AGC, it's no longer strictly from the mic so might it be more accurate to label the tab "To coder"?

On a related issue, might it be better for the 'To Spkr/Hdphns' tab to show the level before the 'Spkr Level' gain control and label it 'From decoder'? The rationale is that people tend to use it to observe and comment on the level from another station and doing this post the Spkr level control could confuse this observation.

On the UI theme, some operators get confused by the Analog/Digital button label, as it's unclear whether it indicates the current status or what will happen if the button is pushed. This is because, unlike 'Start', 'Stop' or 'PTT', the label lacks a verb describing the action. I suggest "Switch to analog" and "Switch to digital" labels would remove the ambiguity.

Similarly, the 'Voice Keyer' button could usefully be labelled 'Start voice keyer' or 'Stop voice keyer' depending on the action that will happen.

tmiw:
> @tmiw thanks for all the recent improvements.

👍 

> Now that we have libebur128-based AGC and are only using WebRTC_AGC as a limiter, it's probably safe to relax the limiting from -3 dBFS to -1 dBFS, i.e:
> 
> -agcConfig_.targetLevelDbfs = 3; // default 3 (-3 dBOv) +agcConfig_.targetLevelDbfs = 1; // default 3 (-3 dBOv)
> 
> Then the waveform won't look "clipped" to about ±0.7 on the 'Frm mic' display.

Done.

> As this display is post-AGC, it's no longer strictly from the mic so might it be more accurate to label the tab "To coder"?

This is actually pre-AGC as of [206a4d4](https://github.com/drowe67/freedv-gui/pull/1024/commits/206a4d4fcba0ad4acb1143a6c0e520a9ea217f17).

> On a related issue, might it be better for the 'To Spkr/Hdphns' tab to show the level before the 'Spkr Level' gain control and label it 'From decoder'? The rationale is that people tend to use it to observe and comment on the level from another station and doing this post the Spkr level control could confuse this observation.

Not sure about this one, will have to think some more.

> On the UI theme, some operators get confused by the Analog/Digital button label, as it's unclear whether it indicates the current status or what will happen if the button is pushed. This is because, unlike 'Start', 'Stop' or 'PTT', the label lacks a verb describing the action. I suggest "Switch to analog" and "Switch to digital" labels would remove the ambiguity.
> 
> Similarly, the 'Voice Keyer' button could usefully be labelled 'Start voice keyer' or 'Stop voice keyer' depending on the action that will happen.

Updated accordingly in [a65e86e](https://github.com/drowe67/freedv-gui/pull/1024/commits/a65e86e7fd1a8cebdb12a8f0f5358675ad29ad65).

Tyrbiter:
Something is causing freedv-gui to use a lot more CPU than it was, it's OK with https://github.com/drowe67/freedv-gui/pull/1024/commits/c7fc7517afdd9bf0e7df49e35f76e8daf28fb556 but the usage jumps from ~50% of one core to as much as 110% but mostly about 90-95% with https://github.com/drowe67/freedv-gui/pull/1024/commits/a65e86e7fd1a8cebdb12a8f0f5358675ad29ad65

It doesn't make much sense to me, but I would guess that this might be caused by https://github.com/drowe67/freedv-gui/pull/1024/commits/dae36baebf10659fbc6378c8ae0921b1fa9f418c

I commented out lines src/voicekeyer.cpp:392 to :397 and rebuilt, this now runs with normal CPU usage. Over to you @tmiw

barjac:
> Something is causing freedv-gui to use a lot more CPU than it was, it's OK with [c7fc751](https://github.com/drowe67/freedv-gui/commit/c7fc7517afdd9bf0e7df49e35f76e8daf28fb556) but the usage jumps from ~50% of one core to as much as 110% but mostly about 90-95% with [a65e86e](https://github.com/drowe67/freedv-gui/commit/a65e86e7fd1a8cebdb12a8f0f5358675ad29ad65)
> 
> It doesn't make much sense to me, but I would guess that this might be caused by [dae36ba](https://github.com/drowe67/freedv-gui/commit/dae36baebf10659fbc6378c8ae0921b1fa9f418c)
> 
> I commented out lines src/voicekeyer.cpp:392 to :397 and rebuilt, this now runs with normal CPU usage. Over to you @tmiw

All looks normal here with a65e86, top shows around 50% of one core on RX and around 24% on TX with or without speech.

Also checked its the same when using VK after noticing your comment about dae36ba.

g4dya:
On a65e86e I'm seeing more or less the same as @barjac

Tyrbiter:
Perhaps it's something in Fedora, at present my main machine is running Fedora 42 (my 'server' machine is now on Fedora 43 beta but with a 2007 Core2Duo CPU it's not up to modern tasks).

Let's see what @tmiw says, I suspect that somehow the button state checking is getting stuck in a tight loop somehow. My results are on receive when the VK status should not matter unless a new file is being selected or recorded.

I do have PR #1028 rolled into my test build, but I saw the same thing without it.

tmiw:
Is [719cb3f](https://github.com/drowe67/freedv-gui/pull/1024/commits/719cb3f59f02b37563e3013170424a7907eafb82) any better for you, @Tyrbiter?

Tyrbiter:
> Is [719cb3f](https://github.com/drowe67/freedv-gui/pull/1024/commits/719cb3f59f02b37563e3013170424a7907eafb82) any better for you, @Tyrbiter?

It appears to have the CPU reduced back down into the 35-60% range depending on exactly what it's doing but typically in the 40-50% range on Rx. The same with PR #1028 included on top of this PR.

It's interesting how the Rx CPU load goes from ~30% up to ~60% without anything appearing to cause it, I don't click anything or interact with the GUI.

Tyrbiter:
A thought; after changing the button text for the Analog/Digital and VK buttons, should we change Start/Stop to include the word 'modem'?

I made this change locally and to me it makes the buttons in the Control box look better balanced in terms of the text width, I don't see any need to do the same with 'PTT' unless someone thinks that a retro 'Push To Talk' would be better.

However...

Screenshot From 2025-09-14 19-13-05

barjac:
> > @tmiw thanks for all the recent improvements.
> 
> 👍
> 
> > Now that we have libebur128-based AGC and are only using WebRTC_AGC as a limiter, it's probably safe to relax the limiting from -3 dBFS to -1 dBFS, i.e:
> > -agcConfig_.targetLevelDbfs = 3; // default 3 (-3 dBOv) +agcConfig_.targetLevelDbfs = 1; // default 3 (-3 dBOv)
> > Then the waveform won't look "clipped" to about ±0.7 on the 'Frm mic' display.
> 
> Done.
> 
> > As this display is post-AGC, it's no longer strictly from the mic so might it be more accurate to label the tab "To coder"?
> 
> This is actually pre-AGC as of [206a4d4](https://github.com/drowe67/freedv-gui/pull/1024/commits/206a4d4fcba0ad4acb1143a6c0e520a9ea217f17).

Really? I don't think it should be, as it will confuse everyone, including anyone testing. Without a separate output from FreeDV at the input to the encoder, its the only way to see what's really going on in the AGC.

>
> > On a related issue, might it be better for the 'To Spkr/Hdphns' tab to show the level before the 'Spkr Level' gain control and label it 'From decoder'? The rationale is that people tend to use it to observe and comment on the level from another station and doing this post the Spkr level control could confuse this observation.
> 
> Not sure about this one, will have to think some more.

I agree with Richard on this.

Tyrbiter:
> > > @tmiw thanks for all the recent improvements.
> > 
> > 
> > 👍
> > > Now that we have libebur128-based AGC and are only using WebRTC_AGC as a limiter, it's probably safe to relax the limiting from -3 dBFS to -1 dBFS, i.e:
> > > -agcConfig_.targetLevelDbfs = 3; // default 3 (-3 dBOv) +agcConfig_.targetLevelDbfs = 1; // default 3 (-3 dBOv)
> > > Then the waveform won't look "clipped" to about ±0.7 on the 'Frm mic' display.
> > 
> > 
> > Done.
> > > As this display is post-AGC, it's no longer strictly from the mic so might it be more accurate to label the tab "To coder"?
> > 
> > 
> > This is actually pre-AGC as of [206a4d4](https://github.com/drowe67/freedv-gui/pull/1024/commits/206a4d4fcba0ad4acb1143a6c0e520a9ea217f17).
> 
> Really? I don't think it should be, as it will confuse everyone, including anyone testing. Without a separate output from FreeDV at the input to the encoder, its the only way to see what's really going on in the AGC.
> 
> > > On a related issue, might it be better for the 'To Spkr/Hdphns' tab to show the level before the 'Spkr Level' gain control and label it 'From decoder'? The rationale is that people tend to use it to observe and comment on the level from another station and doing this post the Spkr level control could confuse this observation.
> > 
> > 
> > Not sure about this one, will have to think some more.
> 
> I agree with Richard on this.

Acting as devil's advocate, if the new changes and AGC work as intended, will anyone need to comment on other's levels at all in future?  :-p

tmiw:
> A thought; after changing the button text for the Analog/Digital and VK buttons, should we change Start/Stop to include the word 'modem'?

Seems reasonable to me. I went ahead and pushed this change.

> > This is actually pre-AGC as of [206a4d4](https://github.com/drowe67/freedv-gui/pull/1024/commits/206a4d4fcba0ad4acb1143a6c0e520a9ea217f17).
> 
> Really? I don't think it should be, as it will confuse everyone, including anyone testing. Without a separate output from FreeDV at the input to the encoder, its the only way to see what's really going on in the AGC.

Good point. I reverted this for now; once this is merged and we've had more testing on the AGC we can always do a PR to bring back this change.

> > > On a related issue, might it be better for the 'To Spkr/Hdphns' tab to show the level before the 'Spkr Level' gain control and label it 'From decoder'? The rationale is that people tend to use it to observe and comment on the level from another station and doing this post the Spkr level control could confuse this observation.
> > >
> > Not sure about this one, will have to think some more.
> 
> I agree with Richard on this.

Done.

> Acting as devil's advocate, if the new changes and AGC work as intended, will anyone need to comment on other's levels at all in future? :-p

AGC is optional but enabled by default. Someone will inevitably disable it and mess up the audio levels, so someone will comment about that in the future :)

g4dya:
In ab14cea the Frm Decoder tab is showing the Frm radio waveform.

Also, the audio gain seems to increase during pauses. I haven't seen this in previous builds.  R128 has three measurement intervals. From Wikipedia:

    Momentary (M), using a sliding time window of 400 ms, best describes the instantaneous loudness;
    Short-term (S), using a sliding time window of 3 seconds, describes a more averaged, less event-dependant loudness of the past three seconds;
    Integrated (I), averaging the programme from start to finish, describes the loudness of the whole programme.

The gating function, in which pauses and silences more than 10 dB down are ignored, only applies to the Integrated measurement. (I had forgotten this in my earlier comments.) AIUI we're using the momentary (400ms) measurement, in which case I suggest we freeze AGC gain changes if the measured loudness is below -33 LUFS.

With AGC turned off, it looks from the From mic display as if the WebRTC_AGC limiter is in operation with a target level of -3 dBFS. (It's peaking to a smidge over ±0.6.) Shouldn't it be disabled when AGC is turned off?

tmiw:
> In [ab14cea](https://github.com/drowe67/freedv-gui/commit/ab14cea0edff6d3c086acb0ebbd5f203c63bb7e7) the Frm Decoder tab is showing the Frm radio waveform.

Oops, this should be fixed now.

> The gating function, in which pauses and silences more than 10 dB down are ignored, only applies to the Integrated measurement. (I had forgotten this in my earlier comments.) AIUI we're using the momentary (400ms) measurement, in which case I suggest we freeze AGC gain changes if the measured loudness is below -33 LUFS.

Done.

> With AGC turned off, it looks from the From mic display as if the WebRTC_AGC limiter is in operation with a target level of -3 dBFS. (It's peaking to a smidge over ±0.6.) Shouldn't it be disabled when AGC is turned off?

It is indeed disabled. There's logic to add AGC to the audio pipeline as follows:

```c++
        // AGC step (optional)
        auto eitherOrProcessAgc = new AudioPipeline(inputSampleRate_, inputSampleRate_);
        auto eitherOrBypassAgc = new AudioPipeline(inputSampleRate_, inputSampleRate_);

        auto agcStep = new AgcStep(inputSampleRate_);
        eitherOrProcessAgc->appendPipelineStep(agcStep);

        auto eitherOrAgcStep = new EitherOrStep(
            []() { return g_agcEnabled.load(std::memory_order_acquire); },
            eitherOrProcessAgc,
            eitherOrBypassAgc);
        pipeline_->appendPipelineStep(eitherOrAgcStep); 
```

`g_agcEnabled` is basically an atomic variable copy of the equivalent configuration object for AGC and is supposed to be getting set to `false` if AGC is unchecked in the Filter window (and `true` otherwise). If `false`, `EitherOrStep` above is supposed to just pass through the audio unchanged and not pass it through `AgcStep`.

As for the behavior you're seeing with AGC disabled, I'm not sure why. Is it possible it's something at a lower level (i.e. pipewire/Core Audio/Windows WASAPI) doing AGC itself before passing the recorded audio to FreeDV?

barjac:
> The gating function, in which pauses and silences more than 10 dB down are ignored, only applies to the Integrated measurement. (I had forgotten this in my earlier comments.) AIUI we're using the momentary (400ms) measurement, in which case I suggest we freeze AGC gain changes if the measured loudness is below -33 LUFS.
> 

That explains why chatGPT used external silence detection, since after real time testing 'momentary' was used to reduce latency.

barjac:
I am seeing low level audio in the 'From Decoder' tab during transmissions, not sure if it matters or if it is new.
#e72cf

Tyrbiter:
> I am seeing low level audio in the 'From Decoder' tab during transmissions, not sure if it matters or if it is new. #e72cf

Might need a bit more code tweaking given the number of changes there have been.

tmiw:
> I am seeing low level audio in the 'From Decoder' tab during transmissions, not sure if it matters or if it is new. #e72cf

I don't think it's new but we never noticed until now. Fixed in [2960450](https://github.com/drowe67/freedv-gui/pull/1024/commits/2960450c3ce1183408952a89d86b08b3b14d74a4).

g4dya:
329f496 seems to be working well. Time to merge into master?

Tyrbiter:
> [329f496](https://github.com/drowe67/freedv-gui/commit/329f49693a432fd8e52a8fac8f863e5c9e014f0e) seems to be working well. Time to merge into master?

Good news if you're saying that @g4dya :)

tmiw:
> [329f496](https://github.com/drowe67/freedv-gui/commit/329f49693a432fd8e52a8fac8f863e5c9e014f0e) seems to be working well. Time to merge into master?

Sounds good.

PR #1026, opened 2025-09-11
FreeDV Reporter: Always highlight Msg field when receiving from server.

### LINKED ISSUE #1025: Feature gone misssing? Re-sending the same message in reporter no longer re-triggers the purple notification.
Maybe this was seen as a bug rather than a feature, however it was useful to attempt to draw attention to a message if it appeared that it had be overlooked.

tmiw:
Resolves #1025 by forcing updates whenever it receives a Msg update from the server, even when no changes have been detected to the Msg text.

Note: this also requires a server update which has already been done. See https://bitbucket.org/tmiw/freedv-reporter/commits/7304cf93e7adeec842b2e28b0d3cdcf4bb37e66d.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -856,7 +856,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Add Automatic Gain Control (AGC) to microphone input. (PR #997)
     * Linux: Search for and list serial devices from /dev/serial for PTT config. (PR #999)
     * Add RADEV1 sample file and remove samples for unsupported modes. (PR #998)
-    * Various GUI and FreeDV Reporter performance improvements. (PR #1002)
+    * Various GUI and FreeDV Reporter performance improvements. (PR #1002, #1026)
     * Use rigerror2() for friendlier Hamlib errors. (PR #1020)
 3. Build system:
     * Update Hamlib to v4.6.4. (PR #989)


Tyrbiter:
I've tested this and it seems to work as expected in the freedv-gui reporter window.

I also did a shift-reload on the web version on the server, when I send the same message again from the freedv-gui reporter window it does not appear on the web version.

tmiw:
Actually, I don't think we ever implemented the purple highlighting for messages in the web version. Is that something that would be useful there?

In the meantime, I'll merge this PR as the web updates can be worked on separately.

Tyrbiter:
> Actually, I don't think we ever implemented the purple highlighting for messages in the web version. Is that something that would be useful there?
> 
> In the meantime, I'll merge this PR as the web updates can be worked on separately.

I can't remember if the web version ever showed the message changes, but I think it is a useful feature to be fair, it makes a new station appearing more obvious.

tmiw:
> I can't remember if the web version ever showed the message changes, but I think it is a useful feature to be fair, it makes a new station appearing more obvious.

I took a look at the JavaScript and it definitely wasn't, but it should be now (you may need to Shift-F5 for the changes to take effect).

Tyrbiter:
> > I can't remember if the web version ever showed the message changes, but I think it is a useful feature to be fair, it makes a new station appearing more obvious.
> 
> I took a look at the JavaScript and it definitely wasn't, but it should be now (you may need to Shift-F5 for the changes to take effect).

That also seems to work well, it's slightly unexpected when a newly opened window is nearly all purple for a few seconds but that's not really important.

PR #1028, opened 2025-09-13
Need to notify TX/RX threads when resetting level to zero.

### LINKED ISSUE #1027: Mic/Spkr re-set to zero using centre mouse button does not affect gain
To reproduce:
Set the slider manually using the mouse to zero
Transmit and note the approx output level in 'from mic'.
Move the slider up to some crazy high value while still transmitting (not on air!).
Use the mouse and middle button click to centre the slider.
The graphic re-centres correctly as do the digits in the widget, but the gain does not change!
Only dragging the slider actually affects the gain and level shown in the 'from mic'.

Tested in master branch @ #9559

tmiw:
Resolves #1027 by using the same logic for reset as with manually moving the TX/RX level slider.

### DIFF
--- a/src/ongui.cpp
+++ b/src/ongui.cpp
@@ -1426,14 +1426,21 @@ void MainFrame::updateReportingFreqList_()
 
 void MainFrame::OnResetMicSpkrLevel(wxMouseEvent& event)
 {
-    bool txState = g_tx.load(std::memory_order_relaxed);
-
-    if (txState)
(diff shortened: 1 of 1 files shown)


barjac:
Yes that's fixed it and no regressions running this patched into https://github.com/drowe67/freedv-gui/pull/1024 to check cpu usage which is normal and has not changed.

Issue #941, opened 2025-06-18
[Software Issue] Intermittent resync on start in full duplex mode


tmiw:
### Platform

macOS

### Platform Version

15.4.1

### FreeDV Version

2.0.1-dev-909b108 

### Steps to Reproduce

Set up FreeDV in full duplex mode using a VB-Cable virtual audio cable as the "radio" sound device. Push Start with one of 700D, E or 1600 selected.

### Expected Behavior

Sync counter increments to 1 and does not change until I push Stop.

### Actual Behavior

Sync counter intermittently increments to values greater than one right on startup due to the modem immediately resyncing.

### Additional Comments

Discovered while reviewing remaining failed test cases on #939.

tmiw:
This might now be fixed, or at least worked around, as of the latest in `master`. Closing.

PR #1033, opened 2025-09-18
FreeDV Reporter: Switch to ujson to reduce memory allocations.

tmiw:
This PR is to switch the JSON implementation from https://github.com/nlohmann/json to https://bitbucket.org/awangk/ujson/src/master/, mainly to reduce the number of calls into `malloc()` and hopefully reduce the likelihood of audio glitches.

(WIP - DO NOT MERGE)


Tyrbiter:
Is this intended for 2.0.2 @tmiw ?

tmiw:
> Is this intended for 2.0.2 @tmiw ?

Not sure. There seem to be some disadvantages to this JSON library so I might want to do a bit more research on them.

tmiw:
So I'm pretty sure that the remaining test failures on GitHub are at least partly influenced by the time of day (i.e. I seem to see them far less in the middle of the night my time). I'm still not fully sure if it's related to FreeDV Reporter traffic (for example, if someone's signal is really booming and a lot of people are receiving it) or just something to do with how loaded the GitHub infrastructure is whenever I commit stuff.

Unfortunately, if it's the former, the only real way to fix it would be a full C port of RADE and ensuring that no system calls are made by the TX/RX thread (i.e. no `malloc()`) while it's processing a slice of audio. I made some tweaks to Codec2 a while back to ensure the latter, for instance, which is why it's now included in the FreeDV source tree (and why failures in `fullduplex_[700D|700E|1600]` are far more rare). 

So yeah, just a bit of context here. The hope is by reducing the number of allocations the FreeDV Reporter code does (which is the big source of allocations in the application besides the GUI logic), we can at least partly work around the dropouts until a C port of RADE comes (if it does).

FWIW with this PR FreeDV *might* be losing sync less often in RADE. The most recent commit failed, for instance, but looking at the GitHub logs it looks like both attempts at `rade_loss` failed due to losing sync within the first few seconds of the test (i.e. the same issue outlined in #941). 

EDIT: the current JSON library in this PR might not be ideal because (a) all numbers are `double` (which isn't good for stuff that can't handle precision loss, such as frequencies) and (b) the project hadn't been updated in quite some time and I had to hack it to even get it to run in 2025. Looking into switching to RapidJSON or perhaps another library.

tmiw:
Switched to yyjson instead. I feel better about the current state of the PR since we're going with something that I can just include verbatim (no hacks required). 

@Tyrbiter  / @barjac, we can try to get this into 2.0.2 if everything looks fine (i.e. reporting appears correct and there aren't crashes). I'll watch the GitHub tests and see how they go.

Tyrbiter:
> Switched to yyjson instead. I feel better about the current state of the PR since we're going with something that I can just include verbatim (no hacks required).
> 
> @Tyrbiter / @barjac, we can try to get this into 2.0.2 if everything looks fine (i.e. reporting appears correct and there aren't crashes). I'll watch the GitHub tests and see how they go.

I see yyjson states it's the fastest C-based json there is, I hope it lives up to the billing.

I now have it running here, so let's see how it goes. I have left the Reporter window open, not sure if that increases the CPU load or not.

Tyrbiter:
It's been running here for around 6 hours and nothing out of the ordinary is showing up.

I see all the checks have passed too :)

tmiw:
> It's been running here for around 6 hours and nothing out of the ordinary is showing up.
> 
> I see all the checks have passed too :)

I was still getting some failures in `rade_loss` on my Ubuntu aarch64 VM but the latest commits seem to have helped with that (or I just happened to test during a lull in activity).

BTW overnight it looks like we ran into the hanging test issue again, but of course once I set up instrumentation to be able to look into that it stopped. >< If it doesn't come back in this PR today we can just investigate that more later.

Tyrbiter:
> > It's been running here for around 6 hours and nothing out of the ordinary is showing up.
> > I see all the checks have passed too :)
> 
> I was still getting some failures in `rade_loss` on my Ubuntu aarch64 VM but the latest commits seem to have helped with that (or I just happened to test during a lull in activity).
> 
> BTW overnight it looks like we ran into the hanging test issue again, but of course once I set up instrumentation to be able to look into that it stopped. >< If it doesn't come back in this PR today we can just investigate that more later.

Could the server have a load value added to the web page? Might be useful for remote diagnosis.

tmiw:
> > > It's been running here for around 6 hours and nothing out of the ordinary is showing up.
> > > I see all the checks have passed too :)
> > 
> > 
> > I was still getting some failures in `rade_loss` on my Ubuntu aarch64 VM but the latest commits seem to have helped with that (or I just happened to test during a lull in activity).
> > BTW overnight it looks like we ran into the hanging test issue again, but of course once I set up instrumentation to be able to look into that it stopped. >< If it doesn't come back in this PR today we can just investigate that more later.
> 
> Could the server have a load value added to the web page? Might be useful for remote diagnosis.

I'm currently using https://github.com/mxschmitt/action-tmate to see if I can run a debugger on the FreeDV process when it happens. From what I can tell, FreeDV seems to spike to 200%+ CPU in the GH environment but there's no obvious reason why.

PR #1035, opened 2025-09-23
Reduce memory allocations in main thread.

tmiw:
This PR further reduces the number of allocations performed by non-RADE threads, reducing the likelihood that an allocation will stall RX/TX processing.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -85,7 +85,7 @@ jobs:
 
   test-appimage-fedora:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


Tyrbiter:
Odd, macOS seems peculiarly sensitive to a commit that doesn't actually affect any of the compiled values.

Tyrbiter:
Have this running here now. I see the macOS tests have now run and completed.

tmiw:
> Have this running here now. I see the macOS tests have now run and completed.

Yeah, not sure why disabling mimalloc worked but implementing this another way seems to allow it to be enabled on macOS.

PR #1037, opened 2025-09-26
Fix waterfall flicker on macOS

tmiw:
This PR fixes several issues discovered during 2.0.2 testing on macOS:

1. The orange "average frequency" line wasn't properly being suppressed on startup. This is due to a division by zero error in the waterfall logic.
2. With some monitors/systems, the waterfall displays in a "pulsing" type manner. This appears to be triggered when the system resets the window color space when unneeded (performance improvement in an older FreeDV release).
3. Clear to black logic in the waterfall was drawing more than needed (performance improvement).


Tyrbiter:
Is this needed for 2.0.2 @tmiw ?

I presume it is...

tmiw:
> Is this needed for 2.0.2 @tmiw ?
> 
> I presume it is...

Yep. Noticed some rendering bugs on my M1 Mac (but weirdly not on my Intel laptop, not sure why).

Anyway, this PR doesn't fix the ARM problem yet, I ran out of time this morning. Will hopefully investigate more later today.

tmiw:
Going to see what the GitHub tests return and then merge.

PR #1053, opened 2025-10-07
Reduce the number of context switches during heavy FreeDV Reporter load.

tmiw:
Updates the code to only queue updates to the FreeDV Reporter window once there's no longer any data pending in the WebSocket connection. Additionally:

1. Resolves additional ThreadSanitizer warnings on macOS and Linux.
2. Updates `ThreadedTimer` to use GCD timers (macOS only).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -849,6 +849,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 1. Bugfixes:
     * Clean up wxWidgets warnings in Audio Config window. (PR #1044)
     * Fix CAT control and frequency formatting problems on non-English computers. (PR #1050)
+    * Clean up ThreadSanitizer warnings in codebase. (PR #1053)
 2. Enhancements:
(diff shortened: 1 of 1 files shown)


Tyrbiter:
Have been running this branch for a day or so, so far no obvious problems but how to tell if it is doing something useful? Anything in the debug output that might be useful?

tmiw:
> Have been running this branch for a day or so, so far no obvious problems but how to tell if it is doing something useful? Anything in the debug output that might be useful?

I've mainly been using the Instruments program (comes with Xcode on macOS). Example:

image

versus

image

Unfortunately there is a timing component to this as FreeDV Reporter traffic depends on how many people are on and performing various events.

Tyrbiter:
From the graphs it seems to be a good thing, as I say it seems to run fine here.

There is a lot of activity on the reporter, perhaps we need to think of a way to reduce the impact of that in general. As for testing, there's no real need to build a Reporter simulator with adjustable load factors ;-)

PR #1057, opened 2025-10-10
Hamlib: Add support for automatically displaying only the valid baud rates.

tmiw:
This PR adds support for only showing the radio's valid baud rates based on Hamlib `rig_caps` data.


Tyrbiter:
A quick test of a few different radios reveals that this appears to work for the baud rate drop down.

Issue #1062, opened 2025-10-13
We may not want to unconditionally force GDK_BACKEND=x11 for all systems


tmiw:
Re: AppImage, but probably also applies to @barjac's scripts.

At Pacificon, I noticed that adding `GDK_BACKEND=x11` to the command line for FreeDV with the demo we had at the booth caused massive slowdowns to the UI. Additionally, audio was also glitching. Neither of these things occurred when I left off `GDK_BACKEND` and allowed native Wayland to be used.

The configuration used:

* Raspberry Pi 5 with Raspberry Pi OS (bookworm)
* Bluetooth speaker for audio output
* IC-705 as the RX radio (TX was disabled), no CAT control configured
* 1080p HDMI to portable projector (though this also occurred on an external display I brought as a backup)

Questions:

1. Have you noticed similar behavior on other Linux systems? Or was this just some weird Raspberry Pi thing I ran into?
2. If yes to (1), are there enough commonalities that we could add checks to force native Wayland instead of XWayland?

(CC @Tyrbiter as well.)

Tyrbiter:
I get an error when I try to do this, currently I start freedv-gui from a script and the result is:

Authorization required, but no authorization protocol specified

10:10:51: Error: Unable to initialize GTK+, is DISPLAY set properly?

Modified script is this:

#!/bin/sh
cd ~/freedv2
GDK_BACKEND=x11 DISPLAY=:0.0 PYTHONPATH="$(pwd)/rade_src:$PYTHONPATH" src/freedv

There seems to be a long standing problem with accidental deletion of ~/.Xauthority and other similar files. I have not yet fixed it.

tmiw:
I've seen that before but I'm not fully sure how to fix that. I can say that I was able to start FreeDV over SSH with the proper `DISPLAY` variable (with the Pi).

Tyrbiter:
I think it's down to the X11 auth secret having been deleted, but it's system specific so I can't just copy it from my other machine.

In any case, X11 is going away and Wayland is what is going to be developed in future so we have to go in that direction.

tmiw:
> I think it's down to the X11 auth secret having been deleted, but it's system specific so I can't just copy it from my other machine.
> 
> In any case, X11 is going away and Wayland is what is going to be developed in future so we have to go in that direction.

Yep. IIRC the original reason why we added `GDK_BACKEND` was to allow persistence of window positions. Maybe we can replace with additional documentation about how to persist at the compositor level?

[...]

PR #1064, opened 2025-10-15
Reduce CPU usage in Xwayland process when using AppImage.

tmiw:
Based on discussion in #1062, this PR reduces the amount of CPU used by the Xwayland process while rendering the waterfall and Mic In plots. This is accomplished by consolidating large number of draw operations into temporary contexts first prior to displaying on the screen.

Testing with `PYTHONPATH=$(pwd)/rade_src GDK_BACKEND=x11 ctest -R rade_loss` and `top` on a Fedora 42 VM indicates an approximately 1/3-1/2 reduction in CPU usage in the Xwayland process during test execution (for example, during the TX portion of the `rade_loss` test, CPU usage ranged from 15-20% versus ~30% without this PR).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -826,6 +826,14 @@ xattr -d -r com.apple.quarantine FreeDV.app
 
 macOS has a longstanding issue that prevents certain audio devices from appearing. More info [here](http://www.w1hkj.com/files/fldigi/AppleAudioCodec.html).
 
+## FreeDV doesn't preserve window positions on Linux
+
+Many Linux distributions are switching away from X11 and towards a new GUI system known as Wayland.
+Wayland explicitly does not support window positioning by applications, instead leaving it to the user's
+compositor to control window positioning. You can force use of X11 via the Xwayland compatibility layer
+by adding `GDK_BACKEND=x11` to your environment to regain the ability to preserve window positioning
+if desired.
+
 # Converting this document to PDF
 
 For the Linux inclined:


Tyrbiter:
If this reduces CPU load then it's a good thing, although a consequence of that is that the spectrum plot becomes a little bit more jerky and sometimes it stalls for longer than it used to.

It's interesting that the same stall does not happen with the waterfall plot, not sure why the two are different in this regard. I happen to like the spectrum plot but am not going to complain about this new compromise between the responsiveness vs CPU usage.

tmiw:
> If this reduces CPU load then it's a good thing, although a consequence of that is that the spectrum plot becomes a little bit more jerky and sometimes it stalls for longer than it used to.
> 
> It's interesting that the same stall does not happen with the waterfall plot, not sure why the two are different in this regard. I happen to like the spectrum plot but am not going to complain about this new compromise between the responsiveness vs CPU usage.

I'm noticing that too. Interesting that it's behaving that way despite supposedly using the same data source. I think I have some ideas on making that better. 

Re: performance, the Xwayland version also seems to render a bit differently so I'm still considering getting rid of the override in the AppImage anyway (and documenting the limitations of Wayland if we're not already doing that).

Tyrbiter:
https://github.com/drowe67/freedv-gui/pull/1064/commits/14e9513ad9eacacfa1949fd46d2489ef8c4a9ed4 seems to run in a less jerky fashion here, I'm not sure about the CPU load reduction but it's showing a little less when showing the waterfall vs the spectrum.

https://github.com/drowe67/freedv-gui/pull/1064/commits/eb0970c1129e67e5412127c90f202630855b6731 is even better but I have only been running it for a short time, it's visibly improved.

Tyrbiter:
Screenshot From 2025-10-18 18-50-27

This is what I am now seeing, the visual aspect is much grittier so there needs to be something a bit less aggressive for https://github.com/drowe67/freedv-gui/pull/1064/commits/1d0478042dcc1254c07dac5b745712ec1672d3da

tmiw:
> This is what I am now seeing, the visual aspect is much grittier so there needs to be something a bit less aggressive for [1d04780](https://github.com/drowe67/freedv-gui/commit/1d0478042dcc1254c07dac5b745712ec1672d3da)

Any better?

Tyrbiter:
> > This is what I am now seeing, the visual aspect is much grittier so there needs to be something a bit less aggressive for [1d04780](https://github.com/drowe67/freedv-gui/commit/1d0478042dcc1254c07dac5b745712ec1672d3da)
> 
> Any better?

Not really much different. Note the left hand edge where the line comes down from 0dB and then jumps directly across to just below 500Hz on the x-axis. Something is not being calculated correctly I think.

tmiw:
Yeah, I saw a typo in the code from the previous few commits and fixed that in the latest. It seemed to look fine for me because I was running unit tests but I could see how it would be a problem if you were running narrower filters. Hopefully it looks better for you now?

Tyrbiter:
Ah, yes, I see why the typo was bad.

This is partially fixed, but the left side of the plot still has an errant value which means the line starts at the corner instead of somewhere along the x-axis close to where the spectrum value rises above the -40dB level:

Screenshot From 2025-10-18 22-35-59

barjac:
#5400ab3 has fixed RPi4b crash - thanks.

This branch under wayland/xwayland on x86_64 has reduced cpu load by a few percent in both TX and RX.

Tyrbiter:
https://github.com/drowe67/freedv-gui/pull/1064/commits/ef8776275c562839d8c4f087c0b130e947c9a87a seems to have nailed it.

tmiw:
I've been doing some testing on the Pi 5 as well and FWIW, the CPU usage of Xwayland seems to match up with my Fedora VM. I'm not 100% sure yet whether the system-wide GUI slowdown/audio glitching is fixed there or if it was actually caused by something else (such as heavy 2.4 GHz traffic causing audio drops to the BT speaker I was using).

@barjac, how's the Pi 4? Is RADE actually usable there now?

EDIT: I'd say Xwayland is using a *maximum* of 15% CPU with this branch. Usually it's around 10% or lower based on unit test runs I've been doing.

Tyrbiter:
> EDIT: I'd say Xwayland is using a _maximum_ of 15% CPU with this branch. Usually it's around 10% or lower based on unit test runs I've been doing.

I can't confirm the figures precisely but there has been a significant improvement in the feel of the spectrum display and a reduction in CPU load at the same time. Good job @tmiw

barjac:
> I've been doing some testing on the Pi 5 as well and FWIW, the CPU usage of Xwayland seems to match up with my Fedora VM. I'm not 100% sure yet whether the system-wide GUI slowdown/audio glitching is fixed there or if it was actually caused by something else (such as heavy 2.4 GHz traffic causing audio drops to the BT speaker I was using).
> 
> @barjac, how's the Pi 4? Is RADE actually usable there now?
> 
No, I spent some time trying to set one up but the radio mouse is interfered with such that it is unusable with the modem running. A wired mouse works fine but then I ran out of USB ports. Its hard wired ethernet so not wi-fi causing the mouse issue.
All very flakey, crashes a lot and very frustrating.
Even worse when trying to run rigctld on it.
I finally used the desktop machine to serve rigctld which saved a USB port and net rigctl worked OK for PTT, but still had no success getting working audio.
The RX audio in the 'from radio' window has lots of gaps in it like inverted morse. Sadly there were no signals around when I think it may have tried to decode. I will do more testing with a recorded file. But its not looking likely to be usable. The reporter works fine :)
I did not force xwayland - that may be worth a try.

tmiw:
> > I've been doing some testing on the Pi 5 as well and FWIW, the CPU usage of Xwayland seems to match up with my Fedora VM. I'm not 100% sure yet whether the system-wide GUI slowdown/audio glitching is fixed there or if it was actually caused by something else (such as heavy 2.4 GHz traffic causing audio drops to the BT speaker I was using).
> > @barjac, how's the Pi 4? Is RADE actually usable there now?
> 
> No, I spent some time trying to set one up but the radio mouse is interfered with such that it is unusable with the modem running. A wired mouse works fine but then I ran out of USB ports. Its hard wired ethernet so not wi-fi causing the mouse issue. All very flakey, crashes a lot and very frustrating. Even worse when trying to run rigctld on it. I finally used the desktop machine to serve rigctld which saved a USB port and net rigctl worked OK for PTT, but still had no success getting working audio. The RX audio in the 'from radio' window has lots of gaps in it like inverted morse. Sadly there were no signals around when I think it may have tried to decode. I will do more testing with a recorded file. But its not looking likely to be usable. The reporter works fine :) I did not force xwayland - that may be worth a try.

Based on some testing with the upcoming Flex waveform, I suspect it might at least partially be because mimalloc is getting built in Debug mode instead of Release. You can try pulling in the BuildMimalloc.cmake changes from https://github.com/drowe67/freedv-gui/pull/1018/commits/33e25bbac2e761393003cd64b7ad876051427a8d and seeing how that goes. If that doesn't help, you might also need to compile FreeDV itself in RelWithDebInfo or Release--i.e. `BUILD_TYPE=Release ./build_linux.sh`.

PR #1080, opened 2025-11-05
Enable frequency drop-down whenever CAT control is enabled, not just for reporting.

tmiw:
Updates user interface to enable the frequency drop-down in the lower-right of the main window whenever CAT control is configured, not just when reporting is enabled. This reduces user confusion and enables app control of the radio frequency for those who don't want to enable reporting.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -869,6 +869,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Add support for BBWENet BW expander. (PR #1065)
     * Support bulk updates from FreeDV Reporter server. (PR #1079)
     * Hide Stats box by default. (PR #1081)
+    * Enable frequency drop-down whenever CAT control is enabled, not just for reporting. (PR #1080)
 3. Build system:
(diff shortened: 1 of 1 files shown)


barjac:
Hmm...
I often use a remote web SDR to listen to stations on my laptop when not in the shack.
In order to use the reporter and filters to see the band/frequency in the reporter that I am listening to, I have to manually set the frequency using the frequency drop down.
If it was not there because I am not using CAT control I could not do this.
I think the frequency selector should always be available whether reporting or not, CAT or not.

barjac:
Just been testing this branch on the laptop using a web SDR as mentioned above.

If I want to report I can, however I cannot change the reported frequency from the one last saved in the config from before this PR.

I was just listening to a FDV station on 40m but I was reporting our regular 60m frequency.

In the reporting tab I have:
Reporting enabled
Report to FreeDV Reporter enabled
Enabling 'Use manual frequency reporting' it now makes no difference as the widget is still missing.

In CAT and PTT I have all disabled.

tmiw:
I brought back the original behavior of the drop-down as of the latest commit (along with allowing it to be enabled if reporting isn't but CAT control is).

barjac:
This is fine now regarding my previous issue. No problems using remote SDR without CAT and with/without reporting.

PR #1081, opened 2025-11-05
Hide Stats box by default.

tmiw:
This PR implements a request to reduce the minimum amount of vertical space required by hiding the Stats box by default. This can be reenabled by going to Tools->Options->Debugging and checking "Show Decode Stats".

### DIFF
--- a/test/test_zeros.sh
+++ b/test/test_zeros.sh
@@ -37,13 +37,6 @@ if [ "$OPERATING_SYSTEM" == "Linux" ]; then
     DRIVER_INDEX_FREEDV_COMPUTER_TO_SPEAKER=$(createVirtualAudioCable FreeDV_Computer_To_Speaker)
     DRIVER_INDEX_FREEDV_MICROPHONE_TO_COMPUTER=$(createVirtualAudioCable FreeDV_Microphone_To_Computer)
     DRIVER_INDEX_FREEDV_COMPUTER_TO_RADIO=$(createVirtualAudioCable FreeDV_Computer_To_Radio)
-<<<<<<< Updated upstream
-
-    # Make sure cables are actually created before proceeding with looping them back
-    sleep 5
-
-=======
->>>>>>> Stashed changes
     DRIVER_INDEX_LOOPBACK=`pactl load-module module-loopback source="FreeDV_Computer_To_Radio.monitor" sink="FreeDV_Radio_To_Computer"`
 fi
 


Tyrbiter:
Note that if I enable the stats in the Debugging tab it requires restarting the program for the stats box to appear.

Is this intentional?

tmiw:
> Note that if I enable the stats in the Debugging tab it requires restarting the program for the stats box to appear.
> 
> Is this intentional?

Not intentional. Does the latest in this PR work any better for you?

Tyrbiter:
> > Note that if I enable the stats in the Debugging tab it requires restarting the program for the stats box to appear.
> > Is this intentional?
> 
> Not intentional. Does the latest in this PR work any better for you?

Yes, enabling stats and selecting OK or Apply shows the stats box now.

barjac:
Could you do the same for the 'Mode' on the other side as 99% of users don't need it and it would balance up the display so the scope is back central :)

tmiw:
> Could you do the same for the 'Mode' on the other side as 99% of users don't need it and it would balance up the display so the scope is back central :)

I think we still need to discuss when/how we're going to start removing the other modes before we start hiding that box by default.

barjac:
> > Could you do the same for the 'Mode' on the other side as 99% of users don't need it and it would balance up the display so the scope is back central :)
> 
> I think we still need to discuss when/how we're going to start removing the other modes before we start hiding that box by default.

I agree.
If it was not by default but selectable, it would be useful to also preview how the GUI will look and feel going forward.
The fact that those options are still there has already started to confuse some new users.

Tyrbiter:
Perhaps the mode box could offer the option to change to display the legacy modes when the modem is stopped.

It would act like the other buttons and would have 'Switch to RADEV1' with legacy modes enabled and 'Switch to Codec2' with RADEV1 enabled, greyed out when modem is running. Not sure how easy a change of box controls would be.

barjac:
> Perhaps the mode box could offer the option to change to display the legacy modes when the modem is stopped.
> 

The point of my comment was to be able to remove the mode box which is creating a complete widget column when the window height is reduced.

It is not used by 99% of users now so why keep it on show? Legacy modes could be selected in a menu option with the modem stopped, which it has to be anyway.

Mode switching in wsjtx is not done in the main GUI, it is a menu option.

> It would act like the other buttons and would have 'Switch to RADEV1' with legacy modes enabled and 'Switch to Codec2' with RADEV1 enabled, greyed out when modem is running. Not sure how easy a change of box controls would be.

'Switch to Codec2' would mean nothing to many users.

tmiw:
I actually just saw an ezDV user on FreeDV Reporter right now. Not sure how many there are but more steps to enable legacy modes means that it'll take longer to let users know about RADE. I say we defer this to a different PR once there's more discussion on the project team.

barjac:
> I actually just saw an ezDV user on FreeDV Reporter right now.

Yes he just sits there 24/7 on 6m hunting dinosaurs, more in hope than expectation I guess!

> Not sure how many there are but more steps to enable legacy modes means that it'll take longer to let users know about RADE.

You would have to explain that logic, if they are having to navigate to 700D from RADEV1 they already do know about it.
It would involve one more click on a 'Mode' menu item as in wsjtx ;)

> I say we defer this to a different PR once there's more discussion on the project team.

OK

PR #1089, opened 2025-11-10
FreeDV Reporter bug fixes

tmiw:
Fixes the following FreeDV Reporter related bugs:

1. Don't return IPv6 at all if not configured (previously it was returning i.e. `::ffff:1.2.3.4` which may not be the right thing to do on some platforms).
2. Compare `socket_` in `TcpConnectionHandler` against `INVALID_SOCKET` (fixes crash on exit on Windows).
3. Fixes virtual audio device cleanup in Linux tests and ensures Windows tests actually fail if crashed.
4. Use acquire/release semantics for `TcpConnectionHandler` socket.
5. Refactor `ThreadedTimer` to not create more than one thread for all timers on the system (Linux/Windows).

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -228,10 +228,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Setup tmate session
-      uses: mxschmitt/action-tmate@v3
-      with:
-        detached: true
+    #- name: Setup tmate session
+    #  uses: mxschmitt/action-tmate@v3
+    #  with:
+    #    detached: true
 
     - name: Save correct architecture
       id: save-arch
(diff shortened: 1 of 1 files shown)


Tyrbiter:
After building and running this version I now have CPU usage of between 100 and 120% of 1 core on an 8 core i7. That's up from about 30-60% when using the previous build without the Reporter changes.

This happens before the modem is started, so I think it is due to the Reporter changes.

tmiw:
> After building and running this version I now have CPU usage of between 100 and 120% of 1 core on an 8 core i7. That's up from about 30-60% when using the previous build without the Reporter changes.
> 
> This happens before the modem is started, so I think it is due to the Reporter changes.

Known issue with `ThreadedTimer` and should be fixed in #1088.

Tyrbiter:
> > After building and running this version I now have CPU usage of between 100 and 120% of 1 core on an 8 core i7. That's up from about 30-60% when using the previous build without the Reporter changes.
> > This happens before the modem is started, so I think it is due to the Reporter changes.
> 
> Known issue with `ThreadedTimer` and should be fixed in #1088.

And indeed it is.

Issue #1090, opened 2025-11-10
Frequency widget now vanishes when reporting is off.


barjac:
The Frequency widget is now missing when reporting is off and CAT is enabled.

Anyone using the reporter tracking cannot see the current rig frequency on the screen.

Anyone using the widget to control the radio frequency can't.

Current master @ #9a01a60

tmiw:
I tested with macOS on 0bb0529 and the drop-down seems to be hidden only if both reporting and Hamlib are disabled as expected. If you try that commit and still have issues, let me know and I can look further (as it wouldn't surprise me if there was an OS/DE dependency on this bug).

barjac:
In my main system which is set up for full TX/RX running 0bb0529, if I uncheck reporting the Frequency widget vanishes. Same on the laptop running Mageia 9. (Main is Mageia 10).

tmiw:
Tested with Fedora 43 and Plasma, same results on my end. Trying Ubuntu 24.04 + GNOME now...

EDIT: no issues with Ubuntu and GNOME either. Hmm.

barjac:
I was about to start bisecting, however it seems this is not entirely new.

I just checked with an early AppImage and a new .freedv config file and am seeing the same or similar. Maybe in the past I did run with reporting on as a workaround but this is still a bug.

Use case:
I want to be able to not report and manually set a frequency in the frequency widget to be able to use the reporter filters, with or without CAT for listening only.

If I am using a remote web SDR, I certainly do not want to report and I don't need CAT.

The lack of this ability could explain the large number of stations reporting, who would otherwise not need to, if they are just tuning around listening but want to use the reporter and filters.

Why ever remove the frequency widget from the GUI?

tmiw:
I think I understand what you're trying to accomplish now. Yeah, the current implementation doesn't support that use case and you'd either need CAT control and/or reporting turned on. As a workaround, would using the "Hamlib Dummy" device help (possibly with /dev/null as the serial port)? I've never tried this so I'm not 100% sure it'll work.

If I was to make changes, would keeping the drop-down always enabled cause user confusion vs. e.g. adding a separate drop-down in the Reporter window (which would possibly be sync'd with the main window one if CAT control/reporting's turned on)?

barjac:
[...]

PR #1093, opened 2025-11-13
Always enable frequency dropdown, even if reporting and CAT are disabled.

tmiw:
Per request in #1090, the frequency drop-down in the main window is now always enabled.


barjac:
Yes! This is looking good. Thanks!
Will test on laptop tomorrow.

barjac:
Perfect!
Working as expected with no CAT or reporting and frequency widget providing correct frequ/band in reporter along with correct compass headings.
Tested in both Mageia 9 and 10.
Good to go for me :)

PR #1094, opened 2025-11-14
Reduce RADE RX losses due to resampling.

tmiw:
This PR reduces sources of resampling loss by using correct data types where appropriate (for example, `ResampleStep` calculates the sampling ratio as a `double` since that's what libsamplerate needs).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -852,6 +852,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 1. Bugfixes:
     * Update VC Redistributable and suppress reboots during its installation. (PR #1102)
+    * Reduce RADE RX losses due to resampling. (PR #1094)
 2. Enhancements:
     * FreeDV Reporter: Sort empty user messages below non-empty ones. (PR #1105)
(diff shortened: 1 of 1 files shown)


Tyrbiter:
Trying this PR in a new build including PR #1093 so I will report how it goes.

Any idea how much of a difference it should make?

tmiw:
> Trying this PR in a new build including PR #1093 so I will report how it goes.
> 
> Any idea how much of a difference it should make?

Maybe fewer audio dropouts, but honestly the code I changed didn't seem like it would have any effect. We'll see based on what GitHub says.

Tyrbiter:
Is this going to go in 2.1.0? So far I don't see anything bad, but it's not easy to quantify if it improves RADE receive on air.

tmiw:
> Is this going to go in 2.1.0? So far I don't see anything bad, but it's not easy to quantify if it improves RADE receive on air.

Based on the test runs so far, I don't think it's improving anything. I'm going to do some further investigating locally and see if I can figure out why it intermittently loses sync within the first few seconds of the test. 

(In short, probably not in 2.1.0.)

Tyrbiter:
Makes perfect sense to me. I will probably move back to the master branch and see if I find any problems.

PR #1099, opened 2025-11-17
Prepare for v2.1.0 release.

tmiw:
This PR clears the dev flag and disables release expiry in preparation for the v2.1.0 release. We also disable BWE here; once it's a bit more stable we'll reenable.

### DIFF
--- a/src/util/SocketIoClient.cpp
+++ b/src/util/SocketIoClient.cpp
@@ -49,6 +49,14 @@ SocketIoClient::~SocketIoClient()
     enableReconnect_.store(false, std::memory_order_release);
     auto fut = disconnect();
     fut.wait();
+
+    // Make absolutely sure there's nothing else in the queue before release.
+    std::shared_ptr<std::promise<void>> prom = std::make_shared<std::promise<void>>();
+    auto fut2 = prom->get_future();
+    enqueue_([&]() {
+        prom->set_value();
+    });
+    fut2.wait();
 }
 
 void SocketIoClient::setAuthDictionary(yyjson_mut_doc* authJson)


Tyrbiter:
Woo. All tests passed!

barjac:
Not sure if this is new or not but I just hit a minor bug that may be easy to fix before release?
If the VK 'time between repeats' is adjusted with the modem running and reporting on, reporting stops without warning until the modem is restarted.

tmiw:
> Not sure if this is new or not but I just hit a minor bug that may be easy to fix before release? If the VK 'time between repeats' is adjusted with the modem running and reporting on, reporting stops without warning until the modem is restarted.

I tried this several times on macOS and Fedora 43 and haven't been able to reproduce yet. How often can you trigger this on your end?

Tyrbiter:
> > Not sure if this is new or not but I just hit a minor bug that may be easy to fix before release? If the VK 'time between repeats' is adjusted with the modem running and reporting on, reporting stops without warning until the modem is restarted.
> 
> I tried this several times on macOS and Fedora 43 and haven't been able to reproduce yet. How often can you trigger this on your end?

I have just tried this out, and the only thing I noticed was that my entry in the Reporter dropped to the bottom with the default time of connection sort as soon as I clicked OK.

Ah, that was the web Reporter, the in-app Reporter does indeed lose my callsign when I change the VK setting(s). Comes back when modem is restarted.

tmiw:
I think I was able to duplicate the problem @barjac was having. Turns out it was a typo on the server side. I pushed a change there and I think it's okay now. Confirm?

Tyrbiter:
> I think I was able to duplicate the problem @barjac was having. Turns out it was a typo on the server side. I pushed a change there and I think it's okay now. Confirm?

Yes, I remain in the in-app Reporter when changing the VK settings, so it seems it was the dreaded typo at play.

tmiw:
OK, here's hoping CI works this time. (Looks like GitHub went down earlier today, among other issues.)

Issue #1104, opened 2025-11-22
Reporter Msg entries that are blank sort below used fields.


barjac:
This is becoming a real annoyance in nets where we use a number as the first character(s) of the Msg field to indicate our position in the net rotation.

Ideally we want 1 at the top and the others in numerical order below.

Many new stations are calling by during the net to listen/test etc. which is great, however every time they do the whole list gets shifted down sometimes by 2 or 3 lines meaning that actual participating stations are pushed out of sight at the bottom of the list.

It causes a real problem for some aged users with failing sight.

I asked Claude AI whether it thought a simple fix was possible for this and it offered this as the best solution:
"
wxGrid allows you to override the default sorting behavior with a custom comparison function. You can make it treat empty strings as having a very high sort value:
```
class CustomGridTable : public wxGridTableBase {
    int CompareValues(int row1, int row2, int col) override {
        wxString val1 = GetValue(row1, col);
        wxString val2 = GetValue(row2, col);
        
        // For the Message column, empty values sort last
        if (col == MSG_COLUMN) {
            bool empty1 = val1.IsEmpty();
            bool empty2 = val2.IsEmpty();
            
            if (empty1 && !empty2) return 1;   // val1 > val2
            if (!empty1 && empty2) return -1;  // val1 < val2
            if (empty1 && empty2) return 0;    // both empty
            
            // Normal comparison for non-empty values
            return val1.Cmp(val2);
        }
        
        // Default comparison for other columns
        return val1.Cmp(val2);
    }
};
```
"

tmiw:
Since the Msg field is a string and not a number, numerical sort seems inappropriate. Is the request that blank fields should sort above non-blank fields (vs. the other way around)?

barjac:
A picture to demonstrate. We sort Msg low to high moving down the list. Blank Msgs always sort to the top. We would like them to drop to the bottom.

<img width="1279" height="249" alt="Image" src="https://github.com/user-attachments/assets/bb9a8d6d-4a68-436e-b04e-6d91bc2124bb" />

barjac:
> Since the Msg field is a string and not a number, numerical sort seems inappropriate. Is the request that blank fields should sort above non-blank fields (vs. the other way around)?

Whatever sort is in use now works fine for our use case, except that blank/space sorts low  and floats to the top of our inverted [A..Z] list.

tmiw:
See PR #1105. Let me know if that works for you.

tmiw:
[...]

PR #1105, opened 2025-11-23
Sort blank user messages after non-blank messages.

tmiw:
Per #1104, this PR updates the current sorting function to sort blank Msg entries after non-blank entries.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -852,6 +852,8 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 1. Bugfixes:
     * Update VC Redistributable and suppress reboots during its installation. (PR #1102)
+2. Enhancements:
+    * FreeDV Reporter: Sort empty user messages below non-empty ones. (PR #1105)
 
 ## V2.1.0 November 2025
 


barjac:
**Perfect!** Thanks as always :)

EDIT: Just been checking further. Could this be extended to ignore leading spaces before the sort?
Stations that have leading spaces in the Msg are also appearing first in the list as well. No idea why they do that! :\ 

\OT
I notice that the 'BW Enhancement' selector in the filter dialogue is missing now in this branch. It was in 2.1.0 (maybe I built master between changes) but not for Win apparently? I did not notice any audible difference using it so did you remove it?

tmiw:
> EDIT: Just been checking further. Could this be extended to ignore leading spaces before the sort? Stations that have leading spaces in the Msg are also appearing first in the list as well. No idea why they do that! :\

Done.

> \OT I notice that the 'BW Enhancement' selector in the filter dialogue is missing now in this branch. It was in 2.1.0 (maybe I built master between changes) but not for Win apparently? I did not notice any audible difference using it so did you remove it?

It was temporarily disabled just prior to 2.1.0 as it didn't seem to make any obvious difference. Once there's more investigation, it'll likely be turned back on.

barjac:
Excellent! I can't fault it now, for me that's good to go :)

PR #1106, opened 2025-11-24
Add /dev/rfcomm* usb over bluetooth to serial port list on Linux

NespaLa:
Quick fix, might not be in the correct if / endif 
only tested on Linux.


tmiw:
Merged. Thanks!

PR #1108, opened 2025-11-28
Hide all modes except for RADE by default.

tmiw:
This PR adds a new "Enable Legacy Modes" option to Tools->Options->Modem (disabled by default). When enabled, this causes FreeDV to behave as per version <= 2.1.0, but when disabled, RADE is forced as the mode and all mode-specific settings on the main window are either disabled or hidden.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -988,7 +988,7 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent, wxID_ANY, _("FreeDV ")
     SYNC_UNK_LABEL("Sync: unk"),
     VAR_UNK_LABEL("Var: unk"),
     CLK_OFF_UNK_LABEL("ClkOff: unk"),
-    TOO_HIGH_LABEL("Too High"),
+    TOO_HIGH_LABEL("Clip"),
     MIC_SPKR_LEVEL_FORMAT_STR("%s%s"),
     DECIBEL_STR("dB"),
     CURRENT_TIME_FORMAT_STR("%s %s"),
--- a/src/topFrame.cpp
+++ b/src/topFrame.cpp
@@ -462,15 +462,15 @@ TopFrame::TopFrame(wxWindow* parent, wxWindowID id, const wxString& title, const
     //------------------------------
     wxStaticBoxSizer* levelSizer;
     wxStaticBox* levelBox = new wxStaticBox(m_panel, wxID_ANY, _("Level"), wxDefaultPosition, wxSize(100,-1));
-    levelSizer = new wxStaticBoxSizer(levelBox, wxVERTICAL);
-    
-    m_textLevel = new wxStaticText(levelBox, wxID_ANY, wxT(""), wxDefaultPosition, wxDefaultSize, wxALIGN_CENTRE);
-    m_textLevel->SetForegroundColour(wxColour(255,0,0));
-    levelSizer->Add(m_textLevel, 0, wxALIGN_CENTER_HORIZONTAL, 1);
+    levelSizer = new wxStaticBoxSizer(levelBox, wxHORIZONTAL);
 
-    m_gaugeLevel = new wxGauge(levelBox, wxID_ANY, 100, wxDefaultPosition, wxSize(135,15), wxGA_SMOOTH);
+    m_gaugeLevel = new wxGauge(levelBox, wxID_ANY, 100, wxDefaultPosition, wxSize(100,15), wxGA_SMOOTH);
     m_gaugeLevel->SetToolTip(_("Peak of From Radio in Rx, or peak of From Mic in Tx mode.  If Red you should reduce your levels"));
-    levelSizer->Add(m_gaugeLevel, 1, wxALIGN_CENTER_HORIZONTAL|wxALL, 10);
+    levelSizer->Add(m_gaugeLevel, 1, wxALIGN_CENTER_VERTICAL|wxALL, 10);
+    
+    m_textLevel = new wxStaticText(levelBox, wxID_ANY, wxT(""), wxDefaultPosition, wxSize(35,-1), wxALIGN_CENTRE);
+    m_textLevel->SetForegroundColour(wxColour(255,0,0));
+    levelSizer->Add(m_textLevel, 0, wxALIGN_CENTER_VERTICAL, 1);
 
     leftSizer->Add(levelSizer, 0, wxALL|wxEXPAND, 2);
     


barjac:
Could removal of the RX mode column in the reporter be included in this?

Tyrbiter:
> Could removal of the RX mode column in the reporter be included in this?

A fair point, it's taking up space to no end.

Merge master into this branch again?

tmiw:
> Could removal of the RX mode column in the reporter be included in this?

Wouldn't we want to know if anyone is still using a legacy mode so we know to turn the option on? Removing that column could make that difficult.

> Merge master into this branch again?

Done.

Tyrbiter:
> > Could removal of the RX mode column in the reporter be included in this?
> 
> Wouldn't we want to know if anyone is still using a legacy mode so we know to turn the option on? Removing that column could make that difficult.

Would this not show in the Tx mode column?

tmiw:
> Would this not show in the Tx mode column?

For the legacy modes, there's still the multi-receive capability (the one where you can dynamically switch modes instead of having to push Stop and then Start again). That means the TX mode doesn't necessarily correlate with the RX mode.

Tyrbiter:
> > Would this not show in the Tx mode column?
> 
> For the legacy modes, there's still the multi-receive capability (the one where you can dynamically switch modes instead of having to push Stop and then Start again). That means the TX mode doesn't necessarily correlate with the RX mode.

It won't but it will show as not being RADEV1 thus giving a cue to enable the legacy modes if you want to.

tmiw:
OK, I removed and hid some additional controls per email discussion (as well as hid the RX Mode column). How's this now?

barjac:
Looking good - not actually used it yet though.
One point I notice is that with the Frequency widget below the Control section it gets forced into a new wasted column on its own prematurely when the window height is reduced.
If it was placed below the Mic/Spkr Level widget then that would not happen.
This would allow more main window height compression without wasted space crushing the width of the 'scope.
Similarly on the left if Record was placed under Level, and why does the Level widget have so much wasted space over the meter? There is much less space wasted in the same region of SNR widget.

Screenshot_20251201_121800

tmiw:
> One point I notice is that with the Frequency widget below the Control section it gets forced into a new wasted column on its own prematurely when the window height is reduced. If it was placed below the Mic/Spkr Level widget then that would not happen. This would allow more main window height compression without wasted space crushing the width of the 'scope. 

Done.

> Similarly on the left if Record was placed under Level, and why does the Level widget have so much wasted space over the meter? There is much less space wasted in the same region of SNR widget.

There's a hidden "Too High" indicator that's supposed to light up if the audio level is too high. I'm open to suggestions on how to better indicate this.

Tyrbiter:
Is the python-3.14 change merged into this branch?

barjac:
> There's a hidden "Too High" indicator that's supposed to light up if the audio level is too high. I'm open to suggestions on how to better indicate this.

It still does that unless something happened recently. I think it is OK as it is, although some people see the odd flash and start turning down their audio level into FreeDV so it is too low for the 'AGC' to compensate.

EDIT: I only just realized why you mentioned that! Of course that is where the 'too high!' appears. Maybe the top 10% could make the level bar change colour to red as an alternative?

barjac:
> Is the python-3.14 change merged into this branch?

So what happens if we don't have python-3.14 yet? I'm not even sure whether Mageia will have it in the next release. I assume that the python packages installed in the system need to match whatever pip pulls into the venv? Or does the available system version control the versions of what else is pulled in? I am not well versed in python.

tmiw:
> Is the python-3.14 change merged into this branch?

Yes. 

> So what happens if we don't have python-3.14 yet?

The change primarily affects Windows and macOS; Linux will just use whatever Python you have on your system.

barjac:
See edit here https://github.com/drowe67/freedv-gui/pull/1108#issuecomment-3599529802

tmiw:
Not sure we can change the color of the level indicator in a cross-platform fashion, unfortunately. Maybe we can move the label to the right of it instead of above it? If so, what shorter text can we use instead of "Too High" that will convey the same meaning?

barjac:
> Is the python-3.14 change merged into this branch?

So what happens if we don't have python-3.14 yet? I'm not even sure whether Mageia will have it in the next release. I assume that the python packages installed in the system need to match whatever pip pulls into the venv? Or does the available system version control the versions of what else is pulled in? I am not well versed in python.

> Not sure we can change the color of the level indicator in a cross-platform fashion, unfortunately. Maybe we can move the label to the right of it instead of above it? If so, what shorter text can we use instead of "Too High" that will convey the same meaning?

Is there something like an LED symbol that could show green amber red or just grey and red when too high?

tmiw:
> > Is the python-3.14 change merged into this branch?
> 
> So what happens if we don't have python-3.14 yet? I'm not even sure whether Mageia will have it in the next release. I assume that the python packages installed in the system need to match whatever pip pulls into the venv? Or does the available system version control the versions of what else is pulled in? I am not well versed in python.

As mentioned above, there should be no changes on Linux when building (either manually or via the build script). It'll just use whatever version of Python is in the venv you set up prior to running `build_linux.sh`. For example, if you create the venv like so:

```
$ python3.11 -m venv rade-venv
$ . ./rade-venv/bin/activate
$ pip3 install torch ...
$ ./build_linux.sh
```

then FreeDV will be built using Python 3.11.

Note: AppImages are a bit different because we are hardcoding references to 3.14 right now. The script that generates those isn't really intended for use outside of the GitHub Actions environment, though.

> > Not sure we can change the color of the level indicator in a cross-platform fashion, unfortunately. Maybe we can move the label to the right of it instead of above it? If so, what shorter text can we use instead of "Too High" that will convey the same meaning?
> 
> Is there something like an LED symbol that could show green amber red or just grey and red when too high?

Maybe. I'll think about that some more.

tmiw:
Question: would shortening "Too High" to "High" and just putting that to the right of the level meter be okay, or would that still take too much space?

Tyrbiter:
> Question: would shortening "Too High" to "High" and just putting that to the right of the level meter be okay, or would that still take too much space?

It sounds OK to me, as long as there is something obvious in the user manual about it.

barjac:
No I think that would be worse. Better left as it is so it matches the SNR widget width.

Tyrbiter:
Perhaps the width of the level indicator part could be narrowed so that the total width is the same.

tmiw:
> Perhaps the width of the level indicator part could be narrowed so that the total width is the same.

Can you clarify? Do you mean reduce the amount of whitespace on the left and right of the actual meter control, or?

Tyrbiter:
> > Perhaps the width of the level indicator part could be narrowed so that the total width is the same.
> 
> Can you clarify? Do you mean reduce the amount of whitespace on the left and right of the actual meter control, or?

I mean to move the 'Too High' text to the right of the level meter bar, with the bar reduced in width and offset to the left and 'High' moving right, level with the right hand end of it. That should allow the overall height of the box to be reduced as @barjac is requesting.

tmiw:
So there's something called "indeterminate mode" for the wxWidgets control being used for the level. I checked in a change to use that instead of the text label. How does that work for you guys?

barjac:
> So there's something called "indeterminate mode" for the wxWidgets control being used for the level. I checked in a change to use that instead of the text label. How does that work for you guys?

Rather baffling! It reaches around 80% then the illuminated part on the left turns off and switches to the right to occupy the space that was not illuminated. Not usable as it is - it just looks broken. :(

tmiw:
> > So there's something called "indeterminate mode" for the wxWidgets control being used for the level. I checked in a change to use that instead of the text label. How does that work for you guys?
> 
> Rather baffling! It reaches around 80% then the illuminated part on the left turns off and switches to the right to occupy the space that was not illuminated. Not usable as it is - it just looks broken. :(

Fair enough, I reverted the change. Should we just defer updating the Level box for a later PR, or keep this open until we can find a better solution that makes the box shorter?

Tyrbiter:
Is it possible to change the colour of the bar if the level is too high? I suppose the problem then is accessibility for people with colour blindness.

tmiw:
> Is it possible to change the colour of the bar if the level is too high? I suppose the problem then is accessibility for people with colour blindness.

I don't think we can, unfortunately. Going to play around with a shorter label to the right of the bar.

Tyrbiter:
That's actually looking good to me, should please @barjac

barjac:
Yes that looks OK, however it seems that GTK bar graphs can be coloured like a VU meter according to Claud AI.
"
Yes, both GTK and wxWidgets have widgets that can be used to create VU meter-style level indicators with color-coded segments.
GTK Options

GtkLevelBar is exactly what you're looking for. It's designed for this purpose and supports discrete levels with different colors:
c

```
GtkWidget *level_bar = gtk_level_bar_new();
gtk_level_bar_set_mode(GTK_LEVEL_BAR_MODE_CONTINUOUS, level_bar);

// Define thresholds for color changes
gtk_level_bar_add_offset_value(level_bar, "low", 0.0);
gtk_level_bar_add_offset_value(level_bar, "high", 0.75);
gtk_level_bar_add_offset_value(level_bar, "full", 0.90);

// Set the current level (0.0 to 1.0)
gtk_level_bar_set_value(level_bar, 0.65);
```

You can style it with CSS to get green/amber/red colors:
css

```
levelbar block.low {
    background-color: #5cb85c; /* green */
}
levelbar block.high {
    background-color: #f0ad4e; /* amber */
}
levelbar block.full {
    background-color: #d9534f; /* red */
}

```
wxWidgets Options

wxGauge is the standard progress indicator, but it doesn't natively support color zones. You'd typically:

    Use wxGauge and manually change its color based on the current value
    Create a custom widget by subclassing wxPanel and drawing the level bar yourself with different colored rectangles
    Use a third-party widget library

The custom drawing approach gives you the most control for a professional VU meter look with smooth color transitions at your specified thresholds.
"

tmiw:
> Yes that looks OK, however it seems that GTK bar graphs can be coloured like a VU meter according to Claud AI.

Maybe, but we would need a cross-platform solution for this to be viable (or at least something we can update the existing wxWidgets patch we have to support). Unfortunately I don't think this is possible on macOS at least (see [here](https://developer.apple.com/documentation/appkit/nsprogressindicator/controltint?language=objc) which says that the property we'd need to update is simply ignored by the OS); not sure about Windows.

Anyway, I'll go ahead and merge for now. We can do a followup PR if we change our minds on the current solution.

barjac:
What happened? I don't see this in master.

Tyrbiter:
It's in 2.2-dev, but not in master. Not sure why.

tmiw:
> It's in 2.2-dev, but not in master. Not sure why.

Trying to keep feature changes separated from bugfixes in case we need to release a 2.1.1 prior to 2.2.0.

PR #1113, opened 2025-11-29
Reenable BW expander.

tmiw:
Reenables BW expander feature that was disabled prior to 2.1.0 release.


Tyrbiter:
Is there a way to see the BW increase with this enabled? It may be constrained by other things in the audio sub-systems, and then there's my old ears in the loop ;-)

tmiw:
> Is there a way to see the BW increase with this enabled? It may be constrained by other things in the audio sub-systems, and then there's my old ears in the loop ;-)

I've been setting the speaker device to a virtual audio device and recoding that device with Audacity. It's definitely doing something but it's really hard to tell by ear.

Tyrbiter:
> > Is there a way to see the BW increase with this enabled? It may be constrained by other things in the audio sub-systems, and then there's my old ears in the loop ;-)
> 
> I've been setting the speaker device to a virtual audio device and recoding that device with Audacity. It's definitely doing something but it's really hard to tell by ear.

That's the sort of thing I was worried about, presumably Audacity should make it possible to see the extra bandwidth in a frequency response plot.

tmiw:
> That's the sort of thing I was worried about, presumably Audacity should make it possible to see the extra bandwidth in a frequency response plot.

Yep, Audacity does have a frequency plot (Analyze->Plot Spectrum).

PR #1123, opened 2025-12-07
Linux: Force CPU affinity for TX/RX threads.

tmiw:
Experimental PR based on observations discovered during `perf` context switch analysis, where it was taking a long time to migrate threads between CPU cores (to the point where audio drops were occurring). The changes pin the TX and RX threads to the last two cores on a user's system to ensure they don't migrate. 

Note: migration of other threads/processes may still cause audio drops, but can't really do anything about that.


tmiw:
I'm not convinced this is doing anything at all for the test pass rate on Linux here on GitHub. I'm going to close for now.

BTW `perf` is showing pretty high CPU usage for `PlotScalar` when transmitting, which could explain the behavior on my Fedora 43 VM (where `rade_loss` fails most of the time because there are dropouts in the TX audio). Since GH effectively uses X11 (`Xvfb` specifically) this could also be adding additional overhead similar to Xwayland from previous optimization PRs.

tmiw:
Reopening due to #1210 and will bring up to v2.2.0 baseline.

tmiw:
Running tests on Fedora 43 with the following command line (and `CMAKE_BUILD_TYPE` set to Release):

```
$ for i in $(seq 0 19); do PYTHONPATH=$(pwd)/rade_src ctest -R rade_loss 2>&1 | grep "tests failed"; done
```

* master (https://github.com/drowe67/freedv-gui/commit/f4a35367d07d1efbbae19dd9410a320c1c8eb14d): 12/20 passed (60%)
* This PR (https://github.com/drowe67/freedv-gui/pull/1123/commits/7eb0acf16d10d832ffb80b603f55d3a88703051f): 17/20 passed (85%)

Need to run tests with just the changes in https://github.com/drowe67/freedv-gui/pull/1123/commits/e37faf06689cd934750b909f82b6af5b6e6fa5ba to see if we still need the CPU affinity logic or if the wakeup time adjustments are sufficient to improve dropouts. 

Note: this PR is not a full solution since as shown above, there are still dropouts.

PR #1129, opened 2025-12-12
Initial logging support using WSJT-X network protocol.

### LINKED ISSUE #1069: [Feature Request] Add log button to reporter.
### Contact Details

G4MKT G6WPJ G6YRK

### Feature for this project

This feature would add a 'Log' button to the reporter to work just like the similar feature in wsjtx.

If a station line is highlighted (clicked) in the reporter then a 'Log' button (next to the Request QSY button?) would become active.

Clicking the button would send callsign, frequency, band, mode (DIGITALVOICE) and date/time over UDP to a logging program.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

Testing as always.

### Support assistance

Will maintain Mageia packaging when appropriate.

tmiw:
This PR resolves #1069 by implementing logging support using the [WSJT-X network protocol](https://sourceforge.net/p/wsjt/wsjtx/ci/master/tree/Network/NetworkMessage.hpp), supported by KLog and possibly other logging software.

TODO:

- [x] Test with KLog to make sure networking/logging works
- [x] Implement configuration of IP/port for logging

### DIFF
--- a/src/logging/WSJTXNetworkLogger.cpp
+++ b/src/logging/WSJTXNetworkLogger.cpp
@@ -192,7 +192,7 @@ WSJTXNetworkLogger::WSJTXNetworkLogger(std::string hostname, int port)
     , reportHostname_(std::move(hostname))
     , reportPort_(port)
 {
-    open("", 0);
+    open("", 0, reportHostname_.c_str(), reportPort_);
     heartbeatTimer_.start();
     sendHeartbeat_();
 }
--- a/src/util/UdpHandler.cpp
+++ b/src/util/UdpHandler.cpp
@@ -79,7 +79,7 @@ UdpHandler::~UdpHandler()
 #endif // defined(WIN32)
 }
 
-std::future<void> UdpHandler::open(const char* host, int port)
+std::future<void> UdpHandler::open(const char* host, int port, const char* sendIp, int sendPort)
 {
     host_ = host;
     port_ = port;
@@ -87,8 +87,8 @@ std::future<void> UdpHandler::open(const char* host, int port)
     std::shared_ptr<std::promise<void>> prom = std::make_shared<std::promise<void> >();
     auto fut = prom->get_future();
     
-    enqueue_([&, prom]() {
-        openImpl_();
+    enqueue_([&, prom, sendIp, sendPort]() {
+        openImpl_(sendIp, sendPort);
         prom->set_value();
     });
     return fut;
@@ -118,67 +118,41 @@ std::future<void> UdpHandler::send(const char* host, int port, const char* buf, 
     return fut;
 }
 
-void UdpHandler::openImpl_()
+void UdpHandler::openImpl_(const char* sendIp, int sendPort)
 {
     std::stringstream portStream;
     portStream << port_;
     
     log_info("Opening UDP socket (host %s port %d)", host_.c_str(), port_);
     
-    // If empty hostname, use INADDR_ANY
+    // If empty hostname, see if we can infer the socket type from the provided
+    // example send port. If not, default to INADDR_ANY.
     if (host_ == "")
     {
+        auto addressType = AF_INET;
+        if (sendIp != nullptr)
+        {
+            struct addrinfo* result = resolveIpAddress_(sendIp, sendPort);
+            if (result != nullptr)
+            {
+                addressType = result->ai_family;
+                freeaddrinfo(result);
+            }
+        }
+
 #if defined(WIN32)
-        socket_ = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+        socket_ = socket(addressType, SOCK_DGRAM, IPPROTO_UDP);
         if (socket_ == INVALID_SOCKET)
         {
             log_warn("cannot open socket (err=%d)", WSAGetLastError());
         }
 #else
-        socket_ = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+        socket_ = socket(addressType, SOCK_DGRAM, IPPROTO_UDP);
         if(socket_ < 0)
         {
             log_warn("cannot open socket (err=%d)", errno);
         }
 #endif // defined(WIN32)
-
-        // XXX - macOS isn't able to set V6ONLY. Forcing IPv4 for now.
-#if 0
-        // Make sure we can receive IPv4 too
-        int opt_false = 0;
-#if defined(WIN32)
-        auto rv = setsockopt(socket_, IPPROTO_IPV6, IPV6_V6ONLY, (char*)&opt_false, sizeof(opt_false));
-        if (rv != 0)
-        {
-            log_warn("cannot set socket options (err=%d)", WSAGetLastError());
-        }
-#else
-        auto rv = setsockopt(socket_, IPPROTO_IPV6, IPV6_V6ONLY, &opt_false, sizeof(opt_false));
-        if (rv != 0)
-        {
-            log_warn("cannot set socket options (err=%d)", errno);
-        }
-#endif // defined(WIN32)
-        
-        if (port_ > 0)
-        {
-            struct sockaddr_in6 sin6;
-            sin6.sin6_family = AF_INET6;
-            memcpy(sin6.sin6_addr.s6_addr, in6addr_any.s6_addr, 16);
-            sin6.sin6_port = htons(port_);
-            sin6.sin6_flowinfo = 0;
-            sin6.sin6_scope_id = 0;
-            auto rv = ::bind(socket_, (struct sockaddr*)&sin6, sizeof(sin6));
-            if (rv != 0)
-            {
-    #if defined(WIN32)
-                log_warn("cannot bind socket (err=%d)", WSAGetLastError());
-    #else
-                log_warn("cannot bind socket (err=%d)", errno);
-    #endif // defined(WIN32)
-            }
-        }
-#endif
     }
     else
     {
--- a/src/util/UdpHandler.h
+++ b/src/util/UdpHandler.h
@@ -37,7 +37,8 @@ public:
     virtual ~UdpHandler();
     
     // Note: host is an IP here, not hostname.
-    std::future<void> open(const char* host, int port);
+    std::future<void> open() { return open("", 0); }
+    std::future<void> open(const char* host, int port, const char* sendIp = nullptr, int sendPort = 0);
     std::future<void> close();
     std::future<void> send(const char* host, int port, const char* buf, int length);
     
@@ -55,7 +56,7 @@ private:
     std::atomic<int> socket_;
 #endif // defined(WIN32)
 
-    void openImpl_();
+    void openImpl_(const char* sendIp, int sendPort);
     void closeImpl_();
     void sendImpl_(const char* host, int port, const char* buf, int length);
     void receiveImpl_();


tmiw:
Example popup from KLog when it gets a packet from FreeDV:

image

I'm not sure where the 59 is coming from, but 33 *should* be the SNR.

tmiw:
Proposed configuration options:

image

barjac:
I just built it and was surprised to see that the 'log qso' button is in the main window.
I expected that the process would be to select a station that you are in qso with in the reporter and click a 'LOG' button in the reporter to add that station with its current info to klog.
Only one window can be in focus at once so splitting the operation between them would be rather clunky.

I'm not sure that a reliable SNR would ever really be available for logging unless the button was clicked during MY TX so real time SNR from the remote station would be available.

Also what form of reporting is really relevant for freedv? I tend to favour 587 where the 7 would be audio quality similar to tone quality on CW and entered manually.

I decided that here in the PR, rather than the bug tracker is probably the best place for ongoing comments so apologies for some duplication.

The config looks fine and the reporting tab seems the best place.

OK now installed in my main system:

It is logging stations directly into klog but so far I cannot see any correlation between the station selected in the reporter and the station being logged. The logged stations seem random or possibly stations transmitting? This was 40m with multiple stations in TX simultaneously.

How is this intended to work?

I later tested in an actual live qso with tracking on for 5.3685MHz to avoid multiple transmissions and nothing was logged at all irrespective of who was transmitting or whether the remote station's line in the reporter was 'selected'. More testing later.

tmiw:
> How is this intended to work?

I wanted to avoid dependence on Internet or FreeDV Reporter availability, so it's grabbing the most recent callsign/frequency/etc. from the drop down list in the main window. If it's not able to decode anything, there won't be a callsign to log. I'm open to suggestions, though.

barjac:
OK good results today after new build and reboot and with two proper QSOs on 40m.
Fields filled out in klog were as follows:
Callsign OK
Time: OK (In UTC/GMT which is local just now)
Mode: DIGITALVOICE OK
Frequency: 7.000000 which should have been 7.177 [1]
Split: Yes (checked in klog) so wrong. [1]
RX Rprt: 30? strange number (from memory ;)
TX Rprt: 59 (filled in by default by klog if not entered)

Maybe some fields could be filled from the reporter as well, if available: 
1. The exact frequency (if it does not already [1]) from hamlib etc.
2. The locator
3. Maybe skip the SNR as a report for now and let the logger use a default 59 or actually send a default 599 that can be manually edited later [2]

Could a pop up dialogue appear in FreeDV to show what will be sent to klog with the option to add more detail, like name, city and comments along with the other data picked up from FreeDV and the reporter that could be quickly edited before hitting OK to send it on to the logging program, as in wsjtx? I don't see anything similar to your screen-shot above.

[1] There is a bug in klog that is not correctly detecting the band such that it always defaults to 40m so this may be causing the frequency/split issue.

[2] I think that a 599 style subjective report is more suitable than trying to extract a continually changing 'SNR' from the system, at least for now.

barjac:
Running this branch it's fine with tracking enabled in the reporter but shortly (a few seconds) after disabling tracking it segfaults. I will try to get a backtrace if it will reproduce under gdb.

barjac:
Tried to log Matt tonight without success.
His call was the only one in the list below the 'scope'.
The Log button was not greyed.
klog was running.
Pressing the button gave a warning in the FreeDV terminal output:
.....UdpHandler.cpp:289 : Cannot resolve : (null) (err=-2)
I could not see any packets arriving at lo udp/2237

tmiw:
> Tried to log Matt tonight without success. His call was the only one in the list below the 'scope'. The Log button was not greyed. klog was running. Pressing the button gave a warning in the FreeDV terminal output: .....UdpHandler.cpp:289 : Cannot resolve : (null) (err=-2) I could not see any packets arriving at lo udp/2237

Can you check Tools->Options->Reporting again and make sure nothing got corrupted in there?

tmiw:
OK, I think I figured out why the config options for logging seem to intermittently be getting overwritten. You may need to reset hostname/port but other than that, things should work from here on out.

BTW there's also a dialog that appears when you click Log QSO to confirm info before it gets sent to KLog/whatever logger you're using. I was able to see the frequency problem (it was on this end, not KLog) and fix that too, along with adding the ability to grab the other side's locator/grid from FreeDV Reporter. Let me know how this goes.

barjac:
Looking good now!
The config corruption was causing confusion previously - now seems reliable.
A few observations:
1. At present only the top station in the drop down list can be logged. Can this be extended to work by selecting any station in the list? In a group situation it's more convenient to log stations while others are talking and the last to stop TX is not always the one you want to log.
2. The frequency display in the pop-up dialogue is not displayed in the local user's national format in the confirmation dialogue, although it is in klog.
3. Only the TX frequency is entered in klog, the 'Split' frequency checkbox becomes checked and the RX frequency is 0.000000. Maybe both need to be sent to avoid this?
4. The date entered is wrong - its a day previous and confusing 12/24 hour clock formats. Here its 11:47 in the morning and it logged 23:47 the previous day :) Maybe the date/time to be logged could be added to the confirmation dialogue?

Enough to keep you busy for now ;)

tmiw:
> 1. At present only the top station in the drop down list can be logged. Can this be extended to work by selecting any station in the list? In a group situation it's more convenient to log stations while others are talking and the last to stop TX is not always the one you want to log.
> 2. The frequency display in the pop-up dialogue is not displayed in the local user's national format in the confirmation dialogue, although it is in klog.

These are done as of the latest.

> 3. Only the TX frequency is entered in klog, the 'Split' frequency checkbox becomes checked and the RX frequency is 0.000000. Maybe both need to be sent to avoid this?

I tried sending the "status" message from https://sourceforge.net/p/wsjt/wsjtx/ci/master/tree/Network/NetworkMessage.hpp#l76 (which describes the WSJT-X network protocol) but KLog still isn't picking up the RX frequency. I'll need to investigate this more.

> 4. The date entered is wrong - its a day previous and confusing 12/24 hour clock formats. Here its 11:47 in the morning and it logged 23:47 the previous day :) Maybe the date/time to be logged could be added to the confirmation dialogue?

Didn't get a chance to investigate this either but I'll update the PR when I have a fix.

tmiw:
Hmm, I think it might be a KLog bug. I added some logging to get what the Julian date being sent is:

```
12:57:56 [79] INFO /Users/mooneer/devel/freedv-gui/src/logging/WSJTXNetworkLogger.cpp:181: Julian day count: 2461028.373568
```

and according to [this calculator](https://apps.aavso.org/v2/tools/julian-date-converter/) that matches up with the time I confirmed logging in FreeDV (2025-12-18 20:59:48 UTC). KLog, however, gave me 085948 for the time in its own confirmation box, which makes me think it's subtracting 12 hours somewhere in the logging process. I'm going to see if I can find another WSJT-X capable logging tool for macOS and confirm.

EDIT: I'm seeing the same thing in MacLoggerDX so I just checked in a change to bump up the time by 12 hours. As it's supposed to be UTC time anyway I think this will be okay, but let me know how it goes.

tmiw:
Re: split--I think that might very well be a KLog bug. MacLoggerDX is only showing one frequency with the latest changes and the ADIF log from KLog as well.

barjac:
OK I have been testing the latest (4d000) tonight with two other stations.

Time/date is now correct. We think Time/Date should be shown in the confirmation dialogue.

Logging any QSO selected in the list is working. Locator is missing if station has left which is OK.

I asked Jaime (klog) to comment on the RX freq issue.

We think a default 59/59 report should be added to the confirmation dialogue which _could_ then be edited before sending. Also Name, QTH and a short comment field too.

With those additions I think it would be super! (assuming the RX freq can be resolved).

/OT
With VOX TX, if there is a short break in transmission (hang time times out) the 'from mic' display switches to 'from decoder' and never returns to 'from mic' again :\ I suspect that it happens if the PTT drops and returns before EOO. Not sure.

robertprew54:
Logging QSO data from 4d000 to RumLogNG on my Apple Mac Mini M2 (Apple Silicon) running Tahoe 26.2 seems to be working OK for me.

Bob G5BOB

tmiw:
> We think a default 59/59 report should be added to the confirmation dialogue which _could_ then be edited before sending. Also Name, QTH and a short comment field too.

Done, except for QTH (which isn't part of the WSJT-X network protocol).

> /OT With VOX TX, if there is a short break in transmission (hang time times out) the 'from mic' display switches to 'from decoder' and never returns to 'from mic' again :\ I suspect that it happens if the PTT drops and returns before EOO. Not sure.

Create a new issue for this so we don't lose track?

> Logging QSO data from 4d000 to RumLogNG on my Apple Mac Mini M2 (Apple Silicon) running Tahoe 26.2 seems to be working OK for me.

Excellent! What about the latest? (ccfb675)

robertprew54:
> Excellent! What about the latest? ([ccfb675](https://github.com/drowe67/freedv-gui/commit/ccfb67535202b4e320c6501d39ce440ecf43ff4e))

Built and installed ccfb675 last night. Just tested in a QSO and it seems to be logging all data correctly (Datre, Time, Frequency etc) to RumLogNG. I will try logging to "QLog" for my next QSO on FreeDV. 

Update : builds ccfb675 and 496850 are correctly logging all QSO data to QLog as well as RumLogNG (Both are macOS apps). No issues with time and date.

barjac:
Using #4968
At first I thought it had stopped working altogether but in klog the listing was sorting on date and ...

The Date/Time is showing in the confirmation as only time which is the correct time.
The Date/Time arriving in klog is 1752-09-14 - don't think we had FreeDV then! :)
In the klog QSO listing table it shows as '-4990-05-28 16:25:53' (hence sorted to bottom).

The above were all around 11:30 AM today Saturday, 20 December 2025
My local system date/time is:
2025-12-20 11:48:27 Greenwich Mean Time

Edit:
baz@localhost:~ (master)$ date
Sat 20 Dec 12:42:19 GMT 2025


The Name and Comment are being logged OK.

Getting close! :)

tmiw:
> Using #4968 At first I thought it had stopped working altogether but in klog the listing was sorting on date and ...
> 
> The Date/Time is showing in the confirmation as only time which is the correct time. The Date/Time arriving in klog is 1752-09-14 - don't think we had FreeDV then! :) In the klog QSO listing table it shows as '-4990-05-28 16:25:53' (hence sorted to bottom).
> 
> The above were all around 11:30 AM today Saturday, 20 December 2025 My local system date/time is: 2025-12-20 11:48:27 Greenwich Mean Time
> 
> Edit: baz@localhost:~ (master)$ date Sat 20 Dec 12:42:19 GMT 2025
> 
> The Name and Comment are being logged OK.
> 
> Getting close! :)

Hmm, I'm not noticing this on macOS. Just logged the below:

image

I wonder if it's a Linux thing? I'll see if I can get KLog installed on Fedora and confirm there.

tmiw:
OK, try now, @barjac? 757ef5d

barjac:
> OK, try now, @barjac? [757ef5d](https://github.com/drowe67/freedv-gui/commit/757ef5d50149b775426cbea87fe16074fe654480)

Dates now fixed.
Split issue now fixed! Both TX and RX fields are correctly filled.

Perfect!

Thanks Mooneer, many stations have been commenting about how useful this will be.
Super job!!

Edit: The tooltip in Options for the logging IP and port is there. Might it be worth pre-setting 127.0.0.1 and 2237 as this will be correct for most installations, rather than leaving them blank? This is what klog does.

tmiw:
> Split issue now fixed! 

For me double clicking on a log entry added from FreeDV in KLog still shows 0 MHz for the RX frequency. I could be running a different version of KLog though.

> Edit: The tooltip in Options for the logging IP and port is there. Might it be worth pre-setting 127.0.0.1 and 2237 as this will be correct for most installations, rather than leaving them blank? This is what klog does.

We should already be defaulting to those:

```c++
    , udpReportingEnabled("/Reporting/UDP/Enable", false)
    , udpReportingHostname("/Reporting/UDP/Hostname", _("127.0.0.1"))
    , udpReportingPort("/Reporting/UDP/Port", 2237)
```

In fact, I deleted that config from my local config and confirmed that it gets populated again:

image

Anyway, I think we can go ahead and merge. If any issues come up we can do a followup PR to address them.

g4dya:
FWIW I've now got this working with CQRlog.

Listening to port 2237 can be enabled from the CQRlog main window using File->Remote mode for wsjt (Ctrl+J)

One small observation: FREEDV has been a valid sub-mode of DIGITALVOICE in the ADIF spec since version 3.1.4

https://adif.org/

tmiw:
> One small observation: FREEDV has been a valid sub-mode of DIGITALVOICE in the ADIF spec since version 3.1.4

Yeah, we're sending FREEDV as part of the Status message but it doesn't look like logging tools are picking that up. From what I can tell they mostly just listen for the actual "log" message, and that only allows for the mode (DIGITALVOICE).

PR #1132, opened 2025-12-14
Various main window and FreeDV Reporter UI improvements

tmiw:
This PR contains the following UI improvements:

* FreeDV Reporter window:
    * Allow columns to be reordered and/or made invisible based on user preference.
    * Add tooltip to describe Track feature.
* Main window:
    * Hide ReSync button if legacy modes are disabled (missed from previous PR).
    * Move Assistance button from lower left of main window to Help menu (and update URL based on website redesign).

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -14,6 +14,10 @@ env:
   BUILD_TYPE: RelWithDebInfo
   TORCH_RELEASE: 2.9.1
 
+  # Needed to suppress possible cause of test errors.
+  # See https://github.com/alexmurray/emacs-snap/issues/60
+  NO_AT_BRIDGE: 1
+
 jobs:
   lint:
     runs-on: ubuntu-24.04


tmiw:
@peterbmarks is this what you were thinking of from the other day?

Screenshot 2025-12-14 at 12 55 43 AM

Screenshot 2025-12-14 at 12 57 01 AM

peterbmarks:
Yes, both good.
A problem with combining locator and distance like that is that you can't sort by distance (which I do sometimes).
I would personally (as a net controller) like the order in FreeDV Reporter in the app to be:
Call, Locator, distance (heading), MHz, Status, Msg, RxCall, SNR, Last update, last TX, Mode.

As for the second shot. Yes, on macOS there's a help menu and our items should be there. It's the platform convention. On Windows, they also have a Help menu typically. Linux is anyone's guess.

tmiw:
> Yes, both good. A problem with combining locator and distance like that is that you can't sort by distance (which I do sometimes). I would personally (as a net controller) like the order in FreeDV Reporter in the app to be: Call, Locator, distance (heading), MHz, Status, Msg, RxCall, SNR, Last update, last TX, Mode.
> 
> As for the second shot. Yes, on macOS there's a help menu and our items should be there. It's the platform convention. On Windows, they also have a Help menu typically. Linux is anyone's guess.

Speaking of sort order, it's actually by distance when you sort the Locator column now. Hopefully that's not too confusing?

barjac:
I thought we optimised this a couple of years ago. It ain't broke.
I would rather see Locator and Distance as they are now. Mashing them together makes a dogs dinner of the justification and makes reading difficult.
If the object is to save horizontal space it really is not worth it.
To save width please just use mi or km for the distance heading and don't repeat km or mi over and over.
I also think that when direction is in cardinal mode it should be left justified to follow on cleanly after the distance.

tmiw:
> I thought we optimised this a couple of years ago. It ain't broke. I would rather see Locator and Distance as they are now. Mashing them together makes a dogs dinner of the justification and makes reading difficult. If the object is to save horizontal space it really is not worth it. To save width please just use mi or km for the distance heading and don't repeat km or mi over and over. I also think that when direction is in cardinal mode it should be left justified to follow on cleanly after the distance.

Should the Version column stay too, or is that okay to move to a tooltip under Status?

peterbmarks:
I think it is "broke" in that the window is way too wide and, speaking as a net controller, the order is not what I need. We discussed this in the last meeting which I think Barry unfortunately missed.
Yes, version is interesting but could be in a tool tip in my opinion.

tmiw:
> I think it is "broke" in that the window is way too wide and, speaking as a net controller, the order is not what I need. We discussed this in the last meeting which I think Barry unfortunately missed. Yes, version is interesting but could be in a tool tip in my opinion.

Barry isn't part of the PLT BTW.

How about this?

1. Split out Locator and Distance again.
2. Hide Mode entirely unless legacy modes are enabled. As part of this, auto-filter out anyone not using RADEV1 to avoid confusion.
3. Maybe have a "Show Additional Info" checkbox/menu option somewhere that hides Last Update and Last TX unless enabled? (Combined with moving Last TX and Last Update so they're next to each other.)

g6wpj:
Hi I run the UK post RSGB news net so will add my thoughts to this item. 

I would prefer distance and locator to be split. If you must combine items then bearing and distance would seem more logical. 

Agree that version could become a tooltip to save horizontial space. 

Also agree that if legacy modes are not enabled the mode column could be removed. In the UK we add a number as the first char in the user msg based on our position in the net order. When you join the net just use the next avaialble number. We can then run the net without a controller because every station can see at a glance who is next in the running order and speak when it gets to them. It makes for a more natural multiway conversation. Also avoids having to do the 'break-break' thing or keep inviting new joiners. 'If you wanna join grabe the next number'

Thanks and 73

barjac:
> > I think it is "broke" in that the window is way too wide and, speaking as a net controller, the order is not what I need.

Maybe keep the Last update and last TX but drop the '2025' from both?

> How about this?
> 
>     1. Split out Locator and Distance again.
 
Yes please with km/mi column title

>     2. Hide Mode entirely unless legacy modes are enabled. As part of this, auto-filter out anyone not using RADEV1 to avoid confusion.

Yes that would be OK although this is another case where user defined filters would be ideal.
> 
>     3. Maybe have a "Show Additional Info" checkbox/menu option somewhere that hides Last Update and Last TX unless enabled? (Combined with moving Last TX and Last Update so they're next to each other.)

I don't see a problem with that, but removing the years would save 10 chars width.

There is a feature in klog which lets the user select which columns appear in the main logging table.
User defined column visibility in the reporter would solve Peter's issue. :) 

All these custom requests for special filtering and different column visibility and order could all be resolved with user configurable settings.

tmiw:
Reverted the Locator/Distance changes. Maybe we can table the other proposed changes (i.e. being able to hide columns) until work on the filtering changes happens.

Tyrbiter:
> 2. Hide Mode entirely unless legacy modes are enabled. As part of this, auto-filter out anyone not using RADEV1 to avoid confusion.

The only snag with this is it makes legacy users mostly invisible, but probably only those using ezDV stick with these modes now. Slightly rude but at least there is an option to show them.

tmiw:
> > 2. Hide Mode entirely unless legacy modes are enabled. As part of this, auto-filter out anyone not using RADEV1 to avoid confusion.
> 
> The only snag with this is it makes legacy users mostly invisible, but probably only those using ezDV stick with these modes now. Slightly rude but at least there is an option to show them.

Yeah, I'm thinking we need to implement the ability for users to hide columns instead of making decisions like that.

tmiw:
The latest should now have the ability to rearrange columns in the FreeDV Reporter window and preserve this ordering between restarts. Let me know how this works for you (tested on Linux and macOS so far, not yet on Windows).

barjac:
> The latest should now have the ability to rearrange columns in the FreeDV Reporter window and preserve this ordering between restarts. Let me know how this works for you (tested on Linux and macOS so far, not yet on Windows).

That is working well - not sure what I would change about the order as we thrashed that out years ago!

However, by moving less useful columns to the far right out of sight, the width of the whole window can be reduced to the point where it is only limited by the buttons etc. in the footer.
Maybe rather than blocking further width reduction, the footer could re-arrange into two rows?

AND/OR:

We could totally remove 'Stop' as I don't think I have ever used it!
'Website' could be moved to an entry below 'FreeDV Reporter' as 'Web Reporter' in the 'Tools' menu.
'Message:' could be reduced to 'Msg:'
'Send' could have less padding.
'Clear' could be in the Msg context menu.
'Request QSY' could be 'Req QSY' with less padding.

That may then free up enough horizontal space to suit Peter. :)

BTW I guess that a user defined order like that may need to be named (Net?) and saved in a pick list in a menu along with 'Default' etc. so any can be selected for use.

peterbmarks:
Just tried out column reordering on Linux. Really great! Thank you very much.

tmiw:
The FreeDV Reporter window now has a menubar to allow individual columns to be hidden:

Screenshot 2025-12-18 at 1 05 08 AM

This is also preserved across restarts. :)

barjac:
> The FreeDV Reporter window now has a menubar to allow individual columns to be hidden:
>.....
> This is also preserved across restarts. :)

Nice!
However it is of little use without tidying up the footer to allow it to be compressed to half its current length. I can see two approaches.

1. Make the footer switch to two rows if the current elements no longer allow further compression. This would need some (e.g. track, freq, band) that are closely related grouping so they cannot be split across rows. Also maybe removal of useless 'Close' button.  
2. If we have to use a whole row with a single menu item, then move half of the footer items up there.

Here is a possible use case where the reporter has been cut down to it's simplest. With the existing footer it cannot be reduced in size any further.

 
Screenshot_20251219_151447

tmiw:
> > The FreeDV Reporter window now has a menubar to allow individual columns to be hidden:
> > .....
> > This is also preserved across restarts. :)
> 
> Nice! However it is of little use without tidying up the footer to allow it to be compressed to half its current length. I can see two approaches.
> 
> 1. Make the footer switch to two rows if the current elements no longer allow further compression. This would need some (e.g. track, freq, band) that are closely related grouping so they cannot be split across rows. Also maybe removal of useless 'Close' button.
> 2. If we have to use a whole row with a single menu item, then move half of the footer items up there.

Try now? (I accidentally committed the needed changes to v2.2-dev, so I had to merge back down here.)

Also, it's apparently not properly preserving the column order on Windows, so I'll need to fix that before merge. (Column hiding appears to be working fine, though.)

barjac:
The snag with combining the 'version' column into 'status' is that you can't remove 'status' and keep 'version', which I would prefer, as the status column is useless because nobody uses it as intended. It's also useful to be able to sort on version which I do often to see who is testing what.
With the ability to remove columns at will there is now no need to mash them together, anyone not wanting to see version can remove it.

Tyrbiter:
The latest causes this:

19:40:44 [4] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 0
19:40:44 [5] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 1
19:40:44 [6] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 2
19:40:44 [7] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 3
19:40:44 [8] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 4
19:40:44 [9] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 5
19:40:44 [10] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 6
19:40:44 [11] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 7
19:40:44 [12] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 8
19:40:44 [13] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 9
19:40:44 [14] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 10
19:40:44 [15] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 11
19:40:44 [16] INFO /home/bdm/git/freedv-gui-v2.2-dev-ms-reporter-ui-improvements/src/gui/dialogs/freedv_reporter.cpp:243: Creating col 12
Unhandled unknown exception; terminating the application.

New column code it would appear.

tmiw:
Hopefully 5b953c4 works better. Still looking into the issue with saving/restoring column ordering on Windows.

barjac:
Crash on start @ #5b953c4d

Unhandled standard exception of type "St12out_of_range" with message "vector<bool>::_M_range_check: __n (which is 13) >= this->size() (which is 13)"; terminating the application.

Tyrbiter:
> Crash on start @ #5b953c4d
> 
> Unhandled standard exception of type "St12out_of_range" with message "vector::_M_range_check: __n (which is 13) >= this->size() (which is 13)"; terminating the application.

Same thing here, so no https://github.com/drowe67/freedv-gui/commit/5b953c4d884785e6ad65d9d5fca0e75fe726bdef is not better :(

barjac:
Can the Dir column be left justified when in 'cardinal' mode please?

tmiw:
> Crash on start @ #5b953c4d
> 
> Unhandled standard exception of type "St12out_of_range" with message "vector::_M_range_check: __n (which is 13) >= this->size() (which is 13)"; terminating the application.

That's what I get for not testing prior to committing. How about fa0f08b?

Tyrbiter:
Seems to be working for me now, together with the merged logging feature which I will have to learn to use.

tmiw:
> Seems to be working for me now, together with the merged logging feature which I will have to learn to use.

Excellent. The most recent commits seem to have solved the column ordering preservation issues on Windows, too, so we can probably merge as soon as CI builds finish.

PR #1137, opened 2025-12-19
Experiment: use freethreaded Python.

tmiw:
Note: this is only for AppImages and the CI tests, not anything built manually.


tmiw:
This seems to work on Fedora but not Ubuntu for some reason, some error involving `_ARRAY_API` in NumPy. Closing for now and will revisit at a later date.

PR #1145, opened 2025-12-24
Add button to record decoded audio.

### LINKED ISSUE #1091: [Feature Request]
### Contact Details

jfminlg@verizon.net; tom5pennies@yahoo.com; KN6KHV@yahoo.com

### Feature for this project

[Proposed FreeDV feature-Quick Record of decoded audio.txt](https://github.com/user-attachments/files/23468459/Proposed.FreeDV.feature-Quick.Record.of.decoded.audio.txt)

It will help enhance HF Digital Voice for Amateur radio by Enablein memorialization of FreeDV audio decoding.

### VHF/UHF

No

### Deprecated features

No

### Non-standard compiler

No

### Legal worldwide

Yes

### Implementation assistance

Testing and documenting

### Support assistance

Testing future versions

tmiw:
This PR resolves #1091 by adding an additional button on the main window to record decoded audio from the RADE modem:

image

("Record" was also renamed to "Raw" to help differentiate between the two recording types.)

This creates a file named "FreeDV_FromDecoder_..." in the same location as configured for raw recordings (in Tools->Options->Audio->Quick Record).

### DIFF
--- a/.github/workflows/cmake-macos.yml
+++ b/.github/workflows/cmake-macos.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   lint:
-    runs-on: macos-26
+    runs-on: macos-15
 
     steps:
     - uses: actions/checkout@v4
@@ -23,6 +23,7 @@ jobs:
     - name: Install needed packages
       run: |
           brew install automake libtool numpy sox wxwidgets speexdsp portaudio libsndfile libsamplerate hamlib libebur128 
+          brew uninstall git # workaround, forces use of Apple git
 
     - name: Install LLVM and Clang
       uses: KyleMayes/install-llvm-action@v2
@@ -48,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-26]  # ARM64
+        os: [macos-15]  # ARM64
         sanitizer: [WITH_ASAN=0, WITH_ASAN=1, WITH_TSAN=1, WITH_UBSAN=1]
 
     runs-on: ${{ matrix.os }}
@@ -65,6 +66,7 @@ jobs:
     - name: Install needed packages
       run: |
         brew install automake libtool numpy sox wxwidgets speexdsp portaudio libsndfile libsamplerate hamlib libebur128
+        brew uninstall git # workaround, forces use of Apple git
 
     - name: Install VB-Cable
       shell: bash
@@ -137,6 +139,7 @@ jobs:
     - name: Install needed packages
       run: |
         brew install automake libtool numpy sox 
+        brew uninstall git # workaround, forces use of Apple git
 
     - name: Initialize code signing infrastructure
       if: ${{ vars.FREEDV_MACOS_CODE_SIGNING_TEAM_ID != '' }}
@@ -214,6 +217,7 @@ jobs:
     - name: Install needed packages
       run: |
         brew install automake libtool sox
+        brew uninstall git # workaround, forces use of Apple git
 
     - name: Install VB-Cable
       shell: bash


barjac:
This works fine however I think that different paths need to be configurable for each to avoid confusion.

Some users will not fully understand that Raw must be used for playback through FreeDV and that FromDecoder can be used to replay with any audio player or on air using the Voice Keyer.

It would also be useful to have both buttons show a small dialogue to allow the addition of a short text to the filename e.g. callsign etc. replacing any spaces with '_', for easy identification of the files later.

Filenames could also be shorter like:
FDV_Decoded_20251226-122648.wav
FDV_Raw_20251226-122648.wav

...and with text added at end to preserve sort order:
FDV_Decoded_20251226-122648_G2ZZZ.wav
FDV_Raw_20251226-122648_G2ZZZ.wav

My temporary stand-alone program has this and it is really useful having a callsign in the filename.

w4mmp-ron:
Hi,
Is this feature available for Windows?

barjac:
Nice that it can also be used to convert an existing raw recording to a decoded one by recording the playback of a 'FromRadio' file to include both sides of the QSO :)

With that in mind it would be really useful to be able to select .mp3 as an optional output format so decoded recordings could be sent by email when needed, as .wav are generally too large. Obviously not for replay with the VK.

Maybe an option dialogue for filename text could also have a .wav .mp3 option?

barjac:
> Hi, Is this feature available for Windows?

Not yet but it should be in the next release.

w4mmp-ron:
Hi,

Thanks.  I should have been more specific.  Is this feature available in 
a development release for Windows?  I'm very much looking forward to 
this feature.

73,
Ron / W4MMP

barjac:
> Hi, Thanks.  I should have been more specific.  Is this feature available in a development release for Windows?

I don't know how dev releases for Win are decided, Mooneer will no doubt comment on that :)

tmiw:
> > Hi, Thanks.  I should have been more specific.  Is this feature available in a development release for Windows?
> 
> I don't know how dev releases for Win are decided, Mooneer will no doubt comment on that :)

Development versions for Windows are always available from pull requests (albeit unsigned). Click on one of the Windows links at the bottom of the PR (you may need to expand the "checks" box):

image

Then click on Summary and go to Artifacts at the bottom of the page for the download link:

image

tmiw:
> Some users will not fully understand that Raw must be used for playback through FreeDV and that FromDecoder can be used to replay with any audio player or on air using the Voice Keyer.

Should the buttons be renamed to help with that as well? If so, any suggestions?

> With that in mind it would be really useful to be able to select .mp3 as an optional output format so decoded recordings could be sent by email when needed, as .wav are generally too large. Obviously not for replay with the VK.

I think libsndfile can do this, but I might need to investigate further (in case it needs additional dependencies that aren't necessarily available in the Windows or macOS build environment).

barjac:
> > Some users will not fully understand that Raw must be used for playback through FreeDV and that FromDecoder can be used to replay with any audio player or on air using the Voice Keyer.
> 
> Should the buttons be renamed to help with that as well? If so, any suggestions?
> 

No. I made that point to clarify my reasoning for suggesting a different path for the non-raw files.

> > With that in mind it would be really useful to be able to select .mp3 as an optional output format so decoded recordings could be sent by email when needed, as .wav are generally too large. Obviously not for replay with the VK.
> 
> I think libsndfile can do this, but I might need to investigate further (in case it needs additional dependencies that aren't necessarily available in the Windows or macOS build environment).

For conversion from .wav to .mp3 I used:
ffmpeg -i input.wav -codec:a libmp3lame -b:a 64k -ar 22050 output.mp3

I think just one record button as before would be better than using up more screen space.

I suggest:
On hitting Record open a pop-up box (like the logging confirmation but smaller) with check boxes for Raw, Decoded, wav and mp3, defaulting to Raw and wav. An entry box for a filename suffix like a callsign or whatever, with 'Cancel' and 'OK' buttons. (mp3 greyed if Raw selected)
On hitting OK the Record button would toggle to red as before and also stop the recording with another press.

For the same operation as before just hitting OK would record a raw .wav without a suffix.

Flexible extra functionality without any wasted screen space.

w4mmp-ron:
Hi,

OK.  Please see the attached screen shot.   What am I missing?  I don't 
see an option to record decoded audio.

73,
Ron / W4MMP

tmiw:
> Hi, OK.  Please see the attached screen shot.   What am I missing?  I don't see an option to record decoded audio. 73, Ron / W4MMP

Can you log in with a web browser and attach the screenshot there? It's not showing up for me so I think email stripped it out.

w4mmp-ron:
Here's the screen shot.
[2025-12-27_144159.zip](https://github.com/user-attachments/files/24357843/2025-12-27_144159.zip)

tmiw:
> Here's the screen shot. [2025-12-27_144159.zip](https://github.com/user-attachments/files/24357843/2025-12-27_144159.zip)

Thanks. Looks like the commit ID is incorrect--try [this](https://github.com/drowe67/freedv-gui/actions/runs/20495426245/artifacts/4964102436)?

tmiw:
> I think just one record button as before would be better than using up more screen space.
> 
> I suggest: On hitting Record open a pop-up box (like the logging confirmation but smaller) with check boxes for Raw, Decoded, wav and mp3, defaulting to Raw and wav. An entry box for a filename suffix like a callsign or whatever, with 'Cancel' and 'OK' buttons. (mp3 greyed if Raw selected) On hitting OK the Record button would toggle to red as before and also stop the recording with another press.
> 
> For the same operation as before just hitting OK would record a raw .wav without a suffix.
> 
> Flexible extra functionality without any wasted screen space.

So the original reason for the Record button in the main window is to provide a quicker method of starting a recording from the radio. In the past, you'd go to Tools->Start Record File - From Radio to provide additional parameters (e.g. filename and the number of seconds to record). That's why I'm fairly hesitant to add additional steps to the record process. I can see if I can shrink the amount of space needed for the buttons, as well as add a Tools menu option for additional recording options.

w4mmp-ron:
Hi,

Yep, that did the trick.  Tomorrow I will give the "Decoded" option a 
go.  I really want to have folks hear what I am hearing. Essentially 
everyone sounds the same.  It almost like the RADE engine is performing 
a text to speech function with the voice always somewhat the same.  I 
know that is not what is happening, but it somewhat describes my 
impression of the audio produced by RADE.

73,
Ron / W4MMP

barjac:
> So the original reason for the Record button in the main window is to provide a quicker method of starting a recording from the radio. In the past, you'd go to Tools->Start Record File - From Radio to provide additional parameters (e.g. filename and the number of seconds to record). That's why I'm fairly hesitant to add additional steps to the record process. I can see if I can shrink the amount of space needed for the buttons, as well as add a Tools menu option for additional recording options.

All understood but the old original way was really clunky.

This seems to be a retrograde step as regards screen use, now in width.

I envisaged something like this mock-up on pressing the Record button with the defaults shown:


Screenshot_20251229_152504

It is only one more click for instant raw recording but offers so much more, 'almost' instantly.

Anything requiring a trip into the Tools menu is a real turn-off for many users.

tmiw:
OK, I did the following as of 8a377ed:

1. Removed Tools->Start Record File... options from the menu bar (Record button is the only way to start recordings).
2. Went back to a single Record button. This button pops up something similar to what the old Tools->Start Record File... options did, just with a different set of controls on the bottom for raw/decoded recording (raw is default). Example (macOS):

image

3. Removed Quick Record config options from Tools->Options since we're now using the method the old Tools menu options used to preserve the last used folder location.

Hopefully this works for you @barjac?

barjac:
> 
> Hopefully this works for you @barjac?

No not really, but  Happy New Year!

The record button now opens a system file/directory chooser which I think is overkill and takes the user out of FreeDV - I would prefer that the directories for saved files were fixed by user config as before with just the addition of a new path for when decoded audio is being saved.

When saving decoded recordings the path MUST be different if .wav is used for both! 

I have ~/freedv/freedv-rec-raw/ for 'off-air' and ~/freedv/freedv-rec-dec/ for decoded so the two are kept separate, otherwise you have to play them to distinguish. 

I would like the path to be automatically selected depending on which type of recording I select without any system file chooser being needed.

I would also like to optionally enter a callsign or similar identifier to the end of the filename (ahead of .wav) without manually editing the filename (which is error prone).

My previous suggestion of a small pop-up catered for this neatly without resorting to a clunky (in plasma it's not great) system file chooser.

I don't mind using the file chooser in Tools->Options->Audio to set the VK, Raw and Decoded recording paths once but please not during a QSO to make a 'quick' recording.

tmiw:
How about [b239585](https://github.com/drowe67/freedv-gui/pull/1145/commits/b239585aa8b534fce3580743a4e33ee245f4b460)?

barjac:
Yes! This is nice!

I think one extra check box for optional mp3 creation in the dialogue would finish this off nicely!
mp3 could use the same file name and directory with the mp3 extension.

Often users want to email recordings and wav is too large for this. Adding an mp3 by converting from the .wav at the end of the recording seems the most cost effective way, without heavy cpu usage during the recording.
An mp3 is typically only ~5% of the size of wav but is still good enough quality for this use.
I'm not sure if most systems would have ffmpeg with libmp3lame, but this is what I have used in testing for the conversion:
ffmpeg -i input.wav -codec:a libmp3lame -b:a 64k -ar 22050 output.mp3

barjac:
Also maybe 'Off Air' rather than 'Raw' has been suggested, and maybe 'Start' rather than 'OK'?

tmiw:
> I think one extra check box for optional mp3 creation in the dialogue would finish this off nicely! mp3 could use the same file name and directory with the mp3 extension.

Done as of [0688656](https://github.com/drowe67/freedv-gui/pull/1145/commits/06886567b48b8ecd548d4c23cb5a737e34b60f26).

> Also maybe 'Off Air' rather than 'Raw' has been suggested, and maybe 'Start' rather than 'OK'?

And done.

barjac:
This seems to work fine but I have some questions:
1. I assume that 'Off-air' raw .wav is only intended for later replay through the FreeDV decoder, as I can't think of another use for a raw recording? I was thinking that maybe mp3 should be greyed when Off-air is selected, however there is the case where an SSB recording may be wanted for comparison with FreeDV so maybe it should stay?
2. Is mp3 actually suitable for voice keyer use? I discovered that it does transmit OK but I have not yet had any feedback on the received results. If this is acceptable then maybe the VK file chooser should not default to only .wav when launched?
3. Does the recording of mp3 create a .wav then convert it, or is it compressed on the fly?
  If the former then by checking both, could the .wav be retained rather than deleted, or alternatively could they both be created at the same time?
4. I tried to record the decoded output of a file 'played from radio' but the new recording was silent. In my external program I could do that to get both sides of the conversation into an mp3. Is this not possible within FreeDV?

Some explanation will be needed in tooltips for all the options in the dialogue.
After you answer the above I will be happy to prepare some text suggestions for these.

tmiw:
> 1. I assume that 'Off-air' raw .wav is only intended for later replay through the FreeDV decoder, as I can't think of another use for a raw recording? I was thinking that maybe mp3 should be greyed when Off-air is selected, however there is the case where an SSB recording may be wanted for comparison with FreeDV so maybe it should stay?

Considering that Analog vs. FreeDV decoding can be changed on the fly and that MP3 isn't a lossless format, it's probably not a good idea to have MP3 be an option for raw/off-air recording. I'll update the PR accordingly.

> 2. Is mp3 actually suitable for voice keyer use? I discovered that it does transmit OK but I have not yet had any feedback on the received results. If this is acceptable then maybe the VK file chooser should not default to only .wav when launched?

Maybe (with the caveat that I'm not sure how RADE will behave on the decompressed audio given again, MP3 isn't a lossless format).

> 3. Does the recording of mp3 create a .wav then convert it, or is it compressed on the fly?
>    If the former then by checking both, could the .wav be retained rather than deleted, or alternatively could they both be created at the same time?

Compressed on the fly. Having both WAV and MP3 be created at the same time would add more complication in the area of the code where we tap into the audio stream to make the recording, so I'm hesitant to do this.

> 4. I tried to record the decoded output of a file 'played from radio' but the new recording was silent. In my external program I could do that to get both sides of the conversation into an mp3. Is this not possible within FreeDV?

It should still be possible as that's how I tested the MP3 changes on my M1 Mac.

barjac:
Re. 1 OK
Re. 2 Could the file chooser please default to *.* at least for now while we do some testing?
I know it can be done manually but it's just another step that gets in the way.
Re. 3 OK understood.
Re. 4 OK re-tested with both wav and mp3 and it is OK.

One minor issue I came across is that if you are in transmit and start voice keyer to play a file over the air, the file plays OK but the plot is not shown in from mic. Starting VK from cold does show in the from mic.

tmiw:
> 2 Could the file chooser please default to _._ at least for now while we do some testing? I know it can be done manually but it's just another step that gets in the way. 

To clarify, these are the file choosers available via the Voice Keyer button, right? Those look like they're supposed to default to the last path selected, even across restarts of FreeDV, at least according to the code. Should that persistence just be ignored and instead, always use the current folder in the terminal at the time of starting the app? Or am I misunderstanding and you actually mean something else?

barjac:
Yes the VK file selector.
Its good that it defaults to the last used folder used from FreeDV but it currently filters by default to .wav only, so .mp3 are not visible.
It would be better if it could default to showing both .wav and .mp3 or if that is not possible to 'All files'.

Also would it be possible to allow this file selection to be done while in transmit?

Also a really neat addition would be to auto-fill the filename suffix field with a callsign IF a row in either the pull-down or reporter was selected prior to hitting the record button. Similar to logging.

tmiw:
> Yes the VK file selector. Its good that it defaults to the last used folder used from FreeDV but it currently filters by default to .wav only, so .mp3 are not visible. It would be better if it could default to showing both .wav and .mp3 or if that is not possible to 'All files'.
> 
> Also would it be possible to allow this file selection to be done while in transmit?
> 
> Also a really neat addition would be to auto-fill the filename suffix field with a callsign IF a row in either the pull-down or reporter was selected prior to hitting the record button. Similar to logging.

All done as of [8264eda](https://github.com/drowe67/freedv-gui/pull/1145/commits/8264eda9fc717e1f919b0b5a62c8a99a31c1406a).

barjac:
Super - working fine!

Issue #983, opened 2025-07-23
Experimental Vox mode for PTT


barjac:
### Contact Details

G6WPJ G4MKT more will be added I am sure!

### Feature for this project

In the UK we are now establishing a regular Top Band FreeDV evening net which is rapidly gaining popularity.
It is traditional on Top Band in the UK to operate with Vox on SSB in a very relaxed manner and we have been discussing whether or not this would be feasable with FreeDV.

These nets are VERY different to typical US nets. No Net controller or anything like that, there can be long silent periods where nobody speaks, then a comment or two. 

The rapid sync would seem to allow this and we have done some testing using fast exchanges using the PTT, but would now like to experiment with Vox in FreeDV.

Maybe initially this could use the 'Experimantal features' option to enable it while testing.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Non-standard compiler

None

### Implementation assistance

As always we would give as much assistance as possible 

### Support assistance

As much as possible, however this is just an experiment for which the basics with enough functionality to test the concept would be adequate.

tmiw:
There's currently about 200-250ms latency before the radio switches from TX->RX. That doesn't seem like enough for VOX right now, right?

barjac:
If the TX audio was detected before encoding to trigger PTT then how long would encoding take? 
The two delays may cancel so no audio would be lost.
The decode delay may be a small advantage in that RX stations would not hear a pause until after a TX had actually stopped, although the reporter may notify faster?
Some changes in user technique may be needed compared to traditional VOX usage but that may not be a show stopper.  
Lots of variables, which is why it would be good to have the basics to play with and adjust, to evaluate feasibility in operation.
As I said initially, purely an experiment.

tmiw:
Not sure on your questions. This may require changes similar to what would be needed for #987 to get something basic working (or perhaps we can just require full duplex as a prerequisite for VOX?)

barjac:
> or perhaps we can just require full duplex as a prerequisite for VOX?

Preferably not if it can be avoided, unless it reduces delays and provides better performance for vox.

I had an idea to test vox using a backgrounded script to sample the mic audio ahead of freedv to set a vox trip level and delay. 
The script could produce a tone in the other stereo channel input to freedv for PTT.
I will have a play with GnuRadio. ;)

tmiw:
Thanks, I'll schedule this for further discussion with the project team.

barjac:
[...]

PR #1151, opened 2025-12-27
Suspend PTT input changes while transitioning TX<->RX.

tmiw:
Due to feedback in #983, this PR suspends changes to PTT input while transitioning between TX and RX. This should hopefully prevent the main window from being stuck on a plot that's not expected.


barjac:
This is working perfectly, I tried during a couple of QSOs to cause the plot problem and could not make it happen using VOX when speaking just after/at timeout.
No regressions that I can see - thanks!

PR #1162, opened 2026-01-02
FreeDV Reporter: Defer row adds and removes until timer fires

tmiw:
This PR eliminates the immediate add/remove that was performed when the application received a user connect or disconnect message from the server. Adds/removes are now done when the highlight update timer fires, enabling multiple adds and removes to be done in bulk (and thus reducing flicker when filtering rows).

### DIFF
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -581,6 +581,7 @@ FreeDVReporterDialog::FreeDVReporterDialog(wxWindow* parent, wxWindowID id, cons
     m_listSpots->Connect(wxEVT_DATAVIEW_ITEM_CONTEXT_MENU, wxDataViewEventHandler(FreeDVReporterDialog::OnItemRightClick), NULL, this);
     m_listSpots->Connect(wxEVT_DATAVIEW_COLUMN_HEADER_CLICK, wxDataViewEventHandler(FreeDVReporterDialog::OnColumnClick), NULL, this);
     m_listSpots->Connect(wxEVT_DATAVIEW_COLUMN_REORDERED, wxDataViewEventHandler(FreeDVReporterDialog::OnColumnReordered), NULL, this);
+    m_listSpots->Connect(wxEVT_SET_FOCUS, wxFocusEventHandler(FreeDVReporterDialog::OnSetFocus), NULL, this);
 
     m_statusMessage->Connect(wxEVT_TEXT, wxCommandEventHandler(FreeDVReporterDialog::OnStatusTextChange), NULL, this);
     m_buttonSend->Connect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FreeDVReporterDialog::OnStatusTextSend), NULL, this);
@@ -648,6 +649,7 @@ FreeDVReporterDialog::~FreeDVReporterDialog()
     m_listSpots->Disconnect(wxEVT_DATAVIEW_ITEM_CONTEXT_MENU, wxDataViewEventHandler(FreeDVReporterDialog::OnItemRightClick), NULL, this);
     m_listSpots->Disconnect(wxEVT_DATAVIEW_COLUMN_HEADER_CLICK, wxDataViewEventHandler(FreeDVReporterDialog::OnColumnClick), NULL, this);
     m_listSpots->Disconnect(wxEVT_DATAVIEW_COLUMN_REORDERED, wxDataViewEventHandler(FreeDVReporterDialog::OnColumnReordered), NULL, this);
+    m_listSpots->Disconnect(wxEVT_SET_FOCUS, wxFocusEventHandler(FreeDVReporterDialog::OnSetFocus), NULL, this);
 
     m_statusMessage->Disconnect(wxEVT_TEXT, wxCommandEventHandler(FreeDVReporterDialog::OnStatusTextChange), NULL, this);
     m_buttonSend->Disconnect(wxEVT_COMMAND_BUTTON_CLICKED, wxCommandEventHandler(FreeDVReporterDialog::OnStatusTextSend), NULL, this);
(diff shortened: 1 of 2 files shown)


tmiw:
Second attempt at eliminating the flicker that @barjac noticed in a previous PR.

@barjac, any better here? Or does this still cause crashes for you?

barjac:
No crashes now running just this PR.
I did notice in a test QSO on a busy 40m band that it took 3 attempts to 'select' the station I was working while filtered on an odd freq with only us showing.
He did not appear in the drop-down due to SSB splatter so was using the reporter to log him.
While on the subject of needing multiple clicks I am still encountering this with Start/stop modem and PTT. This has been an issue for a very long time and goes back to pre-rade. If I accidentally 'click' these rather than 'press-hold-release' nothing happens. Is there a better way of detecting these events?

tmiw:
> I did notice in a test QSO on a busy 40m band that it took 3 attempts to 'select' the station I was working while filtered on an odd freq with only us showing. He did not appear in the drop-down due to SSB splatter so was using the reporter to log him. 

Checked in a change to hopefully improve that.

> While on the subject of needing multiple clicks I am still encountering this with Start/stop modem and PTT. This has been an issue for a very long time and goes back to pre-rade. If I accidentally 'click' these rather than 'press-hold-release' nothing happens. Is there a better way of detecting these events?

Maybe this is because they're toggle buttons? I'm assuming none of the other buttons in the application (i.e. in the newly added logging dialog) do this, right?

barjac:
> > I did notice in a test QSO on a busy 40m band that it took 3 attempts to 'select' the station I was working while filtered on an odd freq with only us showing. He did not appear in the drop-down due to SSB splatter so was using the reporter to log him.
> 
> Checked in a change to hopefully improve that.
> 

Good - I will test. BTW This affects both the pull down and the reporter row selection and its more than just initial lack of focus.

EDIT I have just figured out how to replicate this in a different branch (ms-rec-decoded-audio). If the mouse is totally stationary over a callsign etc. in a reporter row you can click for ever without it selecting. Move the mouse 1 pixel and the next click selects OK.

EDIT2 Testing now in this branch and your fix is fine - Super! Err.. However it does not select when mouse is over the message field. Maybe its looking at fields rather than rows if that's an option?

> > While on the subject of needing multiple clicks I am still encountering this with Start/stop modem and PTT. This has been an issue for a very long time and goes back to pre-rade. If I accidentally 'click' these rather than 'press-hold-release' nothing happens. Is there a better way of detecting these events?
> 
> Maybe this is because they're toggle buttons? I'm assuming none of the other buttons in the application (i.e. in the newly added logging dialog) do this, right?

True, I have not noticed this with other buttons.

tmiw:
> EDIT I have just figured out how to replicate this in a different branch (ms-rec-decoded-audio). If the mouse is totally stationary over a callsign etc. in a reporter row you can click for ever without it selecting. Move the mouse 1 pixel and the next click selects OK.
> 
> EDIT2 Testing now in this branch and your fix is fine - Super! Err.. However it does not select when mouse is over the message field. Maybe its looking at fields rather than rows if that's an option?

I think it's because of the tooltip that appears. I've noticed (on macOS anyway) that the first click closes the tooltip, while the second click (without moving the mouse) will select. This might just be normal behavior.

> True, I have not noticed this with other buttons.

I can't say I've noticed this myself but it could be DE specific. Maybe @Tyrbiter can chime in?

Tyrbiter:
I have just started using https://github.com/drowe67/freedv-gui/pull/1145/commits/8264eda9fc717e1f919b0b5a62c8a99a31c1406a and a little bit of pointing and clicking has not managed to show the symptoms that @barjac sees, but I have over a fairly long period seen some oddities in mouse-click behaviour in freedv-gui which are really difficult to capture but happen infrequently.

Occasionally I see a failure to register a click, but it might be because I have moved the mouse while attempting to click. Sometimes a mouse movement over the message column doesn't show the full message tooltip. But it's really hard to identify exactly why that was the case.

This could be because of the ever-changing nature of a current Fedora release, all the underlying libraries are in flux and can change anywhere from 6 monthly to daily. Maybe this affects wxWidgets, but I think the actual wxGTK3 packages have been stable for a while now.

I wish I could be more exact. Is there something precise that I can do to identify if this is happening. Mouse control to <1px is not the easiest thing to do precisely but I can try if it's essential.

I have just tried the "move mouse over callsign and right click" to select a row, so far repeated tests (15 attempts) have shown 15 rows selected. Maybe it's something specific in Mageia, but I am unsure of that as I would expect most distros to use the most recent libraries. Fedora doesn't really have a stable release but then the development Mageia version should behave similarly because it will follow the most recent libraries too.

Just realised that I may not be using the code in this PR, is it worth merging the latest in PR #1145 into this branch and I can try again with that.

barjac:
#dc21f92 in Mageia Cauldron plasma-6 Wayland
Clicking in the reporter selects the current row with one left click, except anywhere in the Message column, where it takes two clicks, irrespective of whether there is any message.

Could left click on a message field be made to close the tooltip AND create another left click event to avoid the two click problem? There is no other left click action that I am aware of related to the reporter table.

There is also the long standing issue of removing a selection, at present only left clicking in a clear area of the reporter footer will clear a selection. A secret that most users won't find.

EDIT: I just discovered that right click clears a selection - who knew!
It may be more intuitive to make a selection toggle on each left click if possible. This would avoid conflicting with the right click to get to the message copy feature.

tmiw:
> Could left click on a message field be made to close the tooltip AND create another left click event to avoid the two click problem? There is no other left click action that I am aware of related to the reporter table.

I've been doing some tests to see if we can capture the left-click event for the tooltip window. Unfortunately it doesn't look like we can capture it on either macOS or Linux (Fedora 43 using Plasma on Wayland). I'll see if I can at least hide the tooltip if there's no message, though, which should allow left-click to behave properly hopefully.

tmiw:
Actually, I think I forgot some additional code needed for that to work. I'll keep testing and will let you know.

tmiw:
Never mind, original comment stands. Try [c126ee3](https://github.com/drowe67/freedv-gui/pull/1162/commits/c126ee37aa3ca2f13b242cc1323d007b67a94577)?

tmiw:
I think the left-click behavior as of [43c2c84](https://github.com/drowe67/freedv-gui/pull/1162/commits/43c2c848b4efb73fe5dfa5f26cf2627f4a57650a) is a bit better, but not 100% unfortunately (and I don't think we can fully get there). Let me know how this goes, @barjac.

Tyrbiter:
> I think the left-click behavior as of [43c2c84](https://github.com/drowe67/freedv-gui/pull/1162/commits/43c2c848b4efb73fe5dfa5f26cf2627f4a57650a) is a bit better, but not 100% unfortunately (and I don't think we can fully get there). Let me know how this goes, @barjac.

Well it has passed all the tests... ;-)

PR #1142, opened 2025-12-24
FreeDV Reporter: Add idle filter.

tmiw:
This PR resolves #1056 by implementing an idle filter option in the FreeDV Reporter menu bar. 

Example of what this menu option looks like:

Screenshot 2025-12-23 at 5 24 44 PM

"Idle" is defined for a user as either:

* Last transmitted more than X minutes ago, or if no TX yet:
* Having connected more than X minutes ago.

Issue #1101, opened 2025-11-20
[Feature Request] Reporter filtering improvements for now and the future


barjac:
### Contact Details

G4MKT G6WPJ GW4GTE GW8LJJ 

### Feature for this project

Over the last few months FreeDV usage has increased considerably and this is causing problems for users of the FreeDV Reporter.

In our opinion the only way to keep the number of stations shown in the reporter at any one time relevant to the user, will be to allow 'user configurable filters'.

As things stand now I will try to summarize the comments I am hearing and the problems being encountered:
Station cannot keep track of the stations in the group as the list is dropping off the page.
Most stations on the frequency are only listening but taking up screen space.
Stations on the same frequency showing on the list are out of radio range (daytime NVIS) e.g. USA or Japan showing in the UK.
Stations on the list are camping there 24/7 and causing over long lists that no longer even fit a full screen.
The list goes on...

All these issues could be solved with lots of dedicated filters, however that would cause endless feature requests for different use cases.

A structured, user configurable filtering scheme is needed.
The user would be able to create named filters for selection in different use cases in an easy to use GUI.
Each filter could use any or many reporter column values to narrow down the displayed stations.

Two possible examples:
UK_NVIS_net:- Exclude distance greater than 1000 miles, exclude mode RX Only, exclude Last TX greater than 30mins ago, exclude where Msg does not start with a numeric character (in UK nets we use this to indicate that a station wishes to TX and the order of net).

DX_20m:- Exclude distance less than 2000 miles, exclude mode RX Only, exclude Last TX greater than 1 hour ago.

Obviously we can't expect users to insert regex, but the following could be used in setting up filter conditions in a small table with one condition per row:
```
 < less than
 > greater than
 = equal to 
 != not equal to
 ^ starts with
 N numeric character
 ~= contains

Filter Name: | UK NET
___________________________________
1 | Dist      | <  | 1000          |   ( Dist in current reporter units
2 | Status    | != | "RX Only"     |   
3 | Last TX   | >  | 60            |   (mins)
3 | Msg       | ^  | N             |

```
I appreciate that this would involve a fair amount of work, however in the long run it would 'future-proof' the reporter against more and more filtering requests as user numbers increase.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

Help with structure ideas

### Support assistance

Testing as always.

barjac:
[...]

PR #1166, opened 2026-01-04
FreeDV Reporter: Add filtered indicator to bottom of window.

tmiw:
Per request in #1142, this PR adds a filter on/off indicator to the bottom of the FreeDV Reporter window to indicate whether one of the available FreeDV Reporter filters are enabled.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -905,6 +905,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Flex: add command line option to adjust RX volume. (PR #1159)
     * Flex: Allow user-configurable spot timeout (default 10 minutes). (PR #1169)
     * Display Hamlib version in Help->About. (PR #1169)
+    * FreeDV Reporter: Add right-click menu for callsign lookups. (PR #1171)
 3. Build system:
     * Use Clang to build AppImages for better performance. (PR #1149)
     * Enable link-time optimization for AppImages, DMGs and Windows builds. (PR #1163)
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -539,6 +539,31 @@ FreeDVReporterDialog::FreeDVReporterDialog(wxWindow* parent, wxWindowID id, cons
         wxCommandEventHandler(FreeDVReporterDialog::OnCopyUserMessage),
         NULL,
         this);
+        
+    // Create popup menu for callsign lookups
+    callsignPopupMenu_ = new wxMenu();
+    assert(callsignPopupMenu_ != nullptr);
+    
+    auto qrzMenuItem = callsignPopupMenu_->Append(wxID_ANY, _("Lookup via QRZ"));
+    callsignPopupMenu_->Connect(
+        qrzMenuItem->GetId(), wxEVT_COMMAND_MENU_SELECTED, 
+        wxCommandEventHandler(FreeDVReporterDialog::OnQRZLookup),
+        NULL,
+        this);
+        
+    auto hamQthMenuItem = callsignPopupMenu_->Append(wxID_ANY, _("Lookup via HamQTH"));
+    callsignPopupMenu_->Connect(
+        hamQthMenuItem->GetId(), wxEVT_COMMAND_MENU_SELECTED, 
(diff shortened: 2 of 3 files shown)


barjac:
I was expecting something a bit more eye catching, maybe using colour.
This took a while to notice and I was looking for it ;)
Also to save screen height I often remove the reporter window decoration so that warning would not be visible.

tmiw:
> I was expecting something a bit more eye catching, maybe using colour. This took a while to notice and I was looking for it ;) Also to save screen height I often remove the reporter window decoration so that warning would not be visible.

Color where, exactly?

barjac:
> > I was expecting something a bit more eye catching, maybe using colour. This took a while to notice and I was looking for it ;) Also to save screen height I often remove the reporter window decoration so that warning would not be visible.
> 
> Color where, exactly?

Good question :)

Maybe add a small (square) button with no legend in the reporter footer that is normally green.
If filtering is on make it amber.
When green, tooltip says 'No warnings'
When amber, tooltip displays 'Warning: Idle filter is active'
This could be used for any warnings and use green/grey, amber or red.

tmiw:
> Good question :)
> 
> Maybe add a small (square) button with no legend in the reporter footer that is normally green. If filtering is on make it amber. When green, tooltip says 'No warnings' When amber, tooltip displays 'Warning: Idle filter is active' This could be used for any warnings and use green/grey, amber or red.

How about "Filters:" next to the Band drop-down? No coloring if no filters are enabled, amber otherwise?

barjac:
In order to be able to make real use of the 'Show' feature to actually reduce the width of the reporter, then the footer needs to be shorter not longer.
This is why I suggested just a tiny button as a warning light.

However, yes that could work, possibly to pave the way towards fully user configurable filters :) If your 'Filters' was a button it could be a quick toggle to enable/disable the filter, still with colour.

I still feel that tacking on random filters as they are suggested like this must be more time-consuming in the long run than going for a fully structured user-configurable filter system as outlined in [#1101](https://github.com/drowe67/freedv-gui/issues/1101).

tmiw:
> In order to be able to make real use of the 'Show' feature to actually reduce the width of the reporter, then the footer needs to be shorter not longer. This is why I suggested just a tiny button as a warning light.
> 
> However, yes that could work, possibly to pave the way towards fully user configurable filters :) If your 'Filters' was a button it could be a quick toggle to enable/disable the filter, still with colour.

How about now with [801e178](https://github.com/drowe67/freedv-gui/pull/1166/commits/801e17828ffed33497c3bfe3c73dc9051a4bfaed)? This is just a text label for now and we can revisit in a future PR if additional filtering functionality's added.

barjac:
#801e178
As this now stands it defeats the object, as most users will always be tracking either band or frequency which currently turns the indication on. There is always an indication of the tracking filter state in the reporter already, so maybe 'Idle' would be better than 'Filter' for the text label for now.

The new 'Filters on' should say 'Idle' and be greyed or orange to ONLY warn about the state of the idle filter. 

Otherwise it works fine :)
Except that setting an idle time (especially custom) does not indicate what that time is. Maybe the text label could say 'Idle 15' etc. when orange?

tmiw:
> #801e178 As this now stands it defeats the object, as most users will always be tracking either band or frequency which currently turns the indication on. There is always an indication of the tracking filter state in the reporter already, so maybe 'Idle' would be better than 'Filter' for the text label for now.
> 
> The new 'Filters on' should say 'Idle' and be greyed or orange to ONLY warn about the state of the idle filter.
> 
> Otherwise it works fine :)

If this is intended only for the idle filter, we're already showing enabled/disabled in the menu bar (example below with the filter set to 30 minutes):

image

Granted, it's not as easily accessible as having it in the bottom row (and may not be visible at all without window decorations, not sure). I'll think some more.

barjac:
My point is that when the idle filter is on you have no indication unless you actually open the menu option to go and check. The idea of a warning is to tell you without you going to check :)

Could colour be applied to the 'Filter' menu text?
Show and Filter are still visible with the decorations removed.

tmiw:
> Could colour be applied to the 'Filter' menu text? Show and Filter are still visible with the decorations removed.

Probably not in anything near a platform-independent manner, if it can be done at all. 

Anyway, see [a036e31](https://github.com/drowe67/freedv-gui/pull/1166/commits/a036e314b1a491be8dda1eb6a2438c7a42ff325b) and let me know how it goes.

PR #1180, opened 2026-01-15
FreeDV Reporter: Use connect time reported by server for idle filter.

tmiw:
This PR uses the connect time reported by the server for the idle filter instead of Last Update when there's no TX time available. This is because Last Update is not 100% reliable (e.g. initial startup will show everyone as having connected at the same time).

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -911,7 +911,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * FreeDV Reporter:
         * Allow columns to be rearranged and/or made invisible. (PR #1132)
         * Sort empty user messages below non-empty ones. (PR #1105)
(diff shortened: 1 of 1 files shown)


barjac:
I had a few seg faults with this branch tonight - nothing in debug log and gdb was not running.

The qrz etc. lookup is nice - very handy but it would be better if there was the option to put credentials in FreeDV settings so it could log in. Klog does this as it has the credentials that are also needed for uploading logs to qrz etc.

tmiw:
> I had a few seg faults with this branch tonight - nothing in debug log and gdb was not running.

Is your system set up to save core dumps somewhere? I think it's somewhere in /var by default but I don't recall at the moment.

tmiw:
I'm going to go ahead and merge because I don't think this change is causing the issue reported.

PR #1182, opened 2026-01-18
Try some changes to mitigate empty recording bug.

tmiw:
WIP. This PR is to fix a bug reported on the digitalvoice mailing list where pushing the Record button results in an empty audio file.


tmiw:
Haven't heard back so closing for now.

PR #1195, opened 2026-01-26
Hamlib: Increase timeout to 625ms.

tmiw:
Increases Hamlib timeout to 625ms. Additionally, only set timeout if not using FLrig/rigctld and if the default timeout is longer than 625ms.

(Fixes problem discussed on Discord: https://discord.com/channels/1053600121381716069/1462561587536593066)


tmiw:
@barjac / @Tyrbiter -- maybe also try this branch as a sanity check? Seems to work with SmartSDR for Mac for me, anyway.

PR #1184, opened 2026-01-18
Add mouseover and deselection support for main window callsign list.

tmiw:
Resolves #1181 by adding support for selecting items from the main window dropdown list (and deselecting via right-click).

PR #1196, opened 2026-01-26
Use last selected row (FreeDV Reporter, main window) instead of prioritizing main window

tmiw:
Due to continued issues with phantom selections, this PR tracks the last selected row for logging purposes (either the FreeDV Reporter window or the drop-down at the bottom of the main window). See https://github.com/drowe67/freedv-gui/pull/1184#issuecomment-3799002571 for more info.

### DIFF
--- a/CHANGELOG_OLD.md
+++ b/CHANGELOG_OLD.md
@@ -1,5 +1,32 @@
 # Changes in older releases
 
+## V2.1.0 November 2025
+
+1. Bugfixes:
+    * Clean up wxWidgets warnings in Audio Config window. (PR #1044)
+    * Fix CAT control and frequency formatting problems on non-English computers. (PR #1050)
+    * Clean up ThreadSanitizer warnings in codebase. (PR #1053)
+    * Clean up UndefinedBehaviorSanitizer warnings in codebase. (PR #1061)
+    * Divide displayed TX Attenuation level by 10 on startup. (PR #1060)
+    * Fix issue preventing FreeDV Reporter from properly being restored to the correct display. (PR #1055)
+    * Fix intermittent FreeDV Reporter related crash on Windows. (PR #1075)
+    * Update Visual Studio Redistributable to version 17 to resolve RADE startup crashes. (PR #1074)
+2. Enhancements:
+    * Additional cleanup of code that runs in real-time. (PR #1039, #1068, #1077)
+    * Print error if one attempts to use a stereo WAV file instead of a mono one. (PR #1052)
+    * Hamlib: Add support for automatically displaying only the valid baud rates. (PR #1057)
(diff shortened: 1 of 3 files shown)


Tyrbiter:
Trying this branch, if PR #1195 works should it be merged to try it all up as I imagine the rigctld timeout change is not going to make much difference unless on macOS.

tmiw:
> Trying this branch, if PR #1195 works should it be merged to try it all up as I imagine the rigctld timeout change is not going to make much difference unless on macOS.

Done.

barjac:
As regards the 'use last selected' that seems to now be OK although needs more testing when enough stations are around in QSO. Also see last para.

The main confusion seems to be the inability to collapse the pulldown without selecting the text box and then not being able to clear the selection from the text box. I just found that right clicking in free space in the pull down list clears and collapses it.
Left clicking the down arrow opens the list but clicking it again closes the list but highlights the box. It should just close it.

There is a strange issue with the callsign-pulldown which may be related to its other quirks.
If there are no callsigns listed and the text box is right clicked, a greyed version of the reporter message context menu can sometimes appear, identified by the 'insert emoji' option!

Screenshot_20260128_130142

It can also happen to the frequency selection pull down (not greyed) where it will write an emoji into the frequency on view.

Is this supposed to work with both Record and logging? I tested both in succession and it auto-filled both with a different callsign. One from each selected source, I expected them both to fill from the last selected.

tmiw:
> The main confusion seems to be the inability to collapse the pulldown without selecting the text box and then not being able to clear the selection from the text box. I just found that right clicking in free space in the pull down list clears and collapses it. Left clicking the down arrow opens the list but clicking it again closes the list but highlights the box. It should just close it.

There's now logic that should hopefully deselect on close if the user didn't select a callsign.

> There is a strange issue with the callsign-pulldown which may be related to its other quirks. If there are no callsigns listed and the text box is right clicked, a greyed version of the reporter message context menu can sometimes appear, identified by the 'insert emoji' option!

This might be something DE specific :(

> Is this supposed to work with both Record and logging? I tested both in succession and it auto-filled both with a different callsign. One from each selected source, I expected them both to fill from the last selected.

I wasn't intending on it originally but I added similar logic for Record as of the latest commit.

barjac:
Great! - Looking good now for both logging and recording.
Only done minimal testing with two other stations in the pull-down but everything appeared to work as expected with no stray selections or flakiness \o/ I knew you could do it! Thanks :)

tmiw:
> Great! - Looking good now for both logging and recording. Only done minimal testing with two other stations in the pull-down but everything appeared to work as expected with no stray selections or flakiness \o/ I knew you could do it! Thanks :)

No worries, will merge up in a bit and prepare for the official 2.2.0 release.

PR #1206, opened 2026-01-31
macOS: Add /Applications shortcut to generated DMG.

### LINKED ISSUE #946: [Feature Request] macOS disk image needs an alias to the application directory for easier install
### Contact Details

_No response_

### Feature for this project

The macOS disk image opens up with just the application bundle. This makes it a bit hard to install as the user must open a new Finder window and get the Application directory in to view so we can drag and drop. Many applications include an alias to the Application directory so the user can just drag to install. Also, many applications have simple instructions in this window. Attached is  an example.

Image

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Non-standard compiler

None

### Implementation assistance

I think this needs to be done when creating the disk image. I can't see a script for this so I guess it's manual.

### Support assistance

Testing

tmiw:
Resolves #946 by generating the FreeDV DMG image through [dmgbuild](https://dmgbuild.readthedocs.io/en/latest/). This adds a shortcut to /Applications in the image to make it easier for users to install.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -884,6 +884,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * FlexRadio: Fix command format for UDP port. (PR #1205; thanks @amcdermond!)
 2. Build system:
     * macOS: Fix dylibbundler call for compilation. (PR #1204)
+    * macOS: Add /Applications shortcut to generated DMG. (PR #1206)
 
 ## V2.2.0 January 2026
 


tmiw:
@peterbmarks, how does the generated DMG look for you? (Should be https://github.com/drowe67/freedv-gui/actions/runs/21541851048?pr=1206 once GitHub finishes building it.) On Tahoe for me, there's no arrow as expected but I do see the /Applications shortcut.

tmiw:
Weird, I downloaded the generated DMG onto my M1 Mac Mini and it seems to render okay-ish there:

Screenshot 2026-01-31 at 11 15 53 AM

At least the arrow shows up now? The /Applications shortcut doesn't have an icon there though.

tmiw:
Same generated DMG downloaded to my Intel MBP:

image

Interestingly, both systems are runnning Tahoe.

tmiw:
Anyway, I had to shut down/restart my M1 Mac due to a power outage and I'm now getting the same behavior as with my Intel laptop. I *think* we're good here.

PR #1212, opened 2026-02-03
FreeDV Reporter: Use ItemsAdded/ItemsDeleted instead of Cleared() for performance.

tmiw:
Instead of clearing all rows from the table every time someone connects or disconnects, we use logic to add and remove individual rows.

Note that this wasn't previously done due to intermittent GTK-related crashes on Linux. We can do this now because the add/remove operations have to be done while the rows are in the backing store.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -875,6 +875,13 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
 
 # Release Notes
 
+## V2.3.0 TBD 2026
+
+1. Bugfixes:
+    * TBD
+2. Enhancements:
+    * FreeDV Reporter: Use ItemsAdded/ItemsDeleted instead of Cleared() for performance. (PR #1212)
+
 ## V2.2.1 February 2026
 
 1. Bugfixes:


barjac:
Seems stable and no regressions noticed, been in use for a few days now.

tmiw:
This might go to 2.3 instead of 2.2.1 given the scope of the changes. Holding off on merging for now.

@barjac, have you noticed any reduced CPU consumption due to this PR?

Tyrbiter:
I am now running this build, will report back if I see any problems.

barjac:
> This might go to 2.3 instead of 2.2.1 given the scope of the changes. Holding off on merging for now.
> 
> @barjac, have you noticed any reduced CPU consumption due to this PR?

Not sure but it's certainly not bad at around 30% of a core on RX with decode in operation and around 12% on TX for the FreeDV-GUI process.

This is with #5914353 which seems fine now.

tmiw:
> > This might go to 2.3 instead of 2.2.1 given the scope of the changes. Holding off on merging for now.
> > @barjac, have you noticed any reduced CPU consumption due to this PR?
> 
> Not sure but it's certainly not bad at around 30% of a core on RX with decode in operation and around 12% on TX for the FreeDV-GUI process.
> 
> This is with #5914353 which seems fine now.

For me, I don't see a CPU difference on Linux vs. master (when the modem isn't running), but I think I do see reduced CPU usage on macOS. We'll go ahead and merge for now and continue monitoring.

Issue #1215, opened 2026-02-05
[Software Issue] v2.2.0 digital audio stream (Frm Radio) is like morse code / not useable


kolbma:
### Platform

Linux

### Platform Version

openSUSE Leap 16.0

### FreeDV Version

v2.2.0

### Steps to Reproduce

Start freedv v2.2.0, start Modem, switch to Frm Radio tab and the stream is stammering like morse code. Only short time span audio, then a pause, audio, pause, aso.  
Also the stream buffer is not emptied. There is lag growing because of the pauses.  
When switching to Analog everything seems normal.  
Using v2.1.0 there is no problem.

### Expected Behavior

Working like in v2.1.0

### Actual Behavior

Not useable.

### Additional Comments

Not sure if and what information I could additionally provide.

kolbma:
Ah, and the closing of the v2.2.0 app is also very slow, often the window manager is asking to kill the app.  
Doesn't happen in v2.1.0.

tmiw:
Is this using the AppImage or by compiling yourself (using @barjac's script or otherwise)? Also, are you hearing any stuttering or other audio issues or is this just regarding the Frm Radio plot?

kolbma:
Yes both versions AppImages. I don't hear anything because the "stuttering" audio in digi mode can't be decoded. In analog switch and tabs it looks and sounds ok. But in the meantime I'm not sure if this has not also to do with the distro update and introducing pipewire and pipewire-pulseaudio compatibility layer. Although the v2.1.0 doesn't have this visual stuttering with the pipewire setup, I have detected some more problems with resampling. I try to install not officially any longer provided pulseaudio to replace pipewire tomorrow and will see if it works again with this. I'll give you an update.

tmiw:
Also, I'm working on releasing 2.2.1 right now. Give that a try and let me know if that works any better as I did fix an audio related locking issue I discovered while running unit tests.

kolbma:
Ok, sadly, no real change for my problem in v2.2.1.

## This is how it looks like in v2.1.0 and is working...

<img width="847" height="792" alt="Image" src="https://github.com/user-attachments/assets/e8bcc8f5-7feb-47e4-ac54-bd2eefcb9555" />
<img width="847" height="792" alt="Image" src="https://github.com/user-attachments/assets/438cdc21-6962-45be-9ca0-62bfebe230d1" />

## And here with the "holes" in v2.2.[01]...  

[...]

Issue #1210, opened 2026-02-01
[Software Issue] Intermittent occasional TX audio dropouts


g4dya:
### Platform

Linux

### Platform Version

Linux Mint 22.3 MATE

### FreeDV Version

2.2.1-dev-4902

### Steps to Reproduce

While transmitting, I occasionally see momentary dropouts in the RADE-encoded audio output from FreeDV-gui. This seems to occur at random, and when the dropouts occur there are typically several short dropouts over a period of about 10 or 20 seconds and then it will return to normal.

On my installation the audio is routed via a pipewire null sink to PiHPSDR. I have seen the dropouts occur on both the RF output of piHPSDR and on an audio spectrum analyser connected directly (by pipewire) to the FreeDV-gui TX output.

The presence of these dropouts is confirmed by other stations. The amplitude drops to zero - it's not just a change in level.

A recording of the QSO in RADE-encoded form shows no signs of the dropouts. It looks and sounds completely normal when played back through FreeDV-gui.

### Expected Behavior

No dropouts.

### Actual Behavior

As above.

### Additional Comments

This is not new. I have seen this for months.

Tyrbiter:
Would be worth detailing the pipewire and wireplumber package versions you have installed for completeness.

tmiw:
How is the recording being done? "Record" button in FreeDV or are you using a separate application to do the recording?

Also, are you able to try the AppImage instead of compiling yourself to see if there's a difference? The AppImage build we release is compiled at -O3 with both link-time and profile-guided optimization, which should use less CPU than e.g. a debug build without the latter two.

BTW this is likely a similar issue to the one that's causing test failures on GitHub. I can say that it's gotten a lot better over the last year or so, but still not 100%. Do you have any indication on your end whether the dropouts have gotten better across releases?

g4dya:
$ apt list -i pipewire wireplumber
Listing... Done
pipewire/noble-updates,now 1.0.5-1ubuntu3.2 amd64 [installed]
wireplumber/noble-updates,now 0.4.17-1ubuntu4.1 amd64 [installed]

I did the recording using the Record button in FreeDV.

I'm reluctant to use an Appimage build out of a vague fear it might mess stuff up. The CPU is not stretched - it's an i7 with eight 3.6 GHz cores.

I have not noticed the dropouts get any better or worse across recent releases - they seem much the same as when I first observed them.

Tyrbiter:
[...]

PR #1220, opened 2026-02-10
Linux: fix semaphore wakeup timing to reduce audio dropouts.

tmiw:
This PR contains additional changes to pipewire handling to fix timing issues discovered during testing, which appear to be contributing to reports of audio dropouts on Linux (see #1215 and #1210).

### DIFF
--- a/src/pipeline/RecordStep.cpp
+++ b/src/pipeline/RecordStep.cpp
@@ -41,11 +41,9 @@ RecordStep::RecordStep(
     : inputSampleRate_(inputSampleRate)
     , getSndFileFn_(std::move(getSndFileFn))
     , isFileCompleteFn_(std::move(isFileCompleteFn))
+    , inputFifo_(inputSampleRate_)
     , fileIoThreadEnding_(false)
 {
-    inputFifo_ = codec2_fifo_create(inputSampleRate_);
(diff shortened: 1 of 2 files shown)


Tyrbiter:
Adding this into master plus PR 1217 I see a noticeable CPU usage increase, but perhaps that is inevitable.

tmiw:
> Adding this into master plus PR 1217 I see a noticeable CPU usage increase, but perhaps that is inevitable.

Just looked and I'm getting a peak of 18% CPU during TX and ~30% during RX (as measured by `top` using the `rade_loss` test). That doesn't sound *terrible* but maybe it depends on the system being tested.

Tyrbiter:
btop says between 39% and 55% on 1 core of my i7-9900 when in per-core mode. That's without a signal being demodulated but the modem is running.

tmiw:
> btop says between 39% and 55% on 1 core of my i7-9900 when in per-core mode. That's without a signal being demodulated but the modem is running.

Oh, I was testing while demodulating a signal. It's expected for CPU usage to be higher when there's no signal (the experimental RADEV1 branch the Flex waveform uses, though, significantly optimizes the non-decoding case to the point where that *does* take less CPU than actually decoding a signal).

PR #1231, opened 2026-02-15
FreeDV Reporter: Disable autosizing enable/disable on Linux.

### LINKED ISSUE #1225: Reporter column widths no longer saved between sessions
Seeing this in master @ #bf851832

tmiw:
Partially resolves #1225 by disabling the logic added for 2.2 to disable and reenable column autosizing on Linux. Per testing this still seems to work properly on Windows and macOS so it'll remain enabled on those platforms.


Tyrbiter:
I noticed this morning that the station heard list in the main window also had truncated information in the frequency and SNR columns, is this supposed to autosize? I don't think I have seen this happen before.

tmiw:
> I noticed this morning that the station heard list in the main window also had truncated information in the frequency and SNR columns, is this supposed to autosize? I don't think I have seen this happen before.

Yeah, they're supposed to autosize as of the latest release. A screenshot might be helpful here so I can see what's better going on?

Tyrbiter:
I will see if it happens again, it's possible something else has caused it. I am using a local build of hamlib-4.7.0, could that have anything to do with this?

I've actually rebuilt freedv-gui with the updated hamlib devel packages installed and am now running https://github.com/drowe67/freedv-gui/pull/1231/commits/e7347116f985477fd2124a09451f13e3c469613b again so if I see this again I will report back here.

Tyrbiter:
OK, so the truncation is the actual column headings.
Screenshot From 2026-02-20 12-06-01

That may be easier to fix, or even change the heading to be narrower if that's the best choice. I see in the reporter window it is called MHz rather than Frequency.

tmiw:
> OK, so the truncation is the actual column headings.

Hopefully [63955c7](https://github.com/drowe67/freedv-gui/pull/1231/commits/63955c7545c0e8e683a421b5851cefc3579ce68f) is better in that regard?

EDIT: new commit.

Tyrbiter:
Yes https://github.com/drowe67/freedv-gui/pull/1231/commits/63955c7545c0e8e683a421b5851cefc3579ce68f seems to have fixed it, at least on F43.

tmiw:
I'm going to go ahead and merge for now, we can always do a followup PR for any remaining Windows issues.

PR #1233, opened 2026-02-16
Hamlib: Reenable FLRig/rigctl timeouts.

tmiw:
Partially reverts #1195 by reenabling timeouts for FLRig and rigctl. This is due to issues with stopping FreeDV as a result of the Hamlib request queue receiving an infinite number of frequency, etc. updates when the radio isn't able to respond quickly enough.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -884,6 +884,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Easy Setup: Fix crash while displaying Hamlib error. (PR #1224)
     * Force numpy to 2.3.x due to illegal instruction errors. (PR #1230)
     * FreeDV Reporter: Disable autosizing enable/disable on Linux. (PR #1231)
+    * Hamlib: Reenable FLRig/rigctl timeouts. (PR #1233)
 2. Enhancements:
     * FreeDV Reporter: Use ItemsAdded/ItemsDeleted instead of Cleared() for performance. (PR #1212)
     * Optimize "From XXX" plot performance. (PR #1238)


barjac:
Tested with ms-enable-flrig-timeout @ c311b2f6 although comments below are not related to this PR directly and have applied for a long time. This just prompted me ;)

Is there any need for such a long time-out?
If the modem is started with the radio accidentally off, it is possible to hit the PTT and apparently transmit.
Or if the modem is stopped, all buttons in the control dialogue are greyed which looks like a program hang.
Killing the program does work and then after it has apparently closed the IO error pops up some time later!
All the above is rather disconcerting.
If the time-out was maybe 2 or 3 seconds rather than the ~20 it seems to be now it would be more useful as at present the user is left in limbo wondering what is wrong.

tmiw:
> Is there any need for such a long time-out? If the modem is started with the radio accidentally off, it is possible to hit the PTT and apparently transmit. Or if the modem is stopped, all buttons in the control dialogue are greyed which looks like a program hang. Killing the program does work and then after it has apparently closed the IO error pops up some time later! All the above is rather disconcerting. If the time-out was maybe 2 or 3 seconds rather than the ~20 it seems to be now it would be more useful as at present the user is left in limbo wondering what is wrong.

Before this PR the timeout was 625ms (unless using rigctl or FLRig). Now it's 625ms regardless.

Unfortunately, someone had CAT control problems with the old timeout of 500ms, hence the original PR that bumped up the timeout. We definitely can't go above 1 second though due to the fact that we're requesting frequency/mode once a second.

barjac:
So why does it take more than 20 seconds for the warning dialogue to appear?

tmiw:
> So why does it take more than 20 seconds for the warning dialogue to appear?

The updated timeout only gets put into effect once the connection to the radio's actually made. I forget why this was done vs. having the timeout apply from the start (or if there was even a need to do this), but that should be easy to update.

barjac:
@ #af0dc81
On starting freedv and modem with rigctld daemon running and the radio off, I am seeing at least 7 hamlib communication timeout messages in the freedv log without any FreeDV warning dialog appearing. There may be more as I stopped freedv early to look at the log.
It was only after freedv had completely stopped that the warning popped up.

EDIT: I just timed the above leaving freedv running and it was 38 seconds before the warning dialogue appeared in FreeDV.

This is the full terminal output for a freedv session as above.
[hamlib_timeout.txt](https://github.com/user-attachments/files/25390763/hamlib_timeout.txt)

tmiw:
There's logic to suppress the error message for frequency/mode requests (otherwise users would be getting a popup once a second until they pushed Stop). However, it shouldn't be taking that long to get a popup once you try to push PTT or Stop (which will implicitly set PTT to Off). Try [e383271](https://github.com/drowe67/freedv-gui/pull/1233/commits/e38327111edaeb6e369d7546a082c0386d46fc9d)? This should have some additional logging to show the timeout being used.

barjac:
Using this PR patch on top of 'ms-tog-btn-reliabilty' the Error dialogue does not appear until after the modem stop button is pressed.

I turned off the radio and initially got repeated 'Communication timeout' messages interspersed with the usual reporter connect/disconnects going on.

These 'Communication timeout' messages eventually (a minute or so) became replaced with:
```
IO error
*1:rig.c(3636):rig_set_ptt entered
rig_lock: client lock engaged
*rig.c(3672) trace
netrigctl_set_ptt called vfo=currVFO, ptt=0, ptt_type=1
netrigctl_vfostr: called vfo=VFOA
netrigctl_vfostr: vfo_opt=0
netrigctl_set_ptt: cmd=T 0
netrigctl_transaction: called len=4
network_flush called
write_block():1134 failed -1 - Broken pipe
*1:rig_set_ptt: elapsed=0ms
rig_lock: client lock disengaged
*1:rig.c(3679):rig_set_ptt returning(-6) IO error

IO error
```

These continued to be repeated until I stopped the modem whereupon the Error popup dialogue appeared.

tmiw:
Yeah, I wouldn't expect the most recent to actually modify the behavior. Did you see any lines starting with "Current rig timeout:" or "Setting rig timeout to" in the output?

barjac:
> Yeah, I wouldn't expect the most recent to actually modify the behavior. Did you see any lines starting with "Current rig timeout:" or "Setting rig timeout to" in the output?

No, nothing like that. Starting the program and modem with the radio off does show the message dialogue after around 15 seconds, but if the radio is lost during a session it goes on for ever with IO error messages in the terminal repeating.

I will rebuild just this PR to be sure that the other one is not affecting it.

barjac:
Here is full output from start to finish, including closing the error dialog, stopping  modem and closing FreeDV. You can ignore all my audio links set-up/shut down etc.
Radio was off the whole time.

[start_without_radio.txt](https://github.com/user-attachments/files/25459423/start_without_radio.txt)

Oops I missed it:
```
baz@leno:~/BUGS/freedv/ $ cat start_without_radio.txt | grep "Current rig timeout"
16:13:24 [168] INFO /home/baz/freedv-rade/freedv-gui/src/rig_control/HamlibRigController.cpp:415: Current rig timeout: 0 ms    <<<<<<<<<
baz@leno:~/BUGS/freedv/ $ cat start_without_radio.txt | grep "Setting rig timeout"
baz@leno:~/BUGS/freedv/ $
```

tmiw:
I added some logic in [cc15bfa](https://github.com/drowe67/freedv-gui/pull/1233/commits/cc15bfa253feec44e81178d49327926a4ec7a956) to allow the error popup to occur in the case where the radio is unplugged/turned off during operation. Hopefully that helps?

barjac:
Nope still the same @ cc15bfa 
With FreeDV modem running in RX, if I turn off the radio I get repeated Communication timeouts, which then change to IO errors which continue for ever until the modem is manually stopped.
I then get the FreeDV error dialogue.

The error dialogue does appear if the radio is off at FreeDV program start, however it only happens about 30 seconds after starting the modem.

tmiw:
I think it might have been something rigctl related; I just tested with nDAX/nCAT (which are programs that allow you to use Flex radios in Linux and provide a rigctl interface for CAT control) and saw the same behavior you did. [d52b6e7](https://github.com/drowe67/freedv-gui/pull/1233/commits/d52b6e7674a45e2969ed15b6ee8e7a24b7e7e041) allows the error box to pop up within a second or two for me once I Ctrl-C the nCAT process, so hopefully that works better for you as well.

PR #1236, opened 2026-02-21
Improve reliability of Start/Stop and PTT buttons.

### LINKED ISSUE #1229: PTT button often fails to respond to a 'click' as opposed to 'press, hold, release'
Regularly now I am seeing the PTT button in the GUI failing to respond to a short normal 'click'.

It requires a press, hold and release which is not intuitive.

I have mentioned this before and I hear others having the same issue.

tmiw:
Resolves #1229 by using different indications of on/off and PTT state for determining whether the button actions should proceed.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -254,6 +254,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-24.04-arm, ubuntu-24.04]
+        fedora-version: [42, 43]
 
     runs-on: ${{ matrix.os }}
     needs: [build-appimage]
@@ -274,7 +275,7 @@ jobs:
     - name: Start Fedora container
       shell: bash
       run: |
-          sudo podman run -d --name fedora-test --systemd=always --privileged -v ${{github.workspace}}:/workspace docker.io/jrei/systemd-fedora:42
+          sudo podman run -d --name fedora-test --systemd=always --privileged -v ${{github.workspace}}:/workspace docker.io/jrei/systemd-fedora:${{ matrix.fedora-version }}
 
     - name: Install rtkit for RT threading
       shell: bash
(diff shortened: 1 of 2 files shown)


barjac:
Using #bb2c681 I cannot reproduce this issue, however neither could I in master earlier.
Its one of those issues that does not happen on demand, however this seems no worse at present.
I will continue to run this branch and see how it goes.

barjac:
This branch has PTT misfires on both TX on and off. @ #8218d24

Been testing it with short overs tonight between two of us for about 3/4hr and we both had regular fails :(

I could not decide whether it was some interaction between the tooltip and the button event that was blocking it or not.

/OT
We noticed also that with just the two of us that the pull-down list time remains at the 'start of QSO' time, but the SNR is updated every over. It begins to tell lies as time goes on. Is that intentional? :)

tmiw:
Do you have any feeling on whether they're more or less common with this PR vs. master?

barjac:
Good question. We can repeat a similar session maybe tonight using master for comparison.

tmiw:
> Good question. We can repeat a similar session maybe tonight using master for comparison.

Sounds good. I also added some logic to force tooltips to close when the main window becomes active (as a test).

barjac:
No real news yet, maybe tonight, however I did spend a long time with both master and this branch (still with the tooltips) on my own trying to reproduce the issue and totally failed. I'm wondering if it is only happening with several applications running or maybe after the machine has been running for a long time. Our skeds are late in the evening so machine has normally been on all day. Also time is probably late afternoon or early evening in the USA when there is a lot of USA internet traffic so maybe reporter delays/retries could be involved. Just grasping at straws!

tmiw:
> No real news yet, maybe tonight, however I did spend a long time with both master and this branch (still with the tooltips) on my own trying to reproduce the issue and totally failed. I'm wondering if it is only happening with several applications running or maybe after the machine has been running for a long time. Our skeds are late in the evening so machine has normally been on all day. Also time is probably late afternoon or early evening in the USA when there is a lot of USA internet traffic so maybe reporter delays/retries could be involved. Just grasping at straws!

If it's something with reporter traffic, I can maybe do something in this PR where it only updates a maximum number of rows every time the update timer fires (~150ms IIRC but I could be mistaken). Let's see if the latest in this PR helps at all before doing that, though.

barjac:
We just spent an hour running #0ca06 and we can't reproduce at all now.
However we also ran master for a while and could not reproduce with that either so it's not really conclusive.
We will stick with #0ca06 for a day or so.

tmiw:
wxWidgets was recently updated to 3.3.2. I'm not sure if that'll impact this issue at all but it might be worth trying the latest AppImages from this PR and/or master too.

barjac:
We can no longer reproduce this issue despite switching back to versions where it was quite reproducible, so I can only assume that it has been fixed by other updates.
Master is currently working well as was this branch when last tested after the latest changes.
I have not tested any appimages.
More testing here seems pointless without being able to reproduce the original issue so I would suggest that maybe this could be merged now if if you feel that the changes here have a potential theoretical benefit. AFAICT there are no regressions.

tmiw:
> We can no longer reproduce this issue despite switching back to versions where it was quite reproducible, so I can only assume that it has been fixed by other updates. Master is currently working well as was this branch when last tested after the latest changes. I have not tested any appimages. More testing here seems pointless without being able to reproduce the original issue so I would suggest that maybe this could be merged now if if you feel that the changes here have a potential theoretical benefit. AFAICT there are no regressions.

Sounds good.

PR #1238, opened 2026-02-22
Optimize PlotScalar based plots

tmiw:
Adds optimizations to `PlotScalar` based plots (i.e. "From XXX"). This results in a ~40% processor reduction during the rendering of these plots (as measured by `perf record` while running the `rade_loss` test).

PR #1239, opened 2026-02-24
Fix issue with scalar plot hangs when switching from other plots.

tmiw:
Followup to #1238 due to issues reported with hangs while switching from spectrum or waterfall plots to a scalar plot.


Tyrbiter:
Seems a lot better so far, will report if I see anything unexpected.

Tyrbiter:
Everything working normally now.

PR #1246, opened 2026-03-11
Hamlib/OmniRig: Ignore frequency/mode query during TX.

tmiw:
Updates Hamlib and OmniRig implementations to ignore frequency/mode queries during TX. This is due to a [report](https://groups.google.com/g/digitalvoice/c/dz4aUOD7JQw/m/dVZWBtIXAAAJ?utm_medium=email&utm_source=footer) that a user's radio is introducing noise into the RADE transmission as a result of the SDR program in use briefly stopping TX itself to handle the query.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -886,6 +886,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * FreeDV Reporter: Disable autosizing enable/disable on Linux. (PR #1231)
     * Improve reliability of Start/Stop and PTT buttons. (PR #1236)
     * Hamlib: Reenable FLRig/rigctl timeouts. (PR #1233)
+    * Hamlib/OmniRig: Ignore frequency/mode query during TX. (PR #1246)
 2. Enhancements:
     * FreeDV Reporter: Use ItemsAdded/ItemsDeleted instead of Cleared() for performance. (PR #1212)
     * Optimize "From XXX" plot performance. (PR #1238, #1239)


Tyrbiter:
Building on Fedora 44 beta, it looks like there is another tightening of the compiler warning/error screw:

[ 98%] Building CXX object src/CMakeFiles/freedv.dir/ongui.cpp.o
/home/bdm/git/freedv-gui-ms-hamlib-ignore-freq-tx/src/sox_biquad.c: In function ‘sox_biquad_filter’:
/home/bdm/git/freedv-gui-ms-hamlib-ignore-freq-tx/src/sox_biquad.c:131:18: error: variable ‘clips’ set but not used [-Werror=unused-but-set-variable=]
  131 |     unsigned int clips;
      |                  ^~~~~
At top level:
cc1: note: unrecognized command-line option ‘-Wno-null-pointer-subtraction’ may have been intended to silence earlier diagnostics
cc1: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/freedv.dir/build.make:107: src/CMakeFiles/freedv.dir/sox_biquad.c.o] Error 1

I'm assuming that there needs to be some sort of initialisation of the clips value? No, it's initialised to 0 but then appears to always be 0 within the next few lines. Is that what it means by unused?

tmiw:
> I'm assuming that there needs to be some sort of initialisation of the clips value? No, it's initialised to 0 but then appears to always be 0 within the next few lines. Is that what it means by unused?

The code is writing to it but not reading from it, but I don't know if it really can do the latter. Does [357b831](https://github.com/drowe67/freedv-gui/pull/1246/commits/357b8314fa42617932d2f39db835cff7e6a58a4f) suppress that warning at least?

Tyrbiter:
Well it certainly now builds to completion with https://github.com/drowe67/freedv-gui/pull/1246/commits/357b8314fa42617932d2f39db835cff7e6a58a4f applied

More after I try running it. It starts and seems to run normally.

I did search using the error message and I see this isn't actually new, lots of people complaining about it for some years. I presume this is a change in the latest gcc version used in Fedora 44 as they have probably set system-wide compiler options. Casting to void seems to be the accepted fix.

tmiw:
I wonder if we should disable the frequency drop-down during TX altogether. Frequency changes are still possible at the moment, it just stops TX, switches frequency and then restarts TX. Would make things more consistent I think.

Tyrbiter:
I can't think of why you would want to change frequency during Tx, it would leave a QSO partner in limbo.

tmiw:
> I can't think of why you would want to change frequency during Tx, it would leave a QSO partner in limbo.

Maybe if one accidentally is a bit off frequency during the initial CQ call, but I don't think this comes up often enough to justify keeping it enabled. Checked in [6db55d8](https://github.com/drowe67/freedv-gui/pull/1246/commits/6db55d8dba3ceb7b9c7593a98e68e426f23ae749) to disable the drop-down during TX.

barjac:
This built without any errors at 4c5ef44 and been using it since then without any issues.

Issue #1172, opened 2026-01-10
[Feature Request] Add SNR plot to main window


faustdownunder:
### Contact Details

dpelz@rfcurrent.com vk3ffb, rgatchell@gmail.com k5rwg, nc3z.gary@gmail.com nc3z

### Feature for this project

Before getting back to FreeDV last year after a pause of about 2 years, I used VarAC (Vara modem) frequently. I found the SNR graph feature in VarAC very useful. The instantaneous SNR data is of course of prime importance but it does not provide us with trend information, QSB pattern changes, diagnostics, etc.. Sure, this can be seen as a feature for a minority of users but in VarAC it became quite popular among users. 
The feature would also have a bearing on the improvement of FreeDV insofar as it would reveal propagation path and ionosphere related data (see also DREAM DRM software features). The changes in local noise level and type and the effect on the SNR can probably also be studied. 
Needless to mention that the SNR graph feature would also improve comparison of different antennas. Many hams have for example both horizontal dipoles and vertical antennas for the popular 40m band. 

Thanks in advance

PS  See also [#852](https://github.com/drowe67/freedv-gui/issues/852)  Except that I would not call the feature "cosmetic".

<img width="485" height="294" alt="Image" src="https://github.com/user-attachments/assets/e150e7ef-178f-4b55-aa3a-a2d870111783" />

<img width="485" height="294" alt="Image" src="https://github.com/user-attachments/assets/d366f778-7548-4089-99ce-8b3e19730971" />

Note: Such graphs would of course look somewhat different in FreeDV because of the larger number of short-term SNR data samples.

<img width="706" height="584" alt="Image" src="https://github.com/user-attachments/assets/1dbe0cd1-52d7-4505-b0be-3bed937d6386" />

### VHF/UHF

No

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

Yes

### Implementation assistance

I can help in documenting, creating math algorithms perhaps, testing, supporting. 

### Support assistance

Dealing with related user requests, documenting changes.

tmiw:
This sounds similar to https://github.com/drowe67/freedv-gui/issues/852. Can you confirm?

faustdownunder:
Hi Mooneer,
The answer is a clear "Yes". I would make the graph plot "start-stop" user controllable. The word "cosmetic" is in this context not descriptive or appropriate.

faustdownunder:
I am now able to provide code programming support, if need be.

[...]

PR #1250, opened 2026-03-15
Add SNR plot to main window.

### LINKED ISSUE #852: [Feature Request] Cosmetic: 1 to 5 minute SNR graph
### Contact Details

_No response_

### Feature for this project

May be useful to track path availability, similar resource was found on other softwares such as DreAM DRM or SatDump.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Non-standard compiler

None

### Implementation assistance

Can test from 160 to 6 meter, setup working smooth under Windows 10 with all HF spectrum coverage, CAT with no issues on a Elecraft K3S both via Hamlib or OmniRig. Local KiWi available for the public as well. 

### Support assistance

Not a software guy. Just a DXer / phone cw rtty contester.

tmiw:
Resolves #852 and #1172 by adding an "SNR" tab to the main window with a 60 second SNR graph. Example:

Screenshot 2026-03-15 at 3 08 44 PM

TODO:

- [x] Plot seems to be scrolling a bit too quickly--investigate and resolve.
- [x] Plot seems to improperly clear itself (at least on macOS) when clicking around in the main window.
- [x] Reverse X axis (i.e. should go from -60 to 0, not 0 to 60).

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -57,7 +57,7 @@
 
 // SNR plot constants
 #define NO_SNR_VAL (-10)
-#define MAX_SNR_VAL (35)
+#define MAX_SNR_VAL (40)
 #define SNR_PLOT_SECONDS (180)
 #define SNR_PLOT_SECOND_SEGMENTS (10)
 #define SNR_PLOT_DT (0.1)


barjac:
Looks good but we seem to have lost the 'From decoder' tab!

Also maybe a 3min X axis would be better.

Just noticed that the SNR graph clips at -5dB although the scale goes to -10.
i.e. Anything below -5 produces a flat line at -5.

Could the Y axis stop at +30dB rather than +40dB? This would help expand the Y axis a bit more when the main window is shallow.

Also crashed on program close
 
[Assert_f99a8.txt](https://github.com/user-attachments/files/26010811/Assert_f99a8.txt)

tmiw:
> Looks good but we seem to have lost the 'From decoder' tab! Also maybe a 3min X axis would be better.

That tab still shows up for me. Maybe try disabling experimental features, restarting FreeDV and then reenabling (to clear out the saved tab state)?

barjac:
> > Looks good but we seem to have lost the 'From decoder' tab! Also maybe a 3min X axis would be better.
> 
> That tab still shows up for me. Maybe try disabling experimental features, restarting FreeDV and then reenabling (to clear out the saved tab state)?

Yes after that it reset the default display and Frm Decoder returned. :)
I am seeing the exception on every program stop.

Flat bottoming at -5dB:- (spikes to -10dB are static spikes but true indicated -8dB shows as -5dB on the graph

image

tmiw:
Increased lookback to 3 minutes and hopefully fixed the flat line problem. Maybe the crash issue will fix itself too? Not sure about that one yet.

Re: SNR range--I've seen >30 in practice with extremely good RADE signals so not sure we want to cap at 30.

faustdownunder:
The smooth trace above looks good, as long as the data integrity is maintained (trace passes through all known SNR data points). I was going to suggest linear interpolation between any two SNR data points separated by a horizonal line section. In this way a horizontal trace section can never exceed the length of the minimum separation between two data points. 
The root cause of long horizontal line sections is of course the SNR level discretisation (number of SNR level points) used for the raw data.

A 1-minute time span is perhaps a bit small. Many overs in a QSO exceed one minute. I would suggest we either make it user selectable (1min ...5min... 10 .... 20min) or fixed at 5min. It would further be good if we change trace colour after each over (steep drop of SNR).

Retaining SNR data for later view or processing would be great but is not urgent.. I better stop here :)

barjac:
A few issues with this now using Plasma6 & Wayland under Mageia 10 (Cauldron):
The X timing is only about 30s for full widget width despite the scale.
The graph appears to be initially using short horizontal bars to draw the trace. However if the widget is re-drawn (select waterfall and go back to SNR) it then re-draws without the bars showing more detail.
Where the 'live' graph is almost a straight line (averaged?) it changes on refresh to a much more fluctuating line.

Before refresh:

Screenshot_20260316_170000


Different trace but after a refresh:

Screenshot_20260316_164833

Crash has gone with #16cc221

faustdownunder:
This is an attempt to explain the current SNR graph :

image

My suggestion is to leave out y_5 to y_9 and replace these points with a direct line from y_4 to y_10. Alternatively, the SNR calculation could perhaps be modified towards a finer resolution. Otherwise, the method suggested above or something similar to that can be used in order to smoothen the graph trace.

barjac:
It appears to me that the graph is currently ignoring the first decimal place, so higher precision is already available.

tmiw:
I checked in changes to have the graph render at the correct speed. Hopefully that helps with some of the rendering issues above.

BTW SNR isn't considered valid without sync, which is why the graph line drops off to -10 after an over.

barjac:
Yes! No issues now but will give it a better test tomorrow!

Screenshot_20260317_002803

The above was from playback of a short recording off-air.

faustdownunder:
A typical daytime 40m band high-angle radiation (NVIS) 45km long path:

image

The 60sec window is a known limitation.  
image

Here we see a superimposed linear SNR change.

tmiw:
> The 60sec window is a known limitation. 

The latest Windows build from this PR is at https://github.com/drowe67/freedv-gui/actions/runs/23170177226/artifacts/5954990761, which should get you an increased window. Let me know how that works for you.

g4dya:
A RADE signal recorded and replayed locally will max out the SNR at 34 dB, so 35 dB at the top would suffice.

If the number of seconds/division on the x-axis was a multiple of 10 it would look tidier. Making the numbers integers would provide another cosmetic tweak.

That said, it works well. Thanks for the new feature.

tmiw:
> A RADE signal recorded and replayed locally will max out the SNR at 34 dB, so 35 dB at the top would suffice.
> 
> If the number of seconds/division on the x-axis was a multiple of 10 it would look tidier. Making the numbers integers would provide another cosmetic tweak.
> 
> That said, it works well. Thanks for the new feature.

Just checked in some tweaks accordingly :)

barjac:
In discussions yesterday with two other users we came to the conclusion that the plot should only scroll during RX or during a transmission the plot slowly exits stage left!
So the plot would still scroll continuously in full duplex but not in TX for half duplex. 

The X time labels start to overlap and become meaningless when the main window size is reduced. Could these be reduced to maybe only showing every 30 secs?
Also if possible the Y axis 35dB and -10dB labels could be removed as they mis-align.

Screenshot_20260321_000537

g4dya:
On my display the 35dB, -10dB, -180s and -0s marks do not align with the graticule.
If they can't be aligned could they just be omitted?
If the graticule had six divisions on the x-axis, as it does on the RX spectrum tab, that would make 30s/div which would be tidier. 

Screenshot at 2026-03-21 10-28-07

tmiw:
See https://github.com/drowe67/freedv-gui/pull/1261. Note that the numbers at the very edges are supposed to be differently aligned to avoid having them be cut off for certain smaller window sizes.

PR #1259, opened 2026-03-19
Add Tune button to allow tuning antennas/rigs.

### LINKED ISSUE #1245: [Feature Request] Tune button for netting and ATU activation
### Contact Details

G6FLH, G6WPJ, G4MKT

### Feature for this project

This feature would be similar to that provided in wsjtx.
There are two main use cases:

1. Provide a low power tuning signal for adjusting manual or automatic antenna matching.

2. Provide an accurate 1500Hz (or maybe 1475Hz) carrier to net older transmitters.

Proposed GUI integration:
Use the existing 'TX Attenuation' slider to adjust the level by making it dual function like the Mic/Spkr slider.
Rename the 'TX Attenuation' slider widget to 'TX/Tune Attenuation'
Add a 'Tune' button to the TX/Tune Attenuation widget which has a red background and 'Stop' label when active. 
Save the current Tune level in the config on program close and restore it on start.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

Same as always 

### Support assistance

As always - testing mainly.

tmiw:
Resolves #1245 by adding Tune button that emits a 1500 Hz carrier to enable tuning of antennas/rigs.

Note: this currently uses the same TX Attenuation slider as regular TX, not sure yet if we need a specific setting for tune TX.

### DIFF
--- a/src/gui/dialogs/dlg_options.cpp
+++ b/src/gui/dialogs/dlg_options.cpp
@@ -519,6 +519,9 @@ OptionsDlg::OptionsDlg(wxWindow* parent, wxWindowID id, const wxString& title, c
     wxStaticBox *sb_modemstats = new wxStaticBox(m_modemTab, wxID_ANY, _("Modem Statistics"));
     wxStaticBoxSizer* sbSizer_modemstats = new wxStaticBoxSizer(sb_modemstats, wxVERTICAL);
 
+    m_showDecodeStats = new wxCheckBox(sb_modemstats, wxID_ANY, _("Show Decode Stats"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
+    sbSizer_modemstats->Add(m_showDecodeStats, 0, wxALL | wxALIGN_LEFT, 5); 
+    
     wxBoxSizer* sbSizer_statsResetTime = new wxBoxSizer(wxHORIZONTAL);
     wxStaticText *m_staticTextResetTime = new wxStaticText(sb_modemstats, wxID_ANY, _("Time before resetting stats (in seconds):"), wxDefaultPosition, wxDefaultSize, 0);
     sbSizer_statsResetTime->Add(m_staticTextResetTime, 0, wxALL | wxALIGN_CENTER_VERTICAL, 5);
@@ -672,8 +675,6 @@ OptionsDlg::OptionsDlg(wxWindow* parent, wxWindowID id, const wxString& title, c
     sbDebugOptionsSizer2->Add(m_ckboxTxRxDumpFifoState, 0, wxALL | wxALIGN_LEFT, 5);   
     m_ckboxFreeDVAPIVerbose = new wxCheckBox(sb_fifo2, wxID_ANY, _("APiVerbose"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
     sbDebugOptionsSizer2->Add(m_ckboxFreeDVAPIVerbose, 0, wxALL | wxALIGN_LEFT, 5);   
-    m_showDecodeStats = new wxCheckBox(sb_fifo2, wxID_ANY, _("Show Decode Stats"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
-    sbDebugOptionsSizer2->Add(m_showDecodeStats, 0, wxALL | wxALIGN_LEFT, 5); 
     
     m_experimentalFeatures = new wxCheckBox(sb_fifo2, wxID_ANY, _("Enable Experimental Features"), wxDefaultPosition, wxDefaultSize, wxCHK_2STATE);
     sbDebugOptionsSizer3->Add(m_experimentalFeatures, 0, wxALL | wxALIGN_LEFT, 5);   


barjac:
Yes its working fine - thanks and yes we do need to store the tune levels somehow.

I just had a long think about this and considering the various uses for this I think the following may work:

Use cases:-
1. Used as a tune signal for power setting the tune and TX level are the same.
2. Used as a low power auto tune signal the tune will be lower than the TX level.
3. Used as a netting signal the tune signal will be a lot lower than the TX signal.

Since there is a hope that 'per band TX attenuation' levels may become a reality then storing of the 'Tune' could be simply a relative level applied to the TX attenuation figure.
I would currently use it at 0dB ratio for case 1.
Users wanting a tuner activation signal may use -10dB for case 2.
For netting maybe -20db for case 3.

A user would simply set the TX level, then set the Tune level and the tune:TX level ratio (dB difference) would be stored in the config.

This would avoid over complication if/when introducing 'per band TX atten' since the stored values would not be doubled, as the same relative levels would be applied across all bands.

Minor bug: The 'Tune' button border is always highlighted in blue after one use, other buttons only do this when moused over. This highlight only clears when the modem is stopped.

Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?

tmiw:
Separate tune level added as of [d743552](https://github.com/drowe67/freedv-gui/pull/1259/commits/d74355282d6b04e66837b74cb3fd10b2ab58e5cc), as well as some other tweaks and bugfixes.

> Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?

The carrier is on 1500, but I thought RADE was also centered there. Is that not actually the case?

g4dya:
> > Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?
> 
> The carrier is on 1500, but I thought RADE was also centered there. Is that not actually the case?

AIUI RADE has 30 carriers each of which is a multiple of 50 Hz. The centre of the spectrum is halfway between the 15th and 16th carrier, i.e. 1475 Hz.

To get the spectrum centred on 1500 Hz there would need to be an odd number of carriers.

tmiw:
> > > Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?
> > 
> > 
> > The carrier is on 1500, but I thought RADE was also centered there. Is that not actually the case?
> 
> AIUI RADE has 30 carriers each of which is a multiple of 50 Hz. The centre of the spectrum is halfway between the 15th and 16th carrier, i.e. 1475 Hz.
> 
> To get the spectrum centred on 1500 Hz there would need to be an odd number of carriers.

Makes sense. I can do a quick update to have the carrier actually be on 1475 instead.

g4dya:
Just tried f5c0a67. Unfortunately it goes into tune when I push PTT, not just the tune button. Stopping and restarting the modem resets it.

g4dya:
Might it be better to use a 2-tone tune signal? This would (a) replicate RADE's approx. 3dB PAPR, (b) go easier on rigs than a single carrier and (c) assist operators to check linearity.

The two tones could be 1450 and 1500 Hz, each half the amplitude of the single tone case.

tmiw:
> Just tried [f5c0a67](https://github.com/drowe67/freedv-gui/commit/f5c0a678b087833d73a178869103c9931af96150). Unfortunately it goes into tune when I push PTT, not just the tune button. Stopping and restarting the modem resets it.

Thanks! [d4d633e](https://github.com/drowe67/freedv-gui/pull/1259/commits/d4d633e19529939c117676f338da9d81d009ec3b) should fix this.

> Might it be better to use a 2-tone tune signal? This would (a) replicate RADE's approx. 3dB PAPR, (b) go easier on rigs than a single carrier and (c) assist operators to check linearity.
> 
> The two tones could be 1450 and 1500 Hz, each half the amplitude of the single tone case.

One of the use cases is to allow radio/antenna tuning. Would this interfere with that?

g4dya:
> > Just tried [f5c0a67](https://github.com/drowe67/freedv-gui/commit/f5c0a678b087833d73a178869103c9931af96150). Unfortunately it goes into tune when I push PTT, not just the tune button. Stopping and restarting the modem resets it.
> 
> Thanks! [d4d633e](https://github.com/drowe67/freedv-gui/pull/1259/commits/d4d633e19529939c117676f338da9d81d009ec3b) should fix this.

It does. Thanks.
 
> > Might it be better to use a 2-tone tune signal? This would (a) replicate RADE's approx. 3dB PAPR, (b) go easier on rigs than a single carrier and (c) assist operators to check linearity.
> > The two tones could be 1450 and 1500 Hz, each half the amplitude of the single tone case.
> 
> One of the use cases is to allow radio/antenna tuning. Would this interfere with that?

Good question. Auto ATUs often require a single tone. I hadn't thought of that.

barjac:
> Separate tune level added as of [d743552](https://github.com/drowe67/freedv-gui/pull/1259/commits/d74355282d6b04e66837b74cb3fd10b2ab58e5cc), as well as some other tweaks and bugfixes.
> 
> > Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?
> 
> The carrier is on 1500, but I thought RADE was also centered there. Is that not actually the case?

Not according to the FreeDV terminal output I have seen where it reports the RX bandpass filter centre frequency. Maybe I misinterpreted that?

I just searched for the message but as always when you look for these things they are never to be found ;)

tmiw:
> > Separate tune level added as of [d743552](https://github.com/drowe67/freedv-gui/pull/1259/commits/d74355282d6b04e66837b74cb3fd10b2ab58e5cc), as well as some other tweaks and bugfixes.
> > > Is this actually 1500Hz or 1475Hz i.e. the centre of the FreeDV spectrum?
> > 
> > 
> > The carrier is on 1500, but I thought RADE was also centered there. Is that not actually the case?
> 
> Not according to the FreeDV terminal output I have seen where it reports the RX bandpass filter centre frequency. Maybe I misinterpreted that?
> 
> I just searched for the message but as always when you look for these things they are never to be found ;)

I think I know what you're talking about but I can't get it to show up in the FreeDV logs right now, either. We'll stick with 1475 for now unless testing with more users shows that it should be something different.

barjac:
Using this branch (ms-tune-button @ d4d633e1) I just noticed that the 'Stats' widget is back on screen with 'Legacy modes' unchecked in the Options->Modem settings.
I wondered where some screen real estate had gone!

tmiw:
> Using this branch (ms-tune-button @ [d4d633e](https://github.com/drowe67/freedv-gui/commit/d4d633e19529939c117676f338da9d81d009ec3b)) I just noticed that the 'Stats' widget is back on screen with 'Legacy modes' unchecked in the Options->Modem settings. I wondered where some screen real estate had gone!

Hmm, I just recompiled and I'm not seeing this on macOS or Linux (Ubuntu 24.04 with GNOME). Screenshot?

EDIT: I just realized that there's a "Show Decode Stats" option that causes that to appear for me. Maybe that's inadvertently enabled for you?

image

barjac:
> > Using this branch (ms-tune-button @ [d4d633e](https://github.com/drowe67/freedv-gui/commit/d4d633e19529939c117676f338da9d81d009ec3b)) I just noticed that the 'Stats' widget is back on screen with 'Legacy modes' unchecked in the Options->Modem settings. I wondered where some screen real estate had gone!
> 
> Hmm, I just recompiled and I'm not seeing this on macOS or Linux (Ubuntu 24.04 with GNOME). Screenshot?
> 
> EDIT: I just realized that there's a "Show Decode Stats" option that causes that to appear for me. aybe that's inadvertently enabled for you?
> 
> image
> 

Yes I had turned on all the debugging options trying to find that RX filter frequency. I had not  noticed that ' Decode stats' was mixed in with the debugging options, maybe it should be elsewhere?

tmiw:
> Yes I had turned on all the debugging options trying to find that RX filter frequency. I had not noticed that ' Decode stats' was mixed in with the debugging options, maybe it should be elsewhere?

What tab would work better in your opinion?

barjac:
> > Yes I had turned on all the debugging options trying to find that RX filter frequency. I had not noticed that ' Decode stats' was mixed in with the debugging options, maybe it should be elsewhere?
> 
> What tab would work better in your opinion?

Maybe where it is but in a new Misc. group with the Experimental features? Just not mixed in with all the logging debug settings where it is too easy to inadvertantly select.

tmiw:
> Maybe where it is but in a new Misc. group with the Experimental features? Just not mixed in with all the logging debug settings where it is too easy to inadvertantly select.

Looks like we already have something in the Modem tab so I went ahead and moved it there:

image

barjac:
Looks good!

g4dya:
The tone is 1475 Hz but the tooltip still says 1500 Hz.

tmiw:
> The tone is 1475 Hz but the tooltip still says 1500 Hz.

Actually, I think the tone is supposed to be 1500 Hz after all based on the code comment that I saw:

```
#define FDMDV_FCENTRE                                                          \
  1500 /* Centre frequency, Nc/2 carriers below this, Nc/2 carriers above (Hz) \
        */
```

Granted, this is from Codec2, but the logic in the comment seems to apply for RADE too (30 / 2 = 15 carriers on each side). I'll go apply this to one of the active PRs.

PR #1260, opened 2026-03-21
Add ability to hide self from FreeDV Reporter list.

### LINKED ISSUE #1186: Easy access reporting ON/OFF switch needed in main GUI
### Contact Details

G4MKT, G6WPJ, G8VLR, G3VHH, DM4IM

### Feature for this project

This crops up in conversation almost daily.
Not having this easily accessible is probably one of the major contributors to 'reporter clutter'.

A widget below 'Logging' labelled 'Reporting' with a check box labelled 'Enable' would be well worth the space required!

On many occasions users want to listen incognito to avoid being called when busy doing work in the shack without interruption.

Having to dig several layers down into the menu system to do this is really annoying even when you know where to look and very off-putting for those who rarely have cause to modify settings.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

As always

### Support assistance

As always

tmiw:
Resolves #1186 by adding a new button to the main window to allow one to hide themselves from the FreeDV Reporter list.

Example when not being hidden:

image

And when hidden:

image


barjac:
Not tested yet but 'Hide' implies secrecy, 'Show' implies showing the reporter which is confusing. 
What is wrong with the widget being called 'Reporting' with the button labels being 'Turn On' and 'Turn Off' to avoid any 'action' vs. 'state' confusion that we have seen before.

barjac:
Is this supposed to interact with 'Tools->Options->Reporting->Enable reporting' in any way?
Turning off the above 'Enable reporting' does not appear to disable the 'Hide' button.
If reporting is not enabled in 'Tools...' then this button should either be disabled, greyed, or not even displayed?
Alternatively this button could replace the 'Tools->Options->Reporting->Report to FreeDV Reporter' option altogether as it now duplicates the function and could be always displayed unless the Call and Locator are missing.

tmiw:
> Is this supposed to interact with 'Tools->Options->Reporting->Enable reporting' in any way? Turning off the above 'Enable reporting' does not appear to disable the 'Hide' button. If reporting is not enabled in 'Tools...' then this button should either be disabled, greyed, or not even displayed? Alternatively this button could replace the 'Tools->Options->Reporting->Report to FreeDV Reporter' option altogether as it now duplicates the function and could be always displayed unless the Call and Locator are missing.

It was just keying off of the FreeDV Reporter enable option. I added "Enable Reporting" as a prerequisite as well in [eb19054](https://github.com/drowe67/freedv-gui/pull/1260/commits/eb19054fbe5532dc783221d8672b5d24ad805ef8).

> Not tested yet but 'Hide' implies secrecy, 'Show' implies showing the reporter which is confusing. What is wrong with the widget being called 'Reporting' with the button labels being 'Turn On' and 'Turn Off' to avoid any 'action' vs. 'state' confusion that we have seen before.

How about Hide Self/Show Self?

barjac:
That to me would be ambiguous as it sounds as though you would only not see yourself rather than that you would stop reporting. We just want to stop reporting, so why not just have a Turn On/Off button in a widget called 'Reporting'  that does exactly what it says.

tmiw:
> That to me would be ambiguous as it sounds as though you would only not see yourself rather than that you would stop reporting. We just want to stop reporting, so why not just have a Turn On/Off button in a widget called 'Reporting' that does exactly what it says.

Label is now "Turn Off" / "Turn On".

Tyrbiter:
I have this and the updated PR #1261 changes added to a local build of master and it's all working nicely as far as I can tell.

barjac:
Err... no.
The widget title should be 'Reporting' not 'FreeDV reporter'.
This turns reporting on and off, it does not turn the FreeDV reporter on and off!

Works great though! Thanks :)

tmiw:
> Err... no. The widget title should be 'Reporting' not 'FreeDV reporter'. This turns reporting on and off, it does not turn the FreeDV reporter on and off!
> 
> Works great though! Thanks :)

If the title got changed, wouldn't that imply that PSK Reporter reporting's also disabled?

barjac:
> > Err... no. The widget title should be 'Reporting' not 'FreeDV reporter'. This turns reporting on and off, it does not turn the FreeDV reporter on and off!
> > Works great though! Thanks :)
> 
> If the title got changed, wouldn't that imply that PSK Reporter reporting's also disabled?

Maybe it should be...
OR if not, then what about "FDV reporting"? I can't think of anything shorter that has the right meaning.
OR Use "Reporting" and clarify in a tooltip, it's another of those things that once you know, you know.

Tyrbiter:
Personally I don't want this to control PSKreporter, just FreeDV Reporter. If I wanted that off as well I would do it in the settings dialog.

tmiw:
Maybe the better question is: would there be any confusion if someone sees a station on PSK Reporter but not FreeDV Reporter? I don't think there would be as the former in theory is supposed to be on [an up to 5 minute delay](https://pskreporter.info/pskdev.html), but I could be wrong.

Tyrbiter:
> Maybe the better question is: would there be any confusion if someone sees a station on PSK Reporter but not FreeDV Reporter? I don't think there would be as the former in theory is supposed to be on [an up to 5 minute delay](https://pskreporter.info/pskdev.html), but I could be wrong.

I don't think this would be an issue, PSKreporter is much less responsive than the FreeDV Reporter and doesn't have the ability to request a QSY either. I think there's enough configurability in the freedv-gui settings already.

barjac:
So "FDV reporting" then?

tmiw:
> So "FDV reporting" then?

Updated.

barjac:
In testing I have noticed that the FDV Reporting button status is not saved between runs of FreeDV.
I think it should be, since assuming that the default status for a new install will be on, then if it is turned off by the user then that user decision should not be silently reverted by the software.
If this is intentional to avoid inadvertently leaving the button in the off position then this could be mitigated by changing the button b/g colour to a pale amber as a warning wnen set off. It still should be saved between runs.

Tyrbiter:
Should this have latest master merged in, for all up testing?

tmiw:
> In testing I have noticed that the FDV Reporting button status is not saved between runs of FreeDV. I think it should be, since assuming that the default status for a new install will be on, then if it is turned off by the user then that user decision should not be silently reverted by the software.

Updated to preserve state across restarts.

> Should this have latest master merged in, for all up testing?

Done.

Tyrbiter:
@barjac suggested changing the colour of the button when inactive, but perhaps it could be coloured if the setting is different to that in the Options tab as a memory aid.

I notice that the button is inactive if the FreeDV Reporter setting in Options is off, so it only acts to turn reporting off.

Maybe a bit more thought as to whether we want it to be a full override to the Options setting. I'm not sure if there is an argument for that function, what do others think?

barjac:
With this in the main GUI I see no need for the  duplicate control in the settings. They will only conflict one way or anorher if both exist and cause user confusion.
Tbe one in settings is redundant.

Tyrbiter:
As it stands it works. If you turn the Reporter off in the Options tab it's effectively permanent, so maybe just leave it as is.

tmiw:
> As it stands it works. If you turn the Reporter off in the Options tab it's effectively permanent, so maybe just leave it as is.

I'm inclined to agree with this since the main window button is intended as a temporary toggle (i.e. reporting would be enabled most of the time unless someone happens to not be actively using FreeDV). Though if it's supposed to be a temporary toggle, wouldn't we want the state to *not* be preserved across restarts?

barjac:
I see your point, however if the object of having this feature is to partly reduce the number of stations who are really 'inactive' from 'always' showing in the reporter at program start (assuming they do sometimes want to report, so it is enabled in settings) then to have it always turn on at program start is counter productive.
I personally prefer to only report when I am actively prepared to participate in a OSO so would normally leave it off, without having to remember to turn it off every time I start the program, before starting the modem.

It is just perfect for me as it is now :)

We all work in different ways so maybe there should be a 'Remember FDV Reporting state' option somewhere that defaults to ON :)

Tyrbiter:
Actually I think this is exactly what we have with https://github.com/drowe67/freedv-gui/pull/1260/commits/0b1e8a037eef0d78ffba671d6d2d8da9325a8fac so my vote is go with this unless a torrent of complaints land with v2.3.0

PR #1261, opened 2026-03-22
Fix axes of new SNR plot.

tmiw:
Per https://github.com/drowe67/freedv-gui/pull/1250#issuecomment-4103004491 and https://github.com/drowe67/freedv-gui/pull/1250#issuecomment-4103045125, this PR does the following for the SNR graph:

1. Adjusts Y axes labels to improve alignment.
2. Only print X axes labels every 30 seconds.
3. Stop plotting SNR during TX.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -57,7 +57,7 @@
 
 // SNR plot constants
 #define NO_SNR_VAL (-10)
-#define MAX_SNR_VAL (40)
+#define MAX_SNR_VAL (35)
 #define SNR_PLOT_SECONDS (180)
 #define SNR_PLOT_SECOND_SEGMENTS (6)
 #define SNR_PLOT_DT (0.1)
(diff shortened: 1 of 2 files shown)


g4dya:
@tmiw that's certainly an improvement, but I still think it would look cleaner if the numbers at the edges were not displayed at all, as in this image, if that's possible.
Screenshot at 2026-03-22 08-00-27tweaked

barjac:
1. I agree with Richard but I'm guessing that the tools may not allow the removal of the end labels. Would increasing the whole widget X/Y (or padding) allow more space to stop them being pushed out of alignment?
2. The plot is not running in full duplex! It should only not run in TX when in half-duplex.

tmiw:
> 1. I agree with Richard but I'm guessing that the tools may not allow the removal of the end labels. Would increasing the whole widget X/Y (or padding) allow more space to stop them being pushed out of alignment?

First/last labels should be hidden now, but only on the SNR plot.

> 2. The plot is not running in full duplex! It should only not run in TX when in half-duplex.

Fixed.

g4dya:
> First/last labels should be hidden now, but only on the SNR plot.

That's working as expected for me. Thanks.

barjac:
Could the Y axis max. be 35 rather than 40, so the highest label would be 30?
Also I think the minus signs on ALL graph X axis labels could go.
Anyone who gets confused by that must be clairvoyant :)
Otherwise it's good for me.

PR #1265, opened 2026-03-24
Bring back RX frequency indicator for RADE.

### LINKED ISSUE #1235: Tab for Frequency Delta
### Contact Details

torbjorn.toreson@gmail.com and all using older transceivers

### Feature for this project

Before v 2.2.1 we had a Frequency Delta tab. It did not work with RadeV1 but with the other now legacy modes. The problem is that RadeV1 and presumably RadeV2 will not sync/decode if the stations transmitting or receiving have a frequency differece above +/- 50 Hz. Worse than +/- 50 Hz is not uncommon for older transceivers. E.g. my Emcomm group consist of more than 10 stations using older ICOM stations like IC-725 or IC-728, the frequency fault may be up to +/- 100 Hz (if I don't spend many hours to trim them) and the fault will never the less be temperature dependent, the result is that reception of Rade will be none or intermittent. When we use for example Winlink HF with VaraHF as modem, VaraHF will autocorrect up to +/- 100 Hz (called AFC, Automatic Frequency Correction) and will show the frequency delta on a meter. It is then possible to change ones own transceiver setting to eliminate the frequency difference if it is greater than the AFC can handle. It would be most convenient if at least with RadeV2 we would have an automatic frequency correction up to +/- 100 Hz as well as again having the Frequency Delta tab so we can see the frequency difference and understand why we don't get sync and be able to correct it manually, if needed.
/Torbjörn Toreson/SM6AYM

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

Testing

### Support assistance

Testing

tmiw:
Brings back RX frequency tuning for RADE (i.e. if you double click at the top of the waterfall or spectrum plots, the center frequency changes). This allows decoding RADE signals that are off-frequency, within reason.

Resolves #1235.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -899,7 +899,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Optimize "From XXX" plot performance. (PR #1238, #1239)
     * Logging: Send WSJT-X 'Decode' message to indicate RX'd SNR. (PR #1248)
     * Remove Python from RADE implementation to improve performance. (PR #1251)
-    * Add Tune button to allow tuning antennas/rigs. (PR #1259)
+    * Add Tune button to allow tuning antennas/rigs. (PR #1259, #1265)
     * Add SNR plot to main window. (PR #1250, #1261)
     * Bring back RX frequency indicator for RADE. (PR #1265)
 3. Build system:


barjac:
I don't suppose we could get the green RX frequency marker back? Without that it's a bit like shooting blindfold :)

tmiw:
> I don't suppose we could get the green RX frequency marker back? Without that it's a bit like shooting blindfold :)

Unfortunately the current RADE API doesn't support getting that info :(

PR #1266, opened 2026-03-25
Possible fix for audio dropouts on Linux.

tmiw:
In local testing, recordings of TX output from FreeDV had occasional dropouts. This PR contains changes to PulseAudio/pipewire handling that eliminates obvious dropouts in the test.wav file generated by the `rade_loss` test. However, this test is still consistently failing locally, so this requires further testing in the GitHub environment to see if the test pass rates are actually improved.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -269,6 +269,15 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    #- name: Setup tmate session
+    #  uses: mxschmitt/action-tmate@v3
+    #  with:
+    #    detached: true
+
+    - name: Disable AppArmor (needed for Fedora systemd)
+      run: |
+          sudo apparmor_parser -R /etc/apparmor.d/* || true
+
     - name: Save correct architecture
       id: save-arch
       run: |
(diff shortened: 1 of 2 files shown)


tmiw:
I'm not fully sure if this is helping yet as the pass rate on GH is already pretty good. In fact, I can consistently duplicate test failures locally again even with these changes. Maybe @barjac and/or @Tyrbiter can chime in?

Tyrbiter:
What pipewire version is GH using now? I have Fedora 44, with pipewire 1.6.2 but I don't have any obvious way of identifying any dropouts locally. Is there a test that can be run to produce a saved wav file that can be analysed in Audacity or similar?

tmiw:
> What pipewire version is GH using now? I have Fedora 44, with pipewire 1.6.2 but I don't have any obvious way of identifying any dropouts locally. Is there a test that can be run to produce a saved wav file that can be analysed in Audacity or similar?

Compile with `UT_ENABLE=1 ./build_linux.sh`, then you can run `ctest -R rade_loss` to generate a test.wav file that you can open in Audacity.

tmiw:
The Linux tests have passed for every commit in this PR so it doesn't make things *worse*, anyway. I'll merge for now and we can revert if we find later that it's causing problems.

PR #1268, opened 2026-03-28
Replace TX Attenuation slider with more precise and easy to adjust incement/decrement buttons.

barjac:
I have always found the 'TX Attenuation' slider control is too difficult to set accurately and having recently had similar issues with another application for the same reason I decided to try the same button approach in FreeDV.
The result is shown below and I find it much less stressful to use.
The widget name is also updated and there are brief tooltips associated with each button.

image

The double arrows provide 1.0dB adjustment and the single ones 0.1dB.
Nothing else has been changed.

### DIFF
--- a/src/topFrame.cpp
+++ b/src/topFrame.cpp
@@ -712,7 +712,7 @@ TopFrame::TopFrame(wxWindow* parent, wxWindowID id, const wxString& title, const
     txBtnSizer->Add(m_btnTxLevelPP, 1, wxEXPAND, 0);
     wxString fmtString = wxString::Format(MIC_SPKR_LEVEL_FORMAT_STR, wxNumberFormatter::ToString((double)0, 1), DECIBEL_STR);
 
-    m_txtTxLevelNum = new wxStaticText(m_txLevelBox, wxID_ANY, fmtString, wxDefaultPosition, wxDefaultSize, wxALIGN_CENTER);
+    m_txtTxLevelNum = new wxStaticText(m_txLevelBox, wxID_ANY, fmtString, wxDefaultPosition, wxDefaultSize, wxALIGN_CENTER | wxST_NO_AUTORESIZE);
     m_txtTxLevelNum->SetToolTip(_("Use mouse scroll wheel to adjust up or down"));
     m_txtTxLevelNum->SetMinSize(wxSize(100,-1));
     txLevelSizer->Add(m_txtTxLevelNum, 0, wxEXPAND, 0);


Tyrbiter:
Looks sensible to me, I agree that sliders can be awkward.

src/ongui.cpp
```diff
@@ -748,34 +748,34 @@ void MainFrame::OnCmdSliderScroll(wxScrollEvent& event)
...
-void MainFrame::OnChangeTxLevel( wxScrollEvent& )
+void MainFrame::applyTxLevel()
 {
     bool isTuning = m_btnTogTune->GetValue();
     wxString fmtString;
 
     if (isTuning)
     {
-        g_tuneLevel = m_sliderTxLevel->GetValue();
+        if (g_tuneLevel < -300) g_tuneLevel = -300;
```
tmiw: We should have #defines for the min/max tune and TX levels. I can make this change if you'd like?

tmiw:
BTW thanks for submitting this PR!

tmiw:
Question: Should the Speaker/Mic Level slider change too?

barjac:
The change to Release came from a merge back from master related to something else so I left it in. Feel free to revert.
Re. max/min yes please add max/min #defines as you wish.
I have an open mind about the Mic/Spkr slider as its not as critical or twitchy to use and I tend to leave it at zero. It would need 5 buttons with one to re-centre at 0.
I could not see the point of changing the widget title dynamically and TBH I had never even noticed it change previously! Revert that if you wish.

TreeTopHam:
Are 0.1 dB incremental changes really needed? 

Also, I prefer rolling the mouse wheel rather than trying to slide the mouse back and forth to adjust this slider or any of the other sliders. And my transceiver is totally fine with the coarser levels of adjustment.

No need to add extra buttons to clutter up the interface.

tmiw:
> Are 0.1 dB incremental changes really needed?

I was wondering that too. Are there any setups out there that need that level of precision?

barjac:
> Are 0.1 dB incremental changes really needed?
 
> I was wondering that too. Are there any setups out there that need that level of precision?

I suppose it depends on what you are doing. 
If you want to use the tune when taking measurements then its nice to have fine control. In regular QSO type operation 0.2dB steps would be adequate but I don't find a few extra clicks a problem.
I am open to trying 0.2dB steps, but I think the slider also used 0.1dB resolution IIRC. 
@TreeHopHam
Scroll wheels and mice/screen resolutions vary tremendously, button clicks do not!
You may have a very high end mouse and high resolution monitor but many including me don't!
The buttons occupy less screen space than the slider so it's really less clutter not more.
:)

tmiw:
How about 0.5 dB steps? Based on calculators I found online that would be a 10% adjustment in output, which seems low enough for all but the most sensitive setups, right?

barjac:
I don't see the point in doing that. The whole reason for course and fine adjustments is to offer just that. If you want a big step use the course and trim with the fine. I prefer it as is. My output power changes a little over time and where we are limited to quite low power its really nice now to use the fine control to tweak the level slightly during a long QSO.

tmiw:
> I don't see the point in doing that. The whole reason for course and fine adjustments is to offer just that. If you want a big step use the course and trim with the fine. I prefer it as is. My output power changes a little over time and where we are limited to quite low power its really nice now to use the fine control to tweak the level slightly during a long QSO.

Oh, I meant doing that with the slider, not the buttons like in this PR.

TreeTopHam:
What are the incremental dB steps in the slider as currently implemented today?

tmiw:
> What are the incremental dB steps in the slider as currently implemented today?

0.1 dB. I have a PR coming momentarily to bump that up to 0.5 dB in a manner that won't break existing configurations.

TreeTopHam:
Let's see how 0.5 dB works.
For the IC-7300, the TX slider adjusts ALC level, and the current granularity of the slider works very well and is very smooth. 
If I need to change RF power output, I use the RF Power control on the radio and not the TX slider.
I realize that other station equipment configurations may be different, however.

tmiw:
> Let's see how 0.5 dB works. For the IC-7300, the TX slider adjusts ALC level, and the current granularity of the slider works very well and is very smooth. If I need to change RF power output, I use the RF Power control on the radio and not the TX slider. I realize that other station equipment configurations may be different, however.

A Windows build with this change should be available at the bottom of https://github.com/drowe67/freedv-gui/actions/runs/23754431245?pr=1273 once GH finishes building. Let me know how it goes.

barjac:
> > Let's see how 0.5 dB works. For the IC-7300, the TX slider adjusts ALC level, and the current granularity of the slider works very well and is very smooth. If I need to change RF power output, I use the RF Power control on the radio and not the TX slider. I realize that other station equipment configurations may be different, however.

@TreeTopHam
How long did you spend testing this PR, and what problems did you encounter?
You say "I use the RF Power control on the radio and not the TX slider" so how would this PR affect you, other than to reduce the space occupied in the GUI?

> 
> A Windows build with this change should be available at the bottom of https://github.com/drowe67/freedv-gui/actions/runs/23754431245?pr=1273 once GH finishes building. Let me know how it goes.

Not good as far as I am concerned. The whole concept of trying to make sub millimetre adjustments using a mouse to set numeric values accurately is just flawed!

I have just been testing the ms-tx-level-slider-rounding branch and trying to adjust to a specific power level. It is just asking for far too much mouse precision, which neither I nor my mouse and/or screen resolution can achieve reliably and easily, the button approach requires no mouse dexterity at all.

There are also many operators older than me who will have reduced capability in mouse control and vision who probably find the sliders too difficult to use.

TreeTopHam:
>A Windows build with this change should be available at the bottom of https://github.com/drowe67/freedv-gui/actions/runs/23754431245?pr=1273 once GH finishes building. Let me know how it goes.

I tried the build on two different Windows machines this afternoon. My comments:

- The 0.5 dB steps give a smoother user experience. 
- Values are easier to target by moving the slider, and can be fine-tuned using the mouse scroll wheel. 
- The 0.5 dB increments seem to be more consistent across various mouse dpi, screen resolutions, and pointer settings.

That said, 

@barjac 
>The whole concept of trying to make sub millimetre adjustments using a mouse to set numeric values accurately is just flawed!

I totally agree! If one is wanting to set a precise _sub-decimal point number_ with a mouse slider, it becomes very maddening. You are trying to spear one of 300 possible values on a linear slide on a computer screen. In this case, buttons are a great solution.

How many users would need this kind of precision adjustment?

>You say "I use the RF Power control on the radio and not the TX slider" so how would this PR affect you, other than to reduce the space occupied in the GUI?

Actually, I use the TX slider all the time to keep the ALC adjusted properly. So, it plays a significant role.

### All good work and comments - thanks for everyone's skill and effort in making FDV a great product :)

tmiw:
@barjac, something that may help is that if you tab to the slider (or just click on it to give it focus), you can then use the arrow keys or Page Up/Page Down to adjust the value. The latest commit in the PR I previously linked will adjust by 0.5 dB when using the regular arrow keys and in my testing, seems to adjust by 3 dB when using Page Up or Page Down. I get how that still won't solve the problem if you prefer to (or can only) use the mouse, though.

Anyway, if we decide to go with your PR, I think I'd like some other changes:

1. Have the setting value display under the buttons, just like with the current slider.
2. Left/right arrows instead of up/down.

Another idea is to have only two buttons, with << and >> automatically happening if you hold down the associated < or > button long enough (say, after 1 dB of adjustment either way). I'm not sure what people think of that idea, so comments are welcome.

TreeTopHam:
I had not tried the arrow keys or Page Up/Page Down to adjust the slider value before. That's a **great** feature -  
On my PC, I seem to get five (5) left or right arrow presses before the displayed TX Attn value increments to the next 0.5 dB step. Does that mean that you could go back to a 0.1 dB step size and reliably use the arrow keys to navigate 0.1 dB steps?

Regardless of final step size, it would be nice to mention that functionality in a tool tip when hovering the mouse over the TX slider:  

- Pg Up/Dn: 3 dB step
- Arrows: xx dB step

tmiw:
> I had not tried the arrow keys or Page Up/Page Down to adjust the slider value before. That's a **great** feature - On my PC, I seem to get five (5) left or right arrow presses before the displayed TX Attn value increments to the next 0.5 dB step. Does that mean that you could go back to a 0.1 dB step size and reliably use the arrow keys to navigate 0.1 dB steps?
> 
> Regardless of final step size, it would be nice to mention that functionality in a tool tip when hovering the mouse over the TX slider:
> 
> * Pg Up/Dn: 3 dB step
> * Arrows: xx dB step

As a note, the 0.5 dB step with the arrow keys is with the version at https://github.com/drowe67/freedv-gui/actions/runs/23774857257/artifacts/6190479279, not the one previously linked. But yeah, if we continue with the slider instead of adopting this PR, it would be good to document this feature.

TreeTopHam:
Thanks for the updated link, which corrects the issue I mentioned above.

On previous builds with the 0.1 dB step size, the arrow keys change TX attn in 0.1 dB steps, and Pg Up/Dn keys in 3 dB steps. 

These keys provide the equivalent button functionality that @barjac was describing by providing a coarse and fine tune adjust.

Problem solved by just adding a tool tip to make everyone aware of the keyboard functionality :)

However, I find the 0.5 dB step size to be much smoother when using a mouse since the 300 possible slider values have been reduced to 60, a reduction factor of 80%.

tmiw:
I pushed changes to address the comments in this PR so far. The only thing left is to implement the two-button idea, but as I'm still not sure about that, I'm holding off for now.

barjac:
> @barjac, something that may help is that if you tab to the slider (or just click on it to give it focus), you can then use the arrow keys or Page Up/Page Down to adjust the value. The latest commit in the PR I previously linked will adjust by 0.5 dB when using the regular arrow keys and in my testing, seems to adjust by 3 dB when using Page Up or Page Down. I get how that still won't solve the problem if you prefer to (or can only) use the mouse, though.

I was aware of the arrow keys but not the Page up/down however I do prefer to use just the mouse rather than juggle two hands for a simple adjustment.
I am not sure whether the system left/right and pgup/down feature is available in all O.S. and desktop implementations.

> 
> Anyway, if we decide to go with your PR, I think I'd like some other changes:
> 
>     1. Have the setting value display under the buttons, just like with the current slider.

I originally had it that way, but then the mouse cursor can partly obscure the digits while you are trying to adjust it, hence the change to where it is now.

> 
>     2. Left/right arrows instead of up/down.
> 
> 
> Another idea is to have only two buttons, with << and >> automatically happening if you hold down the associated < or > button long enough (say, after 1 dB of adjustment either way). I'm not sure what people think of that idea, so comments are welcome.

I originally had << < > >> but changed it as left/right or fast forward/rewind is inferred and may be confusing, so I looked for UTF-8 up/down arrows which are there now.

Based on comments here I am making several changes:
1. Revert the Widget label to the dynamic one it was originally
2. Increase small step size to 0.2dB
3. Add +/- 0.2dB steps using mouse scroll wheel anywhere within the TX attenuator widget which work in both atten and tune modes.
4. Revert Debug/Release change
5. Add #defines for the min/max tune and TX levels.

barjac:
It looks like we crashed!
I just pushed 1,2 & 3 to my fork after merging back master, so what will happen here I don't know :/

tmiw:
> It looks like we crashed! I just pushed 1,2 & 3 to my fork after merging back master, so what will happen here I don't know :/

I think if you do a git pull there won't be any conflicts, or if there are just on the label change. But we'll see.

barjac:
I just built this PR as it is now and I have lost the 0.2dB steps on the small step, back to 0.1, lost the scroll wheel action but the widget label change is OK and the mouse cursor now partially obscures the digits again! :)
I tested my changes last night in QSO and with the scroll wheel addition it appeared to satisfy all the misgivings. There was no need to touch the buttons unless they were preferred as the scroll wheel provided 0.2dB resolution without any mouse dexterity required at all.

I have reset to your last commit so we are in sync now. 
I would like to see the button/digit reversal reverted and the scroll wheel action put back so people can test it and comment, as I feel it is excellent in operation. Can I do this with your agreement?

tmiw:
> I have reset to your last commit so we are in sync now. I would like to see the button/digit reversal reverted and the scroll wheel action put back so people can test it and comment, as I feel it is excellent in operation. Can I do this with your agreement?

Go for it.

barjac:
OK please test #5d76d92.
I have left the arrow symbols for the moment, but this now:

1. Moves the buttons to the middle row to avoid the mouse cursor from blocking vision of the digits when adjusting.
2. Uses 0.2dB as the smallest step which is precise enough and gives a nice response to small step clicks.
3. Using the mouse scroll wheel anywhere in the TX atten widget provides really easy control of either 'TX atten' or 'Tune' output in 0.2dB steps.

tmiw:
Just pushed a minor change to use names instead of hardcoded numbers for the small and large tuning steps. Seems to work for me here, anyway. I still haven't gotten a chance to try on Windows but hopefully soon, then we can merge unless there are objections?

barjac:
I think we need a tooltip to explain the scroll wheel action and what about the up down or right left arrow symbols? My opinion on that is that without a slider the concept of right and left is rather meaningless whereas up and down are more logical.

tmiw:
Unfortunately the arrows don't render well on Windows:

image

I don't think reverting to the other arrows will help, either. Suggest </<< and >/>> instead if that's okay with you?

barjac:
OK I will revert to to text chevrons and add tooltip. Also I notice that the Tune button tooltip says 1500Hz but I understood you changed it to 1475Hz?
I will correct that as well.

tmiw:
Revert 1475 back to 1500; this was changed in another PR.

barjac:
Scroll tooltip now activates over the whole top row, not just the value.
1500Hz restored in tune tooltip.
Text chevrons rather than graphic char for all operating systems.

Tested OK in Linux.

tmiw:
OK thanks. Let's merge for now and if there are complaints, we can revisit.

PR #1270, opened 2026-03-28
Flex: Prevent multiple slices from being in FreeDV mode.

tmiw:
Per user feedback, this PR adds logic to revert non-TX slices back to USB/LSB when switching the TX slice to FDVU/FDVL. This ensures that audio still routes properly and prevents crashes in the waveform and/or radio.


n61ab:
Mooneer.    I submitted a bug report that makes this PR moot.  That report centered on how the code determines the active slice causing issues when trying to match to the Flex TX slice.  Alan

PR #1283, opened 2026-04-04
Add green frequency indicator for RADE.

tmiw:
Turns out we *can* in fact get the RX offset from the RADE API, so this PR re-adds that functionality.


barjac:
Nice!
I have not yet tested with a RX signal, but since this is in both waterfall and spectrum, could the one in spectrum be moved to the top of the widget rather than the bottom for consistency?
Also I just noticed that the amber RX line is only intermittently showing in the waterfall whilst it is always there in the specrtum (this is with RX band noise)

Tyrbiter:
Excellent! Agree with @barjac that in the waterfall the orange line seems to vanish when at zero offset which it doesn't in the spectrum plot.

I actually prefer it on the edge where the freq values are, so it has to be different for spectrum vs waterfall.

barjac:
> Excellent! Agree with @barjac that in the waterfall the orange line seems to vanish when at zero offset which it doesn't in the spectrum plot.
> 
> I actually prefer it on the edge where the freq values are, so it has to be different for spectrum vs waterfall.

I actually meant to move the frequencies and markers too ;)

tmiw:
The disappearing line should be fixed now.

Re: positioning: the waterfall is the only plot with the X axis labels on top. Should those be moved to the bottom instead to match the others?

Tyrbiter:
I think the idea on the waterfall is that the newest waterfall data is nearest the frequency axis.

I don't know how it got to be this way, I think it was before @tmiw became the maintainer.

tmiw:
> I think the idea on the waterfall is that the newest waterfall data is nearest the frequency axis.
> 
> I don't know how it got to be this way, I think it was before @tmiw became the maintainer.

That would make sense. Seems like we'd also want the offsets to be near the frequencies on the spectrum plot, then?

barjac:
Could the waterfall direction be reversed as is used in some web SDRs? Though I see no reason why the spectrum should not have the frequency scale at the top, which in the case of the top of a FreeDV spectrum plot is probably nearer to the trace.
I have just seen an image in USER_MANUAL.md where both the waterfall and the spectrum were similar with the frequency marks and red line both at the bottom.

tmiw:
> Could the waterfall direction be reversed as is used in some web SDRs? Though I see no reason why the spectrum should not have the frequency scale at the top, which in the case of the top of a FreeDV spectrum plot is probably nearer to the trace. I have just seen an image in USER_MANUAL.md where both the waterfall and the spectrum were similar with the frequency marks and red line both at the bottom.

We'll need to update the manual no matter what we go with.

BTW are there any apps with a similar sort of spectrum plot thing going on? It would probably be good to be more in line with those.

Tyrbiter:
WSJT-X wide spectrum plot has the frequency axis scale at the top.

tmiw:
> WSJT-X wide spectrum plot has the frequency axis scale at the top.

Hmm, that combines a waterfall and a spectrum plot:

image

Frequency lines are at the top too.

Anyway, I'll think some more on this.

tmiw:
X axis on Spectrum plot moved to the top as of https://github.com/drowe67/freedv-gui/pull/1283/commits/290675e4c80a0e859ea9288571bdfc2670578029.

Tyrbiter:
> X axis on Spectrum plot moved to the top as of [290675e](https://github.com/drowe67/freedv-gui/commit/290675e4c80a0e859ea9288571bdfc2670578029).

Well that works, and looks fine to me although it will take a while to get used to it.

barjac:
Err... It has lost the frequ. digits and the tick marks are now little boxes on the spectrum. This is Wayland, Plasma 6.
And no Y axis labels either!

Screenshot_20260408_230745

tmiw:
> Err... It has lost the frequ. digits and the tick marks are now little boxes on the spectrum. This is Wayland, Plasma 6. And no Y axis labels either!

Do the numbers come back if you make the FreeDV window bigger, or are they still hidden?

barjac:
Yes but I will have a PR to fix it later.

Screenshot_20260409_170903

Issue #1199, opened 2026-01-28
Save TX attenuation per band.


barjac:
### Contact Details

G4MKT, MW0CDO, G8VLR, GW8LJJ

### Feature for this project

For various reasons output power across the HF spectrum for a given AF input can vary considerably, also there are local license requirements and personal preferences that come into play.

While we can all tweak settings as part of normal operating practice, it would be good if the FreeDV AF output could be remembered at least from band to band, rather than just having one setting stored.

I would only expect the band/setting to be stored on a clean close of the program so that only settings for bands actually used are remembered.

I would only expect the band/setting to be read once from the config if previously saved (i.e. it exists) on a change of band.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

As ususal

### Support assistance

As usual.

tmiw:
This is for the TX Attenuation slider, right? How much does it tend to vary in practice?

barjac:
> This is for the TX Attenuation slider, right? How much does it tend to vary in practice?

Yes the attenuation slider.

Well for me its as follows for 50W on each band using my Kenwood TS-450S:
```
Band (m)    Atten dB
160         10.4
80          13
60          14
40          15
20          15
17          15
15          15
12          13
10          10.9
```
So for me its a ~5dB range however there's probably no reason to limit the settings to other than the full 30dB provided by the slider in 0.1 dB steps. Some people seem to run with very little attenuation others with a lot from what I have heard them comment on the air.
I use more attenuation than above on top band as we are limited to 32W on the top end of the band, but for the above I kept all bands at 50w into a dummy load with the rig 'mic gain' fixed at the position I use (or used to :) for SSB.

tmiw:
Understood. Let me know once you've added the additional people to the original feature request and I'll forward this on for further discussion.

barjac:
More supporting call-signs now added.

tmiw:
> More supporting call-signs now added.

Thanks, I'll add it to the agenda for our next meeting.

tmiw:
Closing because #1284 has been merged.

PR #1284, opened 2026-04-05
Add optional per-band TX attenuation saving

barjac:
Please consider adding this to freedv-gui.

This adds the option to automatically save the TX audio output (TX power level) and/or Tune level for any or all bands.

It is enabled/disabled for the band in use by using the right click menus when mousing over either the top section of the TX/Tune Attenuation area in the main GUI, or the Tune button. 

The levels for the current band are saved on leaving a band or closing the program.

If a value has been saved it will be used next time the same band is used otherwise the default global value in the configuration will be used as was done in the past.

If you don't need this feature then just don't use it and nothing will change.

If you accidentally enable this feature for a band you can easily disable it using one of the context menus.

EDITED during development to reflect the current mode of operation.

[This enables the feature request here](https://github.com/drowe67/freedv-gui/issues/1199)


tmiw:
This PR isn't compiling for me. Investigating further.

tmiw:
I checked in a change to fix the compiler errors I was seeing. (Might be worth it to enable GitHub Actions in your fork, @barjac, so you can get notified of any build issues for future PRs.)

A few questions:

1. The current code seems to auto-load saved attenuation values when switching to a band with them. Do we need an explicit "load" option in the right-click menu given this?
2. Any way you can update the user manual to talk about this feature?

barjac:
1. No, it loads them if they are there, that's intentional. If anyone doesn't want this then the 'Remove' option clears that config entry and there will be nothing to auto-load. I don't want to add unnecessary actions.

2. Yes I will look at the docs and github actions.

tmiw:
I was questioning why the "Load" option in the right-click menu was needed if it was already auto-loading the saved TX attenuation when switching to a band (screenshot below):

image

Though I guess one could use that to restore the previous setting if they temporarily raised or lowered it for whatever reason?

barjac:
> Though I guess one could use that to restore the previous setting if they temporarily raised or lowered it for whatever reason?

Exactly :)

barjac:
USER_MANUAL.md updated

src/config/FreeDVConfiguration.cpp
```diff
@@ -242,6 +242,21 @@ void FreeDVConfiguration::load(wxConfigBase* config)
     
     load_(config, showDecodeStats);
     load_(config, enableLegacyModes);
+
+    // Load per-band TX attenuation values
+    txAttenByBand.clear();
+    config->SetPath("/TXatten");
```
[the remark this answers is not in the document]
tmiw: We may also want to extend ConfigurationDataElement so we can just pass this config field to the existing `load_` and `save_` functions.

tmiw (commented):
First code review pass. Unfortunately I didn't have much time to review but this should be enough to push through some edits.

src/main.h
```diff
@@ -527,6 +529,11 @@ class MainFrame : public TopFrame
         float      vk_rx_sync_time;
         bool suppressFreqModeUpdates_;
         bool firstFreqUpdateOnConnect_;
+        // Stores a FreeDVReporterDialog::FilterFrequency value. Declared as int
```
tmiw: I think we should bite the bullet and move FilterFrequency to another (possibly new) include file.

USER_MANUAL.md
```diff
@@ -268,7 +285,7 @@ demodulator.
 
 ## USB or LSB?
 
-On bands below 10 MHz, LSB is used for FreeDV.  On 10MHz and above, USB is used. After much debate, the FreeDV community has adopted the same conventions as SSB, based on the reasoning that FreeDV is a voice mode.
+On bands below 10MHz (except 5MHz), LSB is used for FreeDV.  On 10MHz and above, USB is used. After much debate, the FreeDV community has adopted the same conventions as SSB, based on the reasoning that FreeDV is a voice mode.
```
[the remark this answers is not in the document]
tmiw: Agreed + combine with the following paragraph as otherwise it'd be a single sentence paragraph.

src/main.h
barjac: Bed time here now, but as you wish :)

src/config/FreeDVConfiguration.cpp
```diff
@@ -62,6 +62,7 @@ FreeDVConfiguration::FreeDVConfiguration()
     , fifoSizeMs("/Audio/fifoSize_ms", (int)FIFO_SIZE)
     , transmitLevel("/Audio/transmitLevel", 0)
     , tuneLevel("/Audio/tuneLevel", 0)
+    , txAttenByBand("/TXatten", {})
```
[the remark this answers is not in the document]
barjac: Yes I was thinking the same about Tune as it would be expected and rather odd if it was not included.

barjac:
This now is passing tests OK and also working as initially envisaged.
However I have been re-thinking the Save/Load/Remove as it currently only saves values on program close that were manually 'Saved' during the session.
I can see users forgetting to save and losing settings, especially when changing bands.
I now feel that it should simply save the current values for all bands on program close without the need to manually 'save' at all.
If a level is never 'saved' on a band it will use the global setting as before.
If the save button is used once on a band the level will always be saved on close and restored on start for that band unless the setting is deleted by using the remove option.
If a setting is adjusted during a session, using the Load option will restore it to the value it was at the end of the previous session.
Some changes will be needed to the context menus, tooltips and manual and I will do this. 
I think it will mimic the current global saving method seamlessly and allow 'global' and 'per band' settings for those who need them on any or all bands.
All the above would apply equally to the new 'per band' Tune settings.

The context menu options/tooltips would change something like:

Save... ->  'Enable auto-saving TX attenuation for [band]'
Load... ->  'Restore TX level setting from last session for [band]' 
Remove... ->'Disable auto-saving TX level for [band]'
(Invalid options greyed as appropriate) Actually enable/disable could be a toggle.

USER_MANUAL.md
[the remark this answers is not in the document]
barjac: Done

USER_MANUAL.md
```diff
@@ -243,10 +246,26 @@ shift keying (PSK) so is not sensitive to amplitude.
...
+   * **Enable/Disable auto-save of TX atten for [band]** — performs the action described
+     When enabled the current level for that band will be saved on leaving the band or closing the program.
+   * **Restore TX atten level for [band]** — applies the attenuation value that
+     was saved for that band in a previous session.
+    
+   Within the TX Attenuation dialog there is also a Tune button which will output a continuous single tone at 1500Hz for antenna matching, netting etc.
+   The level of this signal while active may also be adjusted using the same controls as for the TX Attenuation.
+   Right click on the tune button to see a different context menu with similar options to the TX attenuation described above.
+   
+   FreeDV will automatically detect band changes (via Hamlib CAT or manual
```
tmiw: Add mention of OmniRig here as FreeDV also support this (Windows only).

USER_MANUAL.md
```diff
@@ -268,8 +287,7 @@ demodulator.
 
 ## USB or LSB?
 
-On bands below 10 MHz, LSB is used for FreeDV.  On 10MHz and above, USB is used. After much debate, the FreeDV community has adopted the same conventions as SSB, based on the reasoning that FreeDV is a voice mode.
-
+FreeDV currently follows the same conventions for upper and lower sideband as are used for amateur SSB.
 As an aid to the above, FreeDV will show the current mode on the bottom of the window upon pressing the Start button if Hamlib is enabled and your radio supports retrieving frequency and mode information over CAT. If your radio is using an unexpected mode (e.g. LSB on 20 meters), it will display that mode on the bottom of the window next to the Clear button in red letters. When a session is not active, Hamlib isn't enabled, or if your radio doesn't support retrieving frequency and mode over CAT, it will remain grayed out with "unk" displaying instead of the mode (for "unknown").
```
tmiw:
```suggestion
FreeDV currently follows the same conventions for upper and lower sideband as are used for amateur SSB. As an aid to the above, FreeDV will show the current mode on the bottom of the window upon pressing the Start button if Hamlib is enabled and your radio supports retrieving frequency and mode information over CAT. If your radio is using an unexpected mode (e.g. LSB on 20 meters), it will display that mode on the bottom of the window next to the Clear button in red letters. When a session is not active, Hamlib isn't enabled, or if your radio doesn't support retrieving frequency and mode over CAT, it will remain grayed out with "unk" displaying instead of the mode (for "unknown").
```

tmiw (commented):
Currently compiling the latest and may have additional feedback once that's done. In the meantime, see below.

src/ongui.cpp
```diff
@@ -755,6 +767,64 @@ void MainFrame::OnCmdSliderScroll(wxScrollEvent& event)
...
+    if (txAtten->find(bandName) != txAtten->end())
+        txAtten->insert_or_assign(bandName, g_txLevel);
+    auto& tuneAtten = wxGetApp().appConfiguration.tuneAttenByBand;
+    if (tuneAtten->find(bandName) != tuneAtten->end())
+        tuneAtten->insert_or_assign(bandName, g_tuneLevel);
+}
+
+// loadTxAttenForBand_() - load saved TX attenuation for the given band
+//-------------------------------------------------------------------------
+void MainFrame::loadTxAttenForBand_(const wxString& bandName)
```
tmiw: This should take the enum as the argument, not a `wxString`.

src/ongui.cpp
```diff
@@ -755,6 +767,64 @@ void MainFrame::OnCmdSliderScroll(wxScrollEvent& event)
...
+{
+    auto& atten = wxGetApp().appConfiguration.txAttenByBand;
+    auto it = atten->find(bandName);
+    g_txLevel = (it != atten->end()) ? it->second : (int)wxGetApp().appConfiguration.transmitLevel;
+    applyTxLevel();
+}
+
+// loadTuneAttenForBand_() - load saved tune attenuation for the given band
+//-------------------------------------------------------------------------
+void MainFrame::loadTuneAttenForBand_(const wxString& bandName)
```
tmiw: This should take the enum as the argument, not a `wxString`.

src/ongui.cpp
```diff
@@ -522,6 +524,16 @@ void MainFrame::onFrequencyModeChange_(IRigFrequencyController*, uint64_t freq,
...
         m_txtModeStatus->Refresh();
+
+        // Auto-save outgoing band levels, then load the new band's levels
+        auto newBandEnum = FreeDVReporterDialog::getFilterForFrequency_(freq);
+        if (newBandEnum != BAND_OTHER && newBandEnum != lastBand_)
+        {
+            autoSaveCurrentBandLevels_();
+            lastBand_ = newBandEnum;
+            loadTxAttenForBand_(bandNameForFilter(newBandEnum));
+            loadTuneAttenForBand_(bandNameForFilter(newBandEnum));
```
tmiw: This would cause `applyTxLevel` to be called twice, which doesn't seem necessary.

src/ongui.cpp
```diff
@@ -805,6 +883,70 @@ void MainFrame::OnTxLevelMouseWheel( wxMouseEvent& event )
...
+        : wxString::Format(_("Enable auto-save of TX atten for %s"),  bandName);
+    auto toggleItem  = menu.Append(wxID_ANY, toggleLabel);
+    auto restoreItem = menu.Append(wxID_ANY, wxString::Format(_("Restore TX atten level for %s"), bandName));
+    restoreItem->Enable(hasSaved);
+
+    menu.Bind(wxEVT_MENU, [bandName, hasSaved](wxCommandEvent&) {
+        if (hasSaved)
+            wxGetApp().appConfiguration.txAttenByBand->erase(bandName);
+        else
+            wxGetApp().appConfiguration.txAttenByBand->insert_or_assign(bandName, g_txLevel);
```
tmiw, line 929: This erases the previously saved TX level so restore effectively doesn't work.

src/ongui.cpp
```diff
@@ -805,6 +883,70 @@ void MainFrame::OnTxLevelMouseWheel( wxMouseEvent& event )
...
+        : wxString::Format(_("Enable auto-save of tune atten for %s"),  bandName);
+    auto toggleItem  = menu.Append(wxID_ANY, toggleLabel);
+    auto restoreItem = menu.Append(wxID_ANY, wxString::Format(_("Restore tune atten level for %s"), bandName));
+    restoreItem->Enable(hasSaved);
+
+    menu.Bind(wxEVT_MENU, [bandName, hasSaved](wxCommandEvent&) {
+        if (hasSaved)
+            wxGetApp().appConfiguration.tuneAttenByBand->erase(bandName);
+        else
+            wxGetApp().appConfiguration.tuneAttenByBand->insert_or_assign(bandName, g_tuneLevel);
```
tmiw, line 971: Similar problem for tuning levels.

tmiw (changes requested):
I don't think the current code is behaving quite right. It looks like it is indeed autosaving the TX/Tune Attenuation for the current band (per manual inspection of the config file), but it's not really restoring. Test steps:

1. Start on 20 meters, set attenuation to -1dB.
2. Switch to 40 meters, set attenuation to -2dB.
3. Switch back to 20 meters and request to restore the 20 meter attenuation.

Expected: attenuation set to -1dB.
Actual: attenuation remains the same.

USER_MANUAL.md
barjac: Since with the change the text is no longer 'above' I edited it from 'As an aid to the above' to 'As an aid to this'

USER_MANUAL.md
barjac: Done and also mentioned in earlier preamble.

src/ongui.cpp
barjac: This is intended. It turns off the auto-save feature by removing the config entry for that band.

Restore works for the current band by storing (mem only) the values it gets from the config when entering/starting use of that band. Any changes in values will be auto saved on leaving the band or on program close. Alternatively they may be restored to the initial value when use of that band began.

barjac:
All requested changes have been made and test OK so far.
I don't know what the 1 Change request is below or when it was posted as there is no date or link that works :\

src/ongui.cpp
tmiw: Makes sense, thanks for the clarification!

tmiw (approved):

PR #1287, opened 2026-04-08
Implement proposed fixes to reduce dropouts.

tmiw:
Claude Code recommended some fixes to `GenericFIFO` and `PlaybackStep` to reduce dropouts in the GitHub environment. This PR is so we can try these changes out and see if there actually is any improvement.

Also contains some refactoring of short/float sample conversions (based on work originally in #1286).


g4dya:
I used 0565ef0 this morning in QSO with @barjac and others and it seemed to give fewer dropouts. That's just an impression, not at all scientific.

PR #1288, opened 2026-04-09
Make Waterfall and Spectrum displays usable and tidy when compressed into a small main window 

barjac:
This updates the Waterfall and Spectrum graph displays to accept more size compression whilst keeping axis labels tidy and readable.

Ideas by me, coding by Claude AI :)

Screenshot_20260409_202841

Below is current master branch at the same size:

Screenshot_20260409_210822


tmiw:
Here's what this PR looks like on macOS:

image

Can we revert the Y axis back to showing labels every 5s? Or if that's not acceptable, maybe every 2 or 2.5s? Otherwise I think it's fine.

barjac:
With 5s second marks I would only see 0s! With 2s marks I would see 0s, 2s and 4s and with 2.5s only 0s and 2.5s.
I will have a think :)
Why do you use up so much desktop with the waterfall? I try to keep it compact :)

My typical layout these days is:

Screenshot_20260409_224756

barjac:
Try it now :)

PR #1290, opened 2026-04-13
Log heard callsigns to a CSV file.

### LINKED ISSUE #1279: [Feature Request] Save station callsigns heard and details to a local file
### Contact Details

@tyrbiter, @digver, @barjac 

### Feature for this project

I would like a permanent record of the stations heard list to be kept. Similar to other local files in ~/freedv/ with a format like that in the drop down list.

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

I think this should be a very simple and straightforward implementation, as simple as writing the same output into a file as it is already generated while freedv-gui is running.

### Support assistance

I don't see this as needing any support as such.

tmiw:
Logs heard callsigns to a CSV file. Resolves #1279.

Disclaimer: implemented with the assistance of Claude Code.


tmiw:
@barjac / @Tyrbiter, let me know how this works for you. The file that's created should end up in your Documents folder but the exact path will show up in the log output.

Tyrbiter:
I have a build with this in now, will report as stations are heard.

Seems to work as expected, the file contains lines similar to this:

date,time,callsign,mode,frequency_hz,snr_db
2026-04-13,14:13:18,G4HJW,RADEV1,5368500,18
2026-04-13,14:14:47,G4HJW,RADEV1,5368500,19

and so on. This is the file head.

tmiw:
Good to hear that it's working. Should this have any form of configurability? For example, where to store the CSV file.

Tyrbiter:
> Good to hear that it's working. Should this have any form of configurability? For example, where to store the CSV file.

I think there should be a setting for this, perhaps added to the Options->Reporting tab with a default path of ~/freedv

Tyrbiter:
Merge in master again?

tmiw:
> Merge in master again?

Done. 

> I think there should be a setting for this, perhaps added to the Options->Reporting tab with a default path of ~/freedv

Added setting to Tools->Options->Reporting but I'm not sure about the name. Suggestions would be good here.

Tyrbiter:
> Added setting to Tools->Options->Reporting but I'm not sure about the name. Suggestions would be good here.

Stations heard log file?

tmiw:
> > Added setting to Tools->Options->Reporting but I'm not sure about the name. Suggestions would be good here.
> 
> Stations heard log file?

Updated. If it looks good on your end, we'll go ahead and merge.

Tyrbiter:
> > Stations heard log file?
> 
> Updated. If it looks good on your end, we'll go ahead and merge.

Seems good here, file getting longer, formatting seems good.

digver:
> > Added setting to Tools->Options->Reporting but I'm not sure about the name. Suggestions would be good here.
> 
> Stations heard log file?

In the options would 'stations received log' sound better. Also rather than a monolithic file maybe save a seperate file for each session (app start up). So name it something like 'freedv-rx-yyyymmdd-hh:mm:ss'?

73, digger ab3xu

digver:
> > Good to hear that it's working. Should this have any form of configurability? For example, where to store the CSV file.
> 
> I think there should be a setting for this, perhaps added to the Options->Reporting tab with a default path of ~/freedv
> 

Having just tried out the build it looks like the default for Linux places it in the \~/Documents folder. Personally, I don't like that, I wouldn't use that folder to store a log file.
I agree it should be an option, but in Linux wouldn't the convention be more like ~/.local/freedv/\<file\>?

73, digger ab3xu

tmiw:
> Having just tried out the build it looks like the default for Linux places it in the ~/Documents folder. Personally, I don't like that, I wouldn't use that folder to store a log file. I agree it should be an option, but in Linux wouldn't the convention be more like ~/.local/freedv/<file>?

I created https://github.com/drowe67/freedv-gui/pull/1293 to ensure that the default path uses the XDG specification, but in its current form it still seems to be ~/Documents on Ubuntu 24.04 anyway. Might be good to get some feedback from other Linux users here.

tmiw:
> In the options would 'stations received log' sound better. Also rather than a monolithic file maybe save a seperate file for each session (app start up). So name it something like 'freedv-rx-yyyymmdd-hh:mm:ss'?

IIRC WSJT-X saves one single file for logs, so it might be okay as-is. We'll see if there are others who prefer this.

barjac:
Does this affect the UDP logging feature in any way?
I have a WIP PR for klog regarding that feature which fixes various issues and is almost finalized, I'm hoping that this won't break it.
\OT
One query I have is why does the UDP packet not include the QSO date/time as shown in the pull down list? A QSO may be logged long after it took place and the only time we can use is the UDP packet transfer time.

tmiw:
> Does this affect the UDP logging feature in any way? I have a WIP PR for klog regarding that feature which fixes various issues and is almost finalized, I'm hoping that this won't break it.

This should not affect that feature.

> One query I have is why does the UDP packet not include the QSO date/time as shown in the pull down list? A QSO may be logged long after it took place and the only time we can use is the UDP packet transfer time.

I looked at the code again to make sure and it in theory *should* be sending the drop-down list time. Does the correct time appear in the log window at least?

barjac:
> > Does this affect the UDP logging feature in any way? I have a WIP PR for klog regarding that feature which fixes various issues and is almost finalized, I'm hoping that this won't break it.
> 
> This should not affect that feature.

OK thanks.
> 
> > One query I have is why does the UDP packet not include the QSO date/time as shown in the pull down list? A QSO may be logged long after it took place and the only time we can use is the UDP packet transfer time.
> 
> I looked at the code again to make sure and it in theory _should_ be sending the drop-down list time. Does the correct time appear in the log window at least?

I'm not sure what you mean by the 'log window list'.
In the logging confirmation dialogue that pops up after hitting the main GUI log button there is only the time.
In the UDP packet I intercept from FreeDV there is no time/date field sent at all.
If the log is sourced from the Reporter then the last TX date/time could be sent, if from the drop-down the actual date/time of the entry at EOO could be used, but these would really need the **date** and time to be of any use for logging to avoid ambiguity around midnight.

tmiw:
If it helps at all, the WSJT-X specification uses the following encoding for date + time:

```c++
    *this   << (int64_t)julianDays.count()
            << (uint32_t)msSinceMidnight
            << (unsigned char)1     // UTC
```

Today is 26108 in Julian days (18 April 2026), which would be encoded in the UDP packet as IIRC 00 00 00 00 00 00 65 FC (as big endian is used).

tmiw:
Example from Wireshark (I clicked on a FreeDV Reporter user and then clicked Log QSO):

image

tmiw:
Er, looks like today's date translates to 00 00 00 00 00 25 8D DD as there's logic to add 58574112 hours to the current number of seconds or whatever since the Unix epoch (58574112 / 24 + 20561 days since January 1, 1970 = 2,461,149 or 0x258DDD).

PR #1309, opened 2026-04-27
Allow use of a custom key for PTT.

### LINKED ISSUE #907: [Feature Request] Configurable PTT Key
### Contact Details

@awatchar, @Tyrbiter, @barjac

### Feature for this project

Hi,

I’d like to request a small but useful feature: the ability to customize the PTT (Push-To-Talk) key in FreeDV GUI.

In particular, it would be helpful if we could map the PTT function to non-standard keys, such as the Volume Up key, or any other key that the user prefers.

Thanks for your great work on FreeDV. This small change could help more people integrate it into their hardware and field setups.

E25FGL

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Non-standard compiler

None

### Implementation assistance

- Add an option in the settings to choose the PTT key
- Support multimedia keys like Volume Up, or allow “press-to-assign” configuration

### Support assistance

I'm happy to support FreeDV for the next 1–5 years, especially in making it more accessible in Southeast Asia.

tmiw:
Allows use of a custom key instead of Space for triggering PTT. Resolves #907.

Disclaimer: implemented with the assistance of Claude Code.


tmiw:
The key selection logic in Tools->Options isn't working (at least on macOS) but I ran out of time to investigate. Will get back to this later.

Tyrbiter:
Is this in addition to the space bar and PTT button?

tmiw:
Interesting, I just tried this branch on my MacBook and the Tools->Options thing worked as expected. Wonder why it didn't work on my Mac Mini.

> Is this in addition to the space bar and PTT button?

Instead of.

Tyrbiter:
> > Is this in addition to the space bar and PTT button?
> 
> Instead of.

How difficult to allow both?

Assume the button will always be clickable.

tmiw:
> > > Is this in addition to the space bar and PTT button?
> > 
> > 
> > Instead of.
> 
> How difficult to allow both?
> 
> Assume the button will always be clickable.

Shouldn't be difficult to do, but wouldn't that create confusion?

Tyrbiter:
> > > > Is this in addition to the space bar and PTT button?
> > > 
> > > 
> > > Instead of.
> > 
> > 
> > How difficult to allow both?
> > Assume the button will always be clickable.
> 
> Shouldn't be difficult to do, but wouldn't that create confusion?

I suppose it might, I wondered if the config options could allow either to be selected or both. That might be hidden behind the experimental features setting?

tmiw:
Wait, I should clarify. PTT button will always work with this PR, but the keyboard shortcut will either be Space or whatever the user selects, not both. Did you mean that Space *and* a user selectable key should work?

Tyrbiter:
> Wait, I should clarify. PTT button will always work with this PR, but the keyboard shortcut will either be Space or whatever the user selects, not both. Did you mean that Space _and_ a user selectable key should work?

That was my suggestion, but I can see it might confuse some users hence the experimental features setting for this aspect. I was thinking of my foot switch, but actually that plus the button should be adequate. I can make the foot switch emulate various key codes.

tmiw:
> > Wait, I should clarify. PTT button will always work with this PR, but the keyboard shortcut will either be Space or whatever the user selects, not both. Did you mean that Space _and_ a user selectable key should work?
> 
> That was my suggestion, but I can see it might confuse some users hence the experimental features setting for this aspect. I was thinking of my foot switch, but actually that plus the button should be adequate. I can make the foot switch emulate various key codes.

Let's keep the behavior as-is for now without hiding under the experimental features setting. If we get feedback that Space should also be allowed, we can always do a followup PR.

tmiw:
OK, I did some more testing here and found the following:

1. PTT key adjustment in Windows, Linux (Fedora 43 in a VM) and my Intel MacBook Pro (running Tahoe) seems to work fine.
2. PTT key adjustment on my M1 Mac Mini does not--the dialog box that pops up when pushing the Change button doesn't close regardless of the key pressed.

I'm still not sure about (2), so it would be good if another ARM-based Mac user could confirm what I'm seeing.

tmiw:
I removed the separate capture dialog box and am now having the capture happen in the Options dialog. Let me know how this works for you, @Tyrbiter. FWIW it seems to fix the problem on ARM Macs.

Tyrbiter:
Will try, but still having github problems here, trying to do a git pull --all is giving me errors.

During my build I see this despite having tried git reset --hard:

+ git pull
remote: Enumerating objects: 100, done.
remote: Counting objects: 100% (72/72), done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 100 (delta 47), reused 54 (delta 38), pack-reused 28 (from 1)
Receiving objects: 100% (100/100), 199.37 KiB | 19.94 MiB/s, done.
Resolving deltas: 100% (56/56), completed with 14 local objects.
fatal: bad object refs/remotes/origin/ms-flex-waveform-doc
error: https://github.com/drowe67/freedv-gui.git did not send all necessary objects

Ah, needed to pull the whole master tree again into a clean directory.

Will build this branch later today. Now built and running, not yet had time to try the PTT key change feature.

tmiw:
Speaking of which, I'll update this branch with current `master` so that the automated builds start working again. Let me know once you do get around to testing the PTT key feature.

Tyrbiter:
OK, I have tried this out with /etc/udev/rules.d/10-usb-footpedal.rules having a rule for my footpedal USB VID and PID and /etc/udev/hwdb.d/60-usb-footpedal.hwdb containing a setting to map the relevant key code identified by evtest to f14 which I can't press on the keyboard. It's also necessary to unplug and replug the footswitch after making these changes.

When I select the Change... button in the Options->Rig control tab and press the footswitch then f14 shows up as Key(437) and pressing the footswitch now enables the PTT.

So, it appears to work when changing the PTT key from Space to f14.

We can decide later if it would be useful to have a setting that allows Space to be retained as well. Of course it also requires the freedv-gui window to have focus although the main window grabs focus from the Reporter window if that is open.

tmiw:
> OK, I have tried this out with /etc/udev/rules.d/10-usb-footpedal.rules having a rule for my footpedal USB VID and PID and /etc/udev/hwdb.d/60-usb-footpedal.hwdb containing a setting to map the relevant key code identified by evtest to f14 which I can't press on the keyboard. It's also necessary to unplug and replug the footswitch after making these changes.
> 
> When I select the Change... button in the Options->Rig control tab and press the footswitch then f14 shows up as Key(437) and pressing the footswitch now enables the PTT.
> 
> So, it appears to work when changing the PTT key from Space to f14.
> 
> We can decide later if it would be useful to have a setting that allows Space to be retained as well. Of course it also requires the freedv-gui window to have focus although the main window grabs focus from the Reporter window if that is open.

Excellent. I checked in a change to hopefully have it show F14 instead of Key(...).

Tyrbiter:
> Excellent. I checked in a change to hopefully have it show F14 instead of Key(...).

Might this depend on which wxGTK library version is available? It doesn't change the Key(437) even after I click Change... and press the foot switch.

tmiw:
> > Excellent. I checked in a change to hopefully have it show F14 instead of Key(...).
> 
> Might this depend on which wxGTK library version is available? It doesn't change the Key(437) even after I click Change... and press the foot switch.

Not sure, I'm getting the same thing on both Plasma and GNOME (I can only test this by using VMware's "Send Key" function). It doesn't look like the AppImage helps either even though it's on a newer wxWidgets.

The good news is that we can manually check for 436 (F13, even though both DEs seem to hijack it to bring up the system settings) through however many keys we have.

Tyrbiter:
That seems to have fixed it, I now see F14 in the key selection field.

tmiw:
> That seems to have fixed it, I now see F14 in the key selection field.

Good to hear. I'll have to figure out a way to test on Windows and macOS (I don't have a keyboard with a F14 key). Might need to wait until after 2.3.0's released.

Tyrbiter:
> > That seems to have fixed it, I now see F14 in the key selection field.
> 
> Good to hear. I'll have to figure out a way to test on Windows and macOS (I don't have a keyboard with a F14 key). Might need to wait until after 2.3.0's released.

I suppose there must be a way to create arbitrary key codes on other OSes, is it possible using alt-<nnn> where nnn is entered on the numpad or something like that.

tmiw:
Looks like Windows Power Toys will. Unfortunately FreeDV doesn't ignore the Windows key (I mapped Windows-F2 to F14), so I'll need to fix that.

Note for self: 393 is the (left?) Windows key.

tmiw:
OK, I think it's working now on Windows at least. We can probably merge immediately after 2.3.0's released. 👍

digver:
Sorry, I'm getting here a little late. As far as I can see PTT is only latching. Would adding the option to make it momentary be a separate feature request? I tend to forget that it's latching and everything goes crazy when I press and hold the space bar!

tmiw:
> Sorry, I'm getting here a little late. As far as I can see PTT is only latching. Would adding the option to make it momentary be a separate feature request? I tend to forget that it's latching and everything goes crazy when I press and hold the space bar!

Yeah, I would say that should be a separate FR. 👍

digver:
Sure. Thinking about it, I can convince myself that either one is better. I'll write up a request for it and see where it goes.

Tyrbiter:
The question is, what makes more sense?

With my HF radios here the Send or Transmit buttons on the front panel are latching, whereas the PTT line on the mic and rear sockets are momentary and so need to be held to remain on transmit.

We're into the options problem again, and in this case there is software involved which can certainly mean extra care is required to avoid things getting stuck in a state that can cause damage to the hardware.

tmiw:
> The question is, what makes more sense?
> 
> With my HF radios here the Send or Transmit buttons on the front panel are latching, whereas the PTT line on the mic and rear sockets are momentary and so need to be held to remain on transmit.
> 
> We're into the options problem again, and in this case there is software involved which can certainly mean extra care is required to avoid things getting stuck in a state that can cause damage to the hardware.

Every radio I've ever used needed the PTT button to be held down while transmitting, so to me that seems like it makes more sense for FreeDV to adopt (especially since PTT input is effectively this). I do get that PTT has generally been latching throughout FreeDV's existence, though.

digver:
I have a handheld PTT with two buttons. One is latching and the other momentary -- the best of both worlds without any reconfiguring.

In Freedv latching is nice because I can go to another window to look at something -- say to look at the waterfall in aether, or look up a call sign in the browser. On the otherhand, I have to get back to the freedv window to end the transmission. 

I'm more accustomed though to momentary ptt. Both with hardware radios and maybe all software radios for voice tx -- consistancy is good. If I'm using  freedv and aether in the same session I'd prefer them both to have the same behavior

w4mmp-ron:
Hi,
I want to chime in on this.  Been a ham for a long time.  There is a distinction between buttons physically on the radio (referring to HF rigs) and the button on the microphone attached to the radio.  The PTT and or TUNE buttons ON the radio have always been (at least on the rigs I have used) are ON/OFF (latching).  The PTT button on the microphone has always been momentary.  This is the same for a foot actuated PTT switch.  It is also momentary.    HTs obviously have almost always been momentary.  It all depends what you are wanting to emulate.  The control panel (face of the rig) or the microphone.  You all can debate what you want to emulate.  
73,
Ron

digver:
I've made a feature request at #1331 "Provide an option for the PTT key to be latching or momentary"

It will need two other people added to it for it to be considered. If it looks worthwhile to you I can add your name/email to it

Tyrbiter:
Now that 2.3.0 is released it would be useful to merge master into this PR I think.

PR #1310, opened 2026-04-27
Improve Reporter message list usability

barjac:
This adds the ability to 
- Delete a message
- Edit a message
- Add a message

- Also keeps the list sorted alphanumerically and increases the maximum entries to 15.

The problems resolved by this PR:
When the current selection list of previously saved messages grows to 10, there is no way in the GUI to remove any obsolete ones. New messages then overwrite essentially random messages with new ones. This is a bug since the user has no control over which messages will be replaced.
The list is not sorted, so messages of a similar type are scattered around the list based on when they were created.
Ten entries is only marginally adequate for some use cases.

Created in collaboration with Claude Code.


src/gui/dialogs/freedv_reporter.cpp
```diff
@@ -64,11 +68,126 @@ extern FreeDVInterface freedvInterface;
...
 #define RX_COLORING_SHORT_TIMEOUT_SEC (5)
 #define MSG_COLORING_TIMEOUT_SEC (5)
-#define STATUS_MESSAGE_MRU_MAX_SIZE (10)
+#define STATUS_MESSAGE_MRU_MAX_SIZE (15)
 #define MESSAGE_COLUMN_ID (6)
 
 using namespace std::placeholders;
 
+// Custom combo popup: a wxListBox that supports per-item right-click context menus.
+class MsgListPopup : public wxListBox, public wxComboPopup
```
tmiw: Can this be moved to a separate .cpp/.h file under src/gui/controls?

src/gui/dialogs/freedv_reporter.cpp
```diff
@@ -64,11 +68,126 @@ extern FreeDVInterface freedvInterface;
...
+            wxPoint screenPos = combo->GetScreenPosition();
+            wxSize  popupSz   = popupWin->GetSize();
+            popupWin->Move(screenPos.x, screenPos.y - popupSz.GetHeight());
+        }
+    }
+
+    wxSize GetAdjustedSize(int minWidth, int /*prefHeight*/, int maxHeight) override
+    {
+        int count = std::max(1, (int)GetCount());
+        // Estimate row height from font metrics; fall back to 22px.
```
tmiw: Why was 22px chosen? Why are we adding 8 to the character height?

src/gui/dialogs/freedv_reporter.cpp
```diff
@@ -64,11 +68,126 @@ extern FreeDVInterface freedvInterface;
...
+    void OnRightUp(wxMouseEvent& event)
+    {
+        if (pendingContextItem_ == wxNOT_FOUND)
+        {
+            event.Skip();
+            return;
+        }
+        int item = pendingContextItem_;
+        pendingContextItem_ = wxNOT_FOUND;
+        Dismiss();
```
tmiw: Should the context menu be shown before dismissing?

src/gui/dialogs/freedv_reporter.cpp
```diff
@@ -368,7 +487,8 @@ FreeDVReporterDialog::FreeDVReporterDialog(wxWindow* parent, wxWindowID id, cons
     auto statusMessageLabel = new wxStaticText(this, wxID_ANY, _("Message:"), wxDefaultPosition, wxDefaultSize);
     statusMessageSizer->Add(statusMessageLabel, 0, wxALL | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL, 5);
 
-    m_statusMessage = new wxComboBox(this, wxID_ANY, _(""), wxDefaultPosition, wxSize(180, -1), 0, nullptr, wxCB_DROPDOWN | wxTE_PROCESS_ENTER);
+    m_statusMessage = new wxComboCtrl(this, wxID_ANY, _(""), wxDefaultPosition, wxSize(180, -1), wxTE_PROCESS_ENTER);
```
tmiw, line 374: Add tooltip to indicate that you can right-click on individual items in the list?

tmiw (commented):
See comments below. I don't think I'll have time to try to compile and run this locally today but hopefully soon.

src/gui/dialogs/freedv_reporter.cpp
barjac: **Moved MsgListPopup to src/gui/controls/** — extracted inline class from  `freedv_reporter.cpp`. Circular dependency broken by replacing `FreeDVReporterDialog*`  with `std::function<void(int)>` callback. Constructor call uses lambda:   `[this](int idx){ ShowMsgItemContextMenu(idx); }`.   `fdv_gui_dialogs` now links against `fdv_gui_controls` (needed for symbol resolution).

src/gui/dialogs/freedv_reporter.cpp
barjac: **GetAdjustedSize magic numbers removed**
Simplified to `GetCharHeight() + 8` unconditionally.
Window is realised by the time this is called so GetCharHeight() always returns a valid value.
+8 approximates wxListBox internal item padding.

src/gui/dialogs/freedv_reporter.cpp
barjac: **Dismiss() before context menu order**
Confirmed correct as-is; showing menu before Dismiss() would cause two overlapping popups with z-order/focus issues.

barjac:
**Text selection after item pick — fixed** — after selecting an item, GTK's focus-in handler selects all text in the GtkEntry.
Fixed with a 20ms one-shot wxTimer (deselTimer_) started in OnLeftUp after Dismiss().
The timer handler calls tc->SetSelection(end, end) which collapses the selection to the cursor at end of text.
20ms is imperceptible but reliably runs after GTK's focus-in processing.
The selected message text remains visible in the entry box.

tmiw:
Thanks for the changes. Let me know once pushed to this PR so I can take a look.

barjac:
I think we doubled - its there now :)

tmiw:
Got the following on macOS while running this PR:

image

src/gui/dialogs/freedv_reporter.cpp
barjac: **Tooltip on combo control** — a tooltip on the wxComboCtrl text entry is not viable on GTK (I-beam cursor means GtkEntry owns the mouse; GTK suppresses tooltips\n   on focused text widgets). The existing tooltip in MsgListPopup::Create() on the popup listbox is the appropriate place.

barjac:
Try it now.

tmiw:
Similar error when right-clicking on the Send button:

image

barjac:
Too late now but I will take a look in the morning :)

barjac:
See how that goes, all instances of the same issue should be fixed now.

barjac:
Now persists the width of the edit/create dialogue between runs which makes repeated editing of long messages more user friendly.

tmiw:
OK I think it's working properly on macOS now. The main addition is the right-click list when you expand the message drop-down, right? All the options there seem to work, at least.

Question: some of this functionality seems to be duplicated in the right-click menus for Send and Clear. Should those be adjusted or removed entirely?

barjac:
> Question: some of this functionality seems to be duplicated in the right-click menus for Send and Clear. Should those be adjusted or removed entirely?

I don't think anything needs removing.
The options in the list context menu complement those accessed from the 'Send' button.
Ah! I see what you mean. I had not seen the options under the 'Clear' button!
I will think about that.

I just found a tiny bug that may or may not be fixable. If there are long messages in the pull down it has a scroll bar at the bottom to allow sliding across to view them, however on releasing the mouse button after using the scroll the list collapses.

tmiw:
> I just found a tiny bug that may or may not be fixable. If there are long messages in the pull down it has a scroll bar at the bottom to allow sliding across to view them, however on releasing the mouse button after using the scroll the list collapses.

This sounds like the kind of thing that might be DE-dependent :( macOS for example doesn't give me a scroll bar:

image

barjac:
Going back to your earlier question, I think the context menu from 'Clear' can be removed now.
Would anyone really want to remove all messages in one go?
This PR will provide single message delete from the list.

Scroll bar now fixed.

tmiw:
> Going back to your earlier question, I think the context menu from 'Clear' can be removed now. Would anyone really want to remove all messages in one go? This PR will provide single message delete from the list.
> 
> Scroll bar now fixed.

I could see someone wanting all of them to be cleared without needing to fully reset their config, but I don't know how often that would happen either.

Another question: should the 10 (?) message limit go away?

barjac:
If the duplication does not worry you then I see no compelling reason to remove the context menu from Clear.
A 'clear all' option could be added to the new menu but I don't like that since it would be too easy to click it accidentally, making a confirmation dialogue necessary. 
If you feel that the extra confirmation step is worthwhile then I am happy to add the 'clear all' option with confirmation so the Clear context menu can be removed.

Regarding the 10 limit, I increased it to 15 as I find that better for my use, while still keeping the full list visible on the screen.
Are you meaning to remove the limit completely? Limitless sounds a bit extreme :)
Maybe add the max value to the config rather than hard coding it and perhaps add it to the misc settings with a default of say 20?

tmiw:
Let's keep the context menus as-is and the limit to 15 so we don't keep churning on this. I'll see if I can test again on macOS as well as Windows and do a final review of the code. I'm not sure we'll be able to merge until after 2.3.0 gets released, though.

tmiw:
Tested on Windows and it seems to be fine. One thing I noticed was that there's an "Add new message" item in the new right-click menu, which works when there are already items in there but the menu isn't accessible if there aren't any items in the list already. Is that okay given there are several "save" options when you right-click on the Send button?

barjac:
Good catch, however I think it is OK. This is primarily for managing messages that already exist. As you say the Send button menu would be used for initial saves.

USER_MANUAL.md
```diff
@@ -420,6 +420,28 @@ in sync with the radio. For example, in the case of ALE, the radio's frequency c
...
 
+The FreeDV Reporter window includes a **Message:** field that allows you to broadcast a
+short status message (such as your location, antenna, or current activity) to other FreeDV
+Reporter users. Type a message and click **Send** to broadcast it, or **Clear** to remove
+the current message.
+
+*Note: messages are only broadcast when reporting is enabled in Tools->Options->Reporting
+and the **Reporting** toggle button in the main window is on.*
+
+Messages are not saved automatically. To save a message for later reuse, right-click the
```
tmiw, line 431: Would it make sense to change the behavior of the Send button so that it actually *does* save when you push it? (i.e. no right-click needed) That seems like something users would expect to have happen, right?

barjac: No. In our regular nets and also in 1:1 QSOs the reporter is also used to make quick comments which are unique and would not be saved. There are many cases where the same messages are used regularly, the list is ideal for these but it would become cluttered if all were saved. I think the way this is currently working in this PR is a good balance and another tester has also given a favourable review, saying that he will now be more inclined to make better use of the message feature.

PR #1323, opened 2026-05-04
TX Atten/Tune: fix context menus dismissed on button release on GTK and incorrect Tune power after program re-start

barjac:
On GTK, wxEVT_CONTEXT_MENU fires on button-press for wxStaticText and wxToggleButton. PopupMenu called from within that event is associated with the held button by GTK, so the menu closes the moment the right button is released.

Switch m_txLevelBox, m_txtTxLevelNum and m_btnTogTune from Connect(wxEVT_CONTEXT_MENU) to Bind(wxEVT_RIGHT_UP, lambda), so the menu is only shown after the button is released. The four increment/ decrement buttons keep wxEVT_CONTEXT_MENU as they are unaffected.


barjac:
This fixes two bugs found during extensive testing today:
The first related to the context menu action for activating the per band atten. saving.
The second was more serious as it was applying full TX power during 'Tune' on re-start of the program despite the reading on the screen being correct.

PR #1329, opened 2026-05-07
Release v2.3.0.

tmiw:
Updates version in source tree to enable tagging of v2.3.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,12 +58,12 @@ endif(NOT DEFINED LINUX)
 # Adds a tag to the end of the version string. Leave empty
 # for official release builds.
 if(NOT DEFINED FREEDV_VERSION_TAG)
-set(FREEDV_VERSION_TAG "dev")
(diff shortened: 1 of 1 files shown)


barjac:
Could I ask that you review PR1323 before the next release?
The fix for unintentionally getting a full power 'Tune' output is potentially dangerous.
It has been there since the Tune button was implemented but has only become more serious since the TX atten. per band has been implemented.
The description in the PR explains the bug. I have been using a system built with 1310 and 1323 merged for a week or so now (in Linux only) without problems.

tmiw:
> Could I ask that you review PR1323 before the next release? The fix for unintentionally getting a full power 'Tune' output is potentially dangerous. It has been there since the Tune button was implemented but has only become more serious since the TX atten. per band has been implemented. The description in the PR explains the bug. I have been using a system built with 1310 and 1323 merged for a week or so now (in Linux only) without problems.

Reviewed and merged.

PR #1330, opened 2026-05-07
Reduce likelihood of fullduplex test failing due to audio dropouts on start.

tmiw:
WIP - DO NOT MERGE


tmiw:
Doesn't seem to be doing as promised. Closing.

PR #1338, opened 2026-05-11
Disable use of pffft during audio resampling.

tmiw:
Disables use of the pffft library during audio resampling as it seems to slightly degrade FlexRadio waveform RX audio.


tmiw:
The weird thing is that based on some audio recordings from FreeDV 2.3.0 vs. freedv-flex 2.3.0 that I received, the latter definitely sounds worse (despite both using the same resampling code internally). On top of that, the rade_loss ctest in the aarch64 GH test environment isn't returning results that differ much from 2.2.1. Testing locally with and without this change indicates that making this change does improve audio quality, so I'm going to check in with the person who reported the problem and see if they notice improvements as well.

EDIT: Forgot to mention, if this doesn't help the next step is to revert the change to r8brain and go back to using libsamplerate. The FlexRadio waveform uses a processing pipeline that is a subset of freedv-gui's, so there's not much more it could be.

tmiw:
Merging at risk due to multiple complaints about the FlexRadio waveform audio quality. I'll create a followup PR if this still doesn't fully resolve said complaints.

PR #1353, opened 2026-05-18
Fix missing Hamlib defines when building from source

barjac:
## Summary

- When `libhamlib-dev` is absent and FreeDV builds Hamlib from source via `cmake/BuildHamlib.cmake`, `HAMLIB_USE_FRIENDLY_ERRORS` and `HAMLIB_CONST_WORKAROUND` are never defined
- The system Hamlib path already sets these via a version check, but the source-build path had no equivalent
- Without `HAMLIB_USE_FRIENDLY_ERRORS`, the `rig_get_conf()` call in `HamlibRigController.cpp` is used instead of `rig_get_conf2()`, which causes a `-Werror=deprecated-declarations` build failure against Hamlib 4.7.1
- Since `BuildHamlib.cmake` always fetches 4.7.1, both defines are unconditionally valid

## Test plan

- [x] Remove `libhamlib-dev`, rebuild FreeDV — confirm build succeeds with Hamlib built from source
- [x] Verify `rig_get_conf2()` path is taken (no deprecated-declarations warning)


barjac:
Build succeeds and runs with this change. With 4.7.1 it can now talk to the radio using port 4532.

tmiw:
Interesting that this is needed, I would have expected the GH builds (especially the AppImage ones) to have started failing by now. The changes look reasonable, though, so I'll merge when I'm on a computer.

PR #1366, opened 2026-05-26
Add Time-Out Timer (TOT) capability to FreeDV.

### LINKED ISSUE #1327: [Feature Request] Add Time-Out Timer (TOT) option
### Contact Details

N3FL, K8CHY, KC5SW, W7YC

### Feature for this project

*Note: Submitting on behalf of W7YC based on email request.*

#### Description:

Provide a timeout function to disable the radio’s transmit capability if the transmitter is active for more than a preset time.

#### Reason for request:

During nets with many operators participating, it is easy for someone to accidentally bump the spacebar on their keyboard and key up the FreeDV client without immediately being aware. This usually blocks (jams) the frequency so that no other users can communicate until the offending station has unkeyed.

This scenario happened recently during one of the FreeDV Activity Nets on Tuesday evenings in the US.

#### GUI Implementation:

Provide a checkbox in the Tools…Options…Rig Control tab, labeled “Enable Timeout Timer” with a popup Tooltip that says “Right Click to Set Timeout Timer Value”

* Enabled by default:  Yes
* Default time: 3 minutes*

\* Note: For FM Repeater operation, the standardized timeout timer period is 3 minutes. This seems a reasonable default for FreeDV also.

#### Reset Procedure:

If the timeout period is exceeded, a visual and audible warning should be generated to alert the operator, and transmit should be disabled until the warning is cleared by the operator.

#### Additional Notes

Stuck mic captured on net livestream: https://www.youtube.com/watch?v=65HtytbtYCk (beginning at ~6:05, with discussion going to ~15:45)

### VHF/UHF

No

### Deprecated features

No

### Non-standard compiler

No

### Legal worldwide

Yes

### Implementation assistance

Not specified.

### Support assistance

Willing to test feature implementation.

tmiw:
Resolves #1327 by adding support for a Time-Out Timer (TOT) during TX, similar to what's already done for FM repeaters. Per the original feature request, this defaults to being enabled and three minutes (180 seconds) in duration.

Disclaimer: Implemented with the assistance of Claude Code.


TreeTopHam:
Testing 2.3.1-dev-87b8 on Linux Mint and Windows 11.
No issues seen so far.

Kudos to Claude Code!

barjac:
Err.. Really enabled by default?

IMHO this should be opt-in not opt-out.

I have yet to hear anyone seriously asking for this, (except here) although it has been discussed, once as I recall and it was years ago.

I will test it if only to hear/see what the local timeout warning is like, or does it just drop you mid sentence?

tmiw:
> I have yet to hear anyone seriously asking for this, (except here) although it has been discussed, once as I recall and it was years ago.

Per the referenced feature request, someone accidentally transmitted and it actually interfered with the weekly FreeDV net for a bit until they were able to turn off TX. It's also why it probably can't default to being off--otherwise it'd be unable to stop instances where someone accidentally transmits or transmits longer than they'd like. 

> I will test it if only to hear/see what the local timeout warning is like, or does it just drop you mid sentence?

Yeah, it just effectively injects a click on the PTT button after the timeout fires.

barjac:
I still think this should be opt-in.

We discussed this on air last night and we agreed that this should certainly not default to on.

As a user commented to me, timeouts are for shared infrastucture like repeaters not for QSOs on HF.

TreeTopHam:
@barjac writes:
>As a user commented to me, timeouts are for shared infrastucture like repeaters not for QSOs on HF.
>

Yes, this is why timeouts were initially implemented on FM repeaters.

**However**, a use case recently came up with FreeDV and **large nets** where a user inadvertently hit the space bar on his keyboard and keyed up his radio unknowingly.  This then effectively jammed the net operating frequency until the offending station de-keyed.

All who were on the FreeDV activity net during this event **unanimously** voted to add the TOT feature, **including** the station who caused the problem!

I've seen this happen twice now - once as mentioned in #1327 (you can watch how this went down because it was captured in a live stream video; the link is listed in the FR) and again yesterday when an operator keyed up the FDV client for 15 - 20 minutes but was unaware until I texted him.

For casual QSOs between one or two stations, the TOT is not as important as it is when we have large groups of stations all camped out on the same frequency.

segue -

I had at first proposed an audible TOT warning, but in listening to how @tmiw has implemented it, there is no real need for the audible warning. The FDV client just unkeys itself. Keeping it simple :) Thanks @tmiw

tmiw:
> I had at first proposed an audible TOT warning, but in listening to how @tmiw has implemented it, there is no real need for the audible warning. The FDV client just unkeys itself. Keeping it simple :) Thanks @tmiw

What do you think about having a pop-up that comes up after it unkeys saying that the TOT fired?

TreeTopHam:
>What do you think about having a pop-up that comes up after it unkeys saying that the TOT fired?
>

That would be useful to reinforce what has taken place, and that the unkey was not a 'bug' in FDV.

Segue -

The TOT will be useful to all operators because it will help protect the finals in the radio's PA from damage in the event of an inadvertent keyup.

barjac:
> > What do you think about having a pop-up that comes up after it unkeys saying that the TOT fired?
> 
> That would be useful to reinforce what has taken place, and that the unkey was not a 'bug' in FDV.
> 
> Segue -
> 
> The TOT will be useful to all operators because it will help protect the finals in the radio's PA from damage in the event of an inadvertent keyup.

Please don't get the impression that I am against this feature, I am not.

I think it has been conceived as a knee-jerk reaction to a single event, without consideration of all the other use cases.

This would not only be a safety feature, it has more regular use in avoiding over-long overs for those who need it.

I think that a pop-up warning (and maybe a short sound) should appear 15sec BEFORE the timeout and include a 'reset' button to allow the current 'over' to be completed normally after the 'reminder'.
I have incorporated this 'warning before timeout' elsewhere in the past for local use when using remote systems. It works well since there is no disruption to the conversation and others are unaware. Rather that than an abrupt drop of transmission.

Since all the stations who witnessed the mistakes are in the same net then surely they would all enable this feature when it is available.

I still think it should not be enabled by default.

TreeTopHam:
>I think that a pop-up warning (and maybe a short sound) should appear 15sec BEFORE the timeout and include a 'reset' button to allow the current 'over' to be completed normally after the 'reminder'.
>

The '**warning before timeout**' is a great idea to allow a smoother transition for a long-winded operator. And for those who prefer to talk longer, they can increase the TOT interval to suit their needs.

Major US manufacturers ship their commercial 2-way radio units with the TOT feature pre-activated. Enabling this feature by default is standard industry practice for commercial mobile and portable two-way radios.

tmiw:
A message box probably wouldn't be a good idea for an "about to timeout" alert. What if the PTT button turned orange or something? Would that be sufficient indication that the TOT is about to fire? If so, how would one extend/reset the timeout if needed?

barjac:
> A message box probably wouldn't be a good idea for an "about to timeout" alert.

Why not? That seems ideal to me, allowing a reset of the timer so the user could continue for a while and complete the over.

What if the PTT button turned orange or something? Would that be sufficient indication that the TOT is about to fire? If so, how would one extend/reset the timeout if needed?

Well in that case you could not, unless clicking the orange PTT would extend the timeout.

A short 'beep beep' in the RX audio chain would help too.

tmiw:
I had Claude add a new box to the main window that appears <= 15 seconds away from timing out indicating the time remaining. This also has a button with a red background to allow 60 seconds to be added to the existing TX. Hopefully this helps address the concerns raised in this PR.

TreeTopHam:
>I had Claude add a new box to the main window that appears <= 15 seconds away 
>

When the warning box appears, how about flashing  the "PTT" button on/off (i.e. Red/no Red) at a a ~2Hz rate to call attention to the TOT issue? 

The Claude box appearing by itself doesn't get my attention like it needs to.

tmiw:
> > I had Claude add a new box to the main window that appears <= 15 seconds away
> 
> When the warning box appears, how about flashing the "PTT" button on/off (i.e. Red/no Red) at a a ~2Hz rate to call attention to the TOT issue?
> 
> The Claude box appearing by itself doesn't get my attention like it needs to.

Would flashing the "extend timeout" button also work? Might be cleaner logic on my end, anyway.

barjac:
Most timeout warnings that I have used incorporate sound.
I have one that speaks "Time" in the RX audio channel but a few beeps would be adequate here, to coincide with the visual warning. Flashes in colour are generally much more noticeable.

TreeTopHam:
@tmiw 
>Would flashing the "extend timeout" button also work? Might be cleaner logic on my end, anyway.
>

Yes, flashing the "extend timeout" button might be better. 
A flashing "PTT" button might give the wrong impression - that the PTT is intermittent or something like that.

@barjac 
>...a few beeps would be adequate here, to coincide with the visual warning. Flashes in colour are generally much more noticeable.
>

Audible beeps to coincide with the flashing button would be a great idea.

tmiw:
Not sure about the audio beeps yet but the Extend Timeout button now flashes 2x/second. Any tweaks needed there at least?

barjac:
> Not sure about the audio beeps yet but the Extend Timeout button now flashes 2x/second. Any tweaks needed there at least?

Yes, this needs to be a pop-up box not a new main GUI widget.
As it is now it rearranges the whole main GUI to accommodate the warning box.

TreeTopHam:
@tmiw :
>...the Extend Timeout button now flashes 2x/second. Any tweaks needed there at least?

The Extend TOT button functionality looks and functions very nicely!

@barjac:
>Yes, this needs to be a pop-up box not a new main GUI widget.
As it is now it rearranges the whole main GUI to accommodate the warning box.

I agree with @barjac about a pop-up box.
For small vertical GUI sizing where the PTT button is near the bottom, the Extend button bumps out the GUI when it triggers - see below:
.
.

Flashing Extend PTT Warning Resizes Small Screens

tmiw:
I have some concerns about making it a separate window. For example, it might have to be a modal window to keep it from not inadvertently being hidden, and I'm not sure how TOT will behave once it times out. Maybe beeping will become more necessary here?

tmiw:
I had Claude revert the changes to the main window and instead, create a new dialog box. How is it so far? The only thing it doesn't include is the flashing "Extend Time" button, but that might not be necessary when a separate window comes up.

barjac:
Adding an image here to help users who cannot build this to test.

image

This now brings another potential problem that I admit I had not previously considered.

The pop-up grabs focus so it has to be used to end or extend the transmission.
The PTT button cannot be used.
I wonder if this then precludes the use of a foot switch or the space bar etc. to end 
the transmission before the timeout?
This could be an issue for many users.
The X close in the dialogue also closed FreeDV!

Looking for alternative approaches I had a word with Claude :)

wxPopupWindow is designed specifically for transient overlays that don't take focus — it's what wxWidgets uses internally for things like autocomplete dropdowns. 
Unlike wxDialog, it leaves keyboard focus and mouse events on the main window, so the PTT button, spacebar and any external triggers (footswitch etc.) would continue to work normally during the warning period.
It also avoids the layout rearrangement that the earlier in-window approach caused.
  
The "Stop Transmitting" and "Extend by 60s" buttons could still be included — they just sit in the popup rather than a dialog. Positioned near or over the PTT button it should be hard to miss.

TreeTopHam:
I like the dialog box! Very nice.

@barjac mentioned:
>The pop-up grabs focus so it has to be used to end or extend the transmission.
The PTT button cannot be used.

I also wondered about possible focus conflicts being an issue...

**However**, I tested dev-3822 on both Linux Mint and Windows 11 and **both** builds worked well for me. I did not have an issue with the pop-up dialog window having focus and preventing the main PTT button from working. I could stop transmitting by clicking on either 'Stop Transmitting' or on 'PTT' or by hitting the 'space bar' regardless of which window had focus.

**BUT** - the only thing that was an issue was closing the pop-up dialog box with the 'X' close button, which shut down FreeDV altogether : (

With the pop-up dialog, I don't know that a flashing "Extend Time" button would be as critical as before, but anything flashing will get my attention much faster than a static pop-up window.

tmiw:
On macOS, closing the warning dialog was resulting in a segfault. The latest commit should fix that. Currently FreeDV isn't taking any action other than closing the window; is that correct or should it i.e. immediately stop TX?

EDIT: foot switches should still work as that's handled outside of the GUI code, but Space doesn't seem to. I'll look into that further.

tmiw:
OK, Space should work to end TX while the TOT window is up.

barjac:
> Currently FreeDV isn't taking any action other than closing the window; is that correct or should it i.e. immediately stop TX?
> 

I would imagine that the only reason for hitting X may be in frustration at being interrupted (again?) by the timeout warning and a misguided attempt to cancel it completely.
Maybe it could oblige on that occasion, rather than closing the dialogue and timing out anyway, which I guess is not what the user was attempting to achieve.

An extra button 'Abort' may be a better solution in the warning dialogue. If a user is normally happy with his 3min timeout but has a long story to tell, he could hit the button to not be interrupted again during this over. The X could also trigger the same action as the button.

barjac:
During a QSO last night with another station (both using this branch to test) I was looking at another screen while talking in TX and completely missed the timeout warning.
A sound notification is essential.

TreeTopHam:
>Currently FreeDV isn't taking any action other than closing the window; is that correct or should it i.e. immediately stop TX?


I'm fine with just closing the window after 'X' because the TOT function is still active and will stop the transmit when it times out.

If we don't have a visual flashing alert then a sound notification becomes an essential feature.

**EDIT**: What if we eliminate the 'Stop Transmitting' button altogether. The user can stop transmitting at any time by clicking the PTT button or pressing space bar, so this button is redundant. What if we rename the 'Extend by 60s' to something like 'Extend this PTT' and change the functionality of this button press so that it _cancels the time out interval for the current transmission_. That way the dialog box won't keep popping up and the user can talk as long as needed for the current 'over'. After the dekey, things would reset back to the normal TOT interval for the next transmit.

New PTT

barjac:
> **EDIT**: What if we eliminate the 'Stop Transmitting' button altogether. The user can stop transmitting at any time by clicking the PTT button or pressing space bar, so this button is redundant. What if we rename the 'Extend by 60s' to something like 'Extend this PTT' and change the functionality of this button press so that it _cancels the time out interval for the current transmission_. That way the dialog box won't keep popping up and the user can talk as long as needed for the current 'over'.

Which would defeat the safety feature.

All that is needed is a 'Reset' timeout button that appears 15 seconds before timeout along with an audible warning sound.

tmiw:
Audio beep support added via Claude, but it doesn't sound quite right in my own testing. I'll have to investigate more when I get time.

TreeTopHam:
The '**Stop Transmitting**' button is redundant. The TOT count down is still active regardless of the 'Stop Transmitting' button. This button should be eliminated.

If we used a single **'Reset'** button, should it reset the TOT interval by 60 seconds only or should it reset to the full TOT interval (i.e. 180 sec or whatever time has been entered by the user)?

barjac:
> The '**Stop Transmitting**' button is redundant. The TOT count down is still active regardless of the 'Stop Transmitting' button. This button should be eliminated.
> 
> If we used a single **'Reset'** button, should it reset the TOT interval by 60 seconds only or should it reset to the full TOT interval (i.e. 180 sec or whatever time has been entered by the user)?

I would say a full reset to the full time set in the config.
Wafflers get a warning but are not rushed into finishing in 30s to avoid a second warning.
No impact on the safety aspect.
In testing we did find the 30s too short and rather annoying.

barjac:
> Audio beep support added via Claude, but it doesn't sound quite right in my own testing. I'll have to investigate more when I get time.

That does work but it keeps repeating, it is only needed once as the warning pops up.
The tones are distorted so 'not quite right' as you say.
Far too loud in headphones set up for normal QSO listening.

A morse '5' sent at about 15wpm & 700Hz once only would be an adequate alarm sound without resorting to an audio file.

TreeTopHam:
- >I would say a full reset to the full time set in the config.

Yes, I agree: reset to the full TOT value set in the config after the 'Reset' button is pressed

- >A morse '5' sent at about 15wpm & 700Hz once only would be an adequate alarm sound without resorting to an audio file.

Agree. Great suggestion.

tmiw:
OK, performed the following changes:

1. Made warning beep only emit once and cut the volume in half. If we still want the Morse code beeping, I can go back and look into it further.
2. Did some backend work to make the warning beep sound better. Let me know if it sounds better to you guys, it definitely does for me.
3. Removed the stop TX button from the warning popup window.
4. Updated the Reset button to reset to the configured timeout, not just add 60s.

Let me know how the above works for you guys.

TreeTopHam:
>OK, performed the following changes:
>
>1. Made warning beep only emit once and cut the volume in half. If we still want the Morse code beeping, I can go back and look into it further.

- The reduced volume is much easier on the ears - thanks! 
- A single beep actually works quite nicely. It's enough to get my attention.
- I'm okay with leaving it as-is. The 5-beep Morse will even be more attention grabbing, and I'm okay with that also.

>2. Did some backend work to make the warning beep sound better. Let me know if it sounds better to you guys, it definitely does for me.

- Beep sounds much better to me on Windows 11. Still sounds rough on Linux Mint.

>3. Removed the stop TX button from the warning popup window.

- Popup window now looks clean and to the point.

>4. Updated the Reset button to reset to the configured timeout, not just add 60s.

- This works as intended on both Windows 11 and Linux Mint.

>
>Let me know how the above works for you guys.


Except for the rough audio on Linux Mint everything else looks good. Nice job!

tmiw:
OK, hopefully the beep audio is a bit better with the latest.

Also, is 180s still a good default value? I know there were questions/concerns around having TOT be enabled by default, so would a longer default help with that (and/or the changes already made in this PR to i.e. reset to configured timeout)?

barjac:
I have no preference for the default timeout as I won't be using it :)

In Linux (pipewire/wireplumber) there is still bad distortion of (or on top of) the warning signal.

TreeTopHam:
The audio beep sounds now sounds good on Linux Mint (as well as Win11).

I'm good with the 180s default value for TOT. The user can always change it if desired.

tmiw:
> In Linux (pipewire/wireplumber) there is still bad distortion of (or on top of) the warning signal.

Does it get better if you turn on "Monitor transmitted audio"? I noticed before the previous commit that it was, anyway.

barjac:
> > In Linux (pipewire/wireplumber) there is still bad distortion of (or on top of) the warning signal.
> 
> Does it get better if you turn on "Monitor transmitted audio"? I noticed before the previous commit that it was, anyway.

Mooneer,
Please keep this as it is just now as I have spent a long time with Claude and have this working with clean audio and 5 beeps (morse 5). It has been a marathon! I will add a new PR which will sit on top of this one for you to test.
See #1382

tmiw:
Please see if this alternate approach works for emitting '5' instead of a single beep. It sounds okay on macOS, anyway.

Looking at @barjac's PR I'm seeing a few other things that I didn't implement with my last couple of commits (like the precharge thing). We can discuss that further if needed.

EDIT: I got an assertion failure on Linux. Will investigate and update the PR if needed.

barjac:
> Please see if this alternate approach works for emitting '5' instead of a single beep. It sounds okay on macOS, anyway.
> 
> Looking at @barjac's PR I'm seeing a few other things that I didn't implement with my last couple of commits (like the precharge thing). We can discuss that further if needed.
> 
> EDIT: I got an assertion failure on Linux. Will investigate and update the PR if needed.

#1366 as it was earlier is breaking the Tune button - I just discovered tonight in testing.
There is no tune output at all with PR1366 installed. Master is fine. Bed time! Good Night!

tmiw:
> #1366 as it was earlier is breaking the Tune button - I just discovered tonight in testing. There is no tune output at all with PR1366 installed. Master is fine. Bed time! Good Night!

Should be fixed now.

TreeTopHam:
- Tune button is fixed. 
- Morse (5) works & sounds very nice! 

Tested on Linux Mint & Win11
Let's keep the Morse (5) going forward :)

barjac:
Tune seems fixed OK and the morse5 is working better.
In #1382 the individual dits were ramped up and down rather than hitting full level instantly and after just comparing the two, here I hear the equivalent of rather jarring key clicks that are not there in #1382. Not major but would be nice to remove the clicks in the same way. Otherwise looking great now. :)

From Claude:

● The fix is minimal — just replace the raw cosine line in execute() with an envelope-shaped version. Here's a snippet to drop in:

  In the constructor, add one line after samplesToGenerate_ is initialised:
  , ramp_((sampleRate * 5) / 1000)   // 5 ms Hann ramp

  And add int ramp_; to the header.

  Then replace the tone generation block in execute():
  // was:
`  outPtr[index] = TONE_AMPLITUDE * cosf((2.0 * M_PI * frequency_ * sampleCtr_) / sampleRate_);`
  
  // replace with:
```
  {
      double env;
      if (sampleCtr_ < ramp_)
          env = 0.5 * (1.0 - cosf(M_PI * sampleCtr_ / ramp_));
      else if (sampleCtr_ >= samplesToGenerate_ - ramp_)
          env = 0.5 * (1.0 + cosf(M_PI * (sampleCtr_ - (samplesToGenerate_ - ramp_)) / ramp_));
      else
          env = 1.0;
      outPtr[index] = (short)(TONE_AMPLITUDE * env * cosf((2.0 * M_PI * frequency_ * sampleCtr_) / sampleRate_));
  }   

```
  
  That's a Hann ramp over the first and last 5ms of each dit, which eliminates the key clicks without any other changes to the architecture.

barjac:
Testing this branch as it is now, RX playback is broken.
The received audio is played in bursts about 4 per second and at slow speed.
I suspect the decode is working as a very short decoded recording sounded OK, just the live playback is broken.
Switching to master all is fine.
I only tested TX and timeout this morning. Testing on air tonight I hit the RX issue.
This was not the case last night before the last changes.

tmiw:
> Testing this branch as it is now, RX playback is broken. The received audio is played in bursts about 4 per second and at slow speed. I suspect the decode is working as a very short decoded recording sounded OK, just the live playback is broken. Switching to master all is fine. I only tested TX and timeout this morning. Testing on air tonight I hit the RX issue. This was not the case last night before the last changes.

I hear that here too, even with stored file playback. The latest commit should hopefully help with that.

barjac:
Yes the key clicks have gone and the RX audio seems OK now :)
Oh and the Tune still works :)
Will do more testing tomorrow (also on RPi4) but this looks finally sorted, fingers crossed. ;)

barjac:
All OK on the Pi4.
Just needs the default state to be OFF and this is good to go for me.

If you are not convinced then maybe a poll in the mailing list would settle it :)

TreeTopHam:
Key clicks gone; Morse(5) audio very clean now :)   
Thanks to all who made this possible!

I'm good with this feature as it now stands.
.
>Just needs the default state to be OFF and this is good to go for me.

Deploying TOT with default state OFF defeats its purpose. 

With TOT we are trying to achieve a good balance between an unintended TX and normal voice 'overs.'

Since unintended transmissions are not deliberate, having the protection feature enabled is essential.

Advanced users and rag chewers can extend the TOT interval to suit their needs or even disable TOT altogether, but disabling should be the exception and not the rule.
.
>If you are not convinced then maybe a poll in the mailing list would settle it :)

Perhaps run it past the PLT members for their opinion.
.
- Segue -

On Linux Mint, the GUI remains in TX state for ~1s  after the PTT button is clicked to de-key the radio.  Pressing space bar to de-key is instantaneous. This delay is not seen in Windows.

I mention this here because the delay from mouse clicking the PTT after the TOT warning fires causes some user confusion when de-key isn't immediate and can result in multiple mouse clicks on the PTT button, which gets the intended timing out of sync.

EDIT: The ~1s delay is not due to this PR. It also exists in 2.3.1 on Linux Mint. It becomes more evident now only because of the addition of the TOT warning.

barjac:
> On Linux Mint, the GUI remains in TX state for ~1s after the PTT button is clicked to de-key the radio. Pressing space bar to de-key is instantaneous. This delay is not seen in Windows.
> 
> I mention this here because the delay from mouse clicking the PTT after the TOT warning fires causes some user confusion when de-key isn't immediate and can result in multiple mouse clicks on the PTT button, which gets the intended timing out of sync.
> 
> EDIT: The ~1s delay is not due to this PR. It also exists in 2.3.1 on Linux Mint. It becomes more evident now only because of the addition of the TOT warning.

The delay is always there, it is intrinsic in the way the system works. Once the system knows that you want to de-key by pressing the space bar or hitting the PTT button a chain of events is started which takes time to complete before the radio is taken out of transmit.

The difference you see is in the way that the colour change of the PTT button is triggered which really should be consistent for all PTT trigger methods.

My choice would be to change the method used by the space bar to match the PTT button so the button colour correctly reflects the actual state of the radio.
The user should expect this delay while encoding completes, buffers are emptied and the end of message callsign is sent.

Two possible alternative ways to avoid the confusion during the delay:
1. There is a guard against multiple triggers of the PTT in place but only for the tty0tty RS232 PTT option. This could be extended to cover all PTT trigger methods, so that a second 'confusion click' does no harm if it occurs during the de-key roll-down period.

2. Have all PTT methods instantly change the PTT button to amber to indicate that the de-key roll-down has started, this would be extinguished when the radio finally de-keys.

I prefer 2 however both would do no harm.

tmiw:
Honestly, the PTT button itself should be disabled during the transition period to avoid clicks during the transition. I'll make sure that's already happening and update the PR as needed.

barjac:
> Honestly, the PTT button itself should be disabled during the transition period to avoid clicks during the transition. I'll make sure that's already happening and update the PR as needed.

That was added but only for the RS232 PTT input. I agree that it should apply to all.

Other ideas came up in our discussions tonight: 
Add the time left before timeout to the right of the PTT button label, updated every second, if the timeout is active.
If the default has to be ON for the safety aspect then set the default to a higher value like 12 mins.

TreeTopHam:
>Honestly, the PTT button itself should be disabled during the transition period to avoid clicks during the transition. I'll make sure that's already happening and update the PR as needed.

That seems like a good solution. No other embellishment involving PTT needed. Thanks.

>If the default has to be ON for the safety aspect then set the default to a higher value like 12 mins.

12 mins is a long time to have the frequency jammed by a rouge PTT, especially during a net with many participants.

As mentioned earlier, 3 min has become the industry standard in commercial 2-way radio. For HF, this also seems like a reasonable balance between average user 'overs' and disruptive channel jamming. Users can always adjust for their own needs if they prefer.

HF rag-chewers will gravitate to their own 'sweet spot' for TOT value, each matching their own personal speaking profiles.

tmiw:
I made some tweaks to the PTT button handling; hopefully that works out better.

Re: default TOT time--apparently 3 minutes used to be an FCC requirement in the US: https://ham.stackexchange.com/questions/20573/when-was-the-fcc-part-97-requirement-for-a-time-out-timer-removed-from-repeaters. Maybe we can start with that and if people complain we can bump that up?

barjac:
My final comments on this subject.
FreeDV is not a repeater. If a net is disrupted for 12 mins its not the end of the world. The net incident that started this debate had not even started when it happened. The longer the timeout the more the embarasment for the culprit which is more likely to stop it happening again. I know nobody yet who would leave this feature enabled. I have already been asked where to turn it off.

tmiw:
I brought the question up to the wider FreeDV community to see what they have to say: https://groups.google.com/g/digitalvoice/c/hdw8yaUVWDw. We'll hold off on merging for now until we hear more from other users.

TreeTopHam:
>I made some tweaks to the PTT button handling; hopefully that works out better.

Yes, PTT behavior on de-key is now consistent between OS platforms (Linux Mint, Win11). 
Thanks :)

barjac:
In the latest the 'PTT' text on the button changes from black to maybe grey and is illegible during the roll-down.

tmiw:
> In the latest the 'PTT' text on the button changes from black to maybe grey and is illegible during the roll-down.

I'm not sure there's a way to override the text color during the time that the button is disabled. It might not necessarily be a good idea to, either, if the idea is to let users know the button can't be pushed. Maybe if the color is reset during the transition and only set to the actual state after completion/reenable?

tmiw:
Updated PTT logic to clear background color while transitioning back to RX, hopefully that helps with readability. I also updated the user manual to talk about this feature.

Speaking of which, people on the mailing list seem to generally either be supportive of the TOT feature or don't mind it as long as it can be configured/disabled. I think we'll leave it defaulted to enabled and 180s and revisit if we get complaints after the fact. I'll merge once we confirm that the most recent commits are okay.

barjac:
If the b/g is cleared then there is no indication that the rig is still in TX which is wrong.
As I suggested previously a b/g colour change during the roll-down would indicate that the roll-down had actually started and avoid the confusion that was complained about by a user previously.

In testing, this PR now has a few strange effects (using a regular breeze light theme):
On hitting PTT to TX the button b/g flickers white > grey > red.
The normally black 'PTT' text changes to grey when the timeout warning dialogue appears. (seems only with older wxGTK)
The warning sound starts about 1 second after the warning dialogue pop-up.
I don't see any of these effects with PR#1382.

TreeTopHam:
>Updated PTT logic to clear background color while transitioning back to RX...

@tmiw - Will this update produce the same behavior for the PTT button during key/de-key as in Windows Release 2.3.1?

In that release, the red PTT background color immediately changes back to the 'off' state color when the PTT button is clicked or space bar is pressed.

Immediately clearing the red background color gives the user immediate visual feedback that the de-key request was received by the GUI. This makes the app seem less sluggish than waiting for the TX roll-down to complete even though it's only ~1.5 sec.

**Other quirks:**

- I also notice the ~1s delay in the Morse(5) tones that has crept in along the way.

- If 'Tune' is activated,  PTT can also be activated by pressing space bar.  The 'Tune' background color remains red after the radio de-keys. Seen on both Win11 and Linux Mint.

tmiw:
OK, I updated the PR to do the following:

1. Instead of clearing the background for the PTT, use an intermediate color (currently a pinkish-red, but we can discuss further as long as it's something that won't require a foreground color change--that doesn't seem to work so well on macOS at least).
2. Tweaked the audio pipeline to reduce the lag for the warning beeps.
3. Fixed the bug with Tune mode and spacebar triggering PTT.

@TreeTopHam: This should work regardless of platform, but let me know if you find differences.

barjac:
There is flicker on PTT start and the colour on roll-down does not change noticeably.
The 'PTT' text changes to grey on roll down.

Please review/test this patch which improves this in linux:
https://github.com/barjac/freedv-rade-build/blob/main/patches/bcj-ptt-amber-rundown.patch

tmiw:
> There is flicker on PTT start and the colour on roll-down does not change noticeably. The 'PTT' text changes to grey on roll down.
> 
> Please review/test this patch which improves this in linux: https://github.com/barjac/freedv-rade-build/blob/main/patches/bcj-ptt-amber-rundown.patch

Going to gray is basically required if we have the button disabled during transition. I'm not sure how I feel about not disabling it and using other mechanisms for preventing duplicate clicks.

Re: color change, maybe we can just change the color for TX->RX and not RX->TX? The former is fairly instant in comparison to the latter, after all, which may be why you're seeing the flickering.

barjac:
> > There is flicker on PTT start and the colour on roll-down does not change noticeably. The 'PTT' text changes to grey on roll down.
> > Please review/test this patch which improves this in linux: https://github.com/barjac/freedv-rade-build/blob/main/patches/bcj-ptt-amber-rundown.patch

> Going to gray is basically required if we have the button disabled during transition.
I'm not sure how I feel about not disabling it and using other mechanisms for preventing duplicate clicks.

This does disable the button while amber during the roll-down.

> 
> Re: color change, maybe we can just change the color for TX->RX and not RX->TX?

This does only change the colour for TX-> RX.

> The former is fairly instant in comparison to the latter, after all, which may be why you're seeing the flickering.

This fixes the flickering by using button down to change to red on pressing PTT rather than button release, since any start of TX delays are minor.

This patch applies cleanly on top of #1366 - did you test it?

tmiw:
With @barjac's patch on macOS:

https://github.com/user-attachments/assets/0801918e-ad66-4525-a2e4-207182c62f34

(I double-clicked the PTT button to stop transmitting and the button's appearance changed on the second click. We can go with this patch but I'm hoping to find an alternative that would allow the button to be disabled at the GUI framework level.)

tmiw:
I ultimately just decided to apply @barjac's patch with minor updates. Combined with a double-click handler, it seems to be okay on macOS at least. @TreeTopHam, how's it look for you?

barjac:
Perfect for me!
Tested in x86_64 and aarch64 linux.

TreeTopHam:
@tmiw I like the changes. Everything works smoothly. Thanks to all!
Tested on Linux Mint & Win11

PR #1368, opened 2026-05-29
Remove freedv-backend and freedv-integrations code from freedv-gui

tmiw:
This PR does the following:

* Removes shared files that have been moved to [freedv-backend](https://github.com/tmiw/freedv-backend). This is due to these files having a different license (BSD) than freedv-gui (LGPL).
* Removes Flex and KA9Q integrations from this repo. These have been moved to the [freedv-integrations](https://github.com/tmiw/freedv-integrations) repo.


Tyrbiter:
Current master with this PR merged gives me this error:

[ 12%] Performing download step (git clone) for 'build_rade'
Cloning into 'build_rnnoise'...
Cloning into 'rade_src'...
In file included from /home/bdm/git/freedv-gui/src/sox/sox_i.h:17,
                 from /home/bdm/git/freedv-gui/src/sox/biquad.h:22,
                 from /home/bdm/git/freedv-gui/src/sox/biquad.c:18:
/home/bdm/git/freedv-gui/src/sox/sox.h:27:10: fatal error: sanitizers.h: No such file or directory
   27 | #include "sanitizers.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [src/sox/CMakeFiles/fdv_sox.dir/build.make:79: src/sox/CMakeFiles/fdv_sox.dir/biquad.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1316: src/sox/CMakeFiles/fdv_sox.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

Not sure exactly where this came from, but it has broken the build completely.

tmiw:
> Not sure exactly where this came from, but it has broken the build completely.

Any difference if you do a full rebuild? Per GH it should be building fine on Linux.

Tyrbiter:
> > Not sure exactly where this came from, but it has broken the build completely.
> 
> Any difference if you do a full rebuild? Per GH it should be building fine on Linux.

My build script does this:

#!/bin/sh
rm -rf build_linux/ && ./build_linux.sh

Is that not enough?

I can't find a sanitizers.h file in my local git master source tree.

Tyrbiter:
OK, I cloned the freedv-gui tree again, this time it built.

I have no idea why it didn't work after a git pull --all with the previous tree.

Comparing old and new git trees shows no differences at all. I have no idea what happened.

And after renaming the old git tree for the comparison, I tried building in that directory. It worked.

You can forget worrying about this, it's some sort of weird local glitch I suppose.

Tyrbiter:
Further investigation shows that it works the first time I run build_linux.sh but then fails on a second run unless I delete the tree and re-clone it from github.

I don't know what could cause this, presumably something in the build process messes with the include paths and changes something when I build?

I have just recompared a new cloned tree with one that has been built from, other than the build_linux subdirectory being present I can't find any differences, I use meld as a diff tool.

tmiw:
> Further investigation shows that it works the first time I run build_linux.sh but then fails on a second run unless I delete the tree and re-clone it from github.
> 
> I don't know what could cause this, presumably something in the build process messes with the include paths and changes something when I build?
> 
> I have just recompared a new cloned tree with one that has been built from, other than the build_linux subdirectory being present I can't find any differences, I use meld as a diff tool.

Just tried running build_linux.sh twice on Ubuntu 24.04, no errors compiling here. Maybe it's something specific to Fedora?

Tyrbiter:
Could be, I will try to pin it down, as usual Fedora's package update rate can cause problems of various sorts.

Tyrbiter:
I have got my F44 installation fully up to date and now have a new kernel and managed to fix a weird virtual desktop problem too. However this build failure continues for a reason I can't yet determine. I also had a mistake in my command line for running with ASAN so when the negative SNR problem occurred I didn't get any output, that's also fixed.

A temporary fix for building is to copy sanitizers.h from the src/util/ directory in 2.3.1 to src/sox in master or whatever branch I am building. This has me back to being able to build and run new versions until I can work out why the include file search path is either not working or is finding a file that is incompatible for some reason.

Hmm, so the only system installed sanitizers.h somewhere in /usr/lib64/... are not the same as the file you had added to src/util/

I'm trying to work out if there is another package that F44 has that installs another version of sanitizers.h as I don't know where you got yours from.

Fun this eh?

Just noticed that some of the files in my include directories are called Sanitizers.h so there is a case issue in the search path

tmiw:
> I have got my F44 installation fully up to date and now have a new kernel and managed to fix a weird virtual desktop problem too. However this build failure continues for a reason I can't yet determine. I also had a mistake in my command line for running with ASAN so when the negative SNR problem occurred I didn't get any output, that's also fixed.
> 
> A temporary fix for building is to copy sanitizers.h from the src/util/ directory in 2.3.1 to src/sox in master or whatever branch I am building. This has me back to being able to build and run new versions until I can work out why the include file search path is either not working or is finding a file that is incompatible for some reason.
> 
> Hmm, so the only system installed sanitizers.h somewhere in /usr/lib64/... are not the same as the file you had added to src/util/
> 
> I'm trying to work out if there is another package that F44 has that installs another version of sanitizers.h as I don't know where you got yours from.
> 
> Fun this eh?
> 
> Just noticed that some of the files in my include directories are called Sanitizers.h so there is a case issue in the search path

Maybe we need to rename it to freedv_sanitizers.h instead. I think I can do that change quickly in another set of PRs (both freedv-backend and freedv-gui as sanitizers.h is in the former).

Tyrbiter:
> > I have got my F44 installation fully up to date and now have a new kernel and managed to fix a weird virtual desktop problem too. However this build failure continues for a reason I can't yet determine. I also had a mistake in my command line for running with ASAN so when the negative SNR problem occurred I didn't get any output, that's also fixed.
> > A temporary fix for building is to copy sanitizers.h from the src/util/ directory in 2.3.1 to src/sox in master or whatever branch I am building. This has me back to being able to build and run new versions until I can work out why the include file search path is either not working or is finding a file that is incompatible for some reason.
> > Hmm, so the only system installed sanitizers.h somewhere in /usr/lib64/... are not the same as the file you had added to src/util/
> > I'm trying to work out if there is another package that F44 has that installs another version of sanitizers.h as I don't know where you got yours from.
> > Fun this eh?
> > Just noticed that some of the files in my include directories are called Sanitizers.h so there is a case issue in the search path
> 
> Maybe we need to rename it to freedv_sanitizers.h instead. I think I can do that change quickly in another set of PRs (both freedv-backend and freedv-gui as sanitizers.h is in the former).

You know best, but currently there is no sanitizers.h file anywhere in the freedv-gui git master tree. I still have no idea how it works once and then fails afterwards in that state, A weirdness in the include search mechanism?

tmiw:
Try latest master again?

Tyrbiter:
> Try latest master again?

That seems to have fixed it, I had not realised that all the /src/util stuff had moved to the backend repo which probably explains why I couldn't find the header file, it only appears in the build_linux/ directory during the build.

One day I might make a software engineer.

PR #1373, opened 2026-06-03
Hamlib/OmniRig: Round received frequency to nearest 100 Hz.

tmiw:
~~When double-clicking on a row in FreeDV Reporter, FreeDV will change your radio's frequency to the frequency of the user that was double-clicked. This PR adjusts this frequency to the closest 100 Hz increment, similar to what's done when manually entering a frequency on the main window.~~

EDIT: New description below.

I received reports that the frequency sent to FreeDV Reporter and the frequency displayed in app aren't consistent due to changes made prior to 2.3.0. This PR rounds the received frequency from the radio instead so that FreeDV changes to the correct frequency when the user double-clicks on a row in FreeDV Reporter.


barjac:
Surely double clicking on a row in the reporter would select the row and then de-select it?

I would think this may be better done with a context menu when right clicking the frequency itself (to be consistent with right clicking the callsign for qrz lookup and right clicking a Msg for copy message).

tmiw:
> Surely double clicking on a row in the reporter would select the row and then de-select it?
> 
> I would think this may be better done with a context menu when right clicking the frequency itself (to be consistent with right clicking the callsign for qrz lookup and right clicking a Msg for copy message).

Going to someone else's frequency seems to be a pretty common activity, no? Wouldn't needing to right-click add friction to that process? Note that it's been double-click for quite a while.

barjac:
> > Surely double clicking on a row in the reporter would select the row and then de-select it?
> > I would think this may be better done with a context menu when right clicking the frequency itself (to be consistent with right clicking the callsign for qrz lookup and right clicking a Msg for copy message).
> 
> Going to someone else's frequency seems to be a pretty common activity, no? Wouldn't needing to right-click add friction to that process? Note that it's been double-click for quite a while.

As one who still actually touches the radio to change frequency I was not even aware that this existed.
I tune the radio and FreeDV and KLog just follow.
I will leave this to others for comment - sorry for the noise! :)

barjac:
OK since nobody else seems interested in testing this I have just tested it with my TS-450S.

It does as I expected, the double click selects and deselects the row.
Unless moused over a message, it does also perform the frequency/mode change.
If moused over a message it ONLY selects the row. No freq or mode change.

This is confusing.
It would be better if it only changed the mode/freq if it was moused over just the frequency required.
A context menu for compatibility with other reporter functions as mentioned earlier would be the best.
A context menu could say "QSY to 7.197MHz" etc. which is more user friendly.
Two single clicks replacing one double click.
No 'friction' for me as there is no chance of timing issues associated with a slow 'double click' being interpreted as two 'single' clicks.

tmiw:
I adjusted the double-click action to only adjust frequency if double-clicking on the frequency itself.

BTW for context, this PR is intended to return FreeDV Reporter to 2.2.1 behavior regarding frequency changing. Another user tested both versions with small frequency changes and got the following:

```
FreeDV 2.2.1
TRX,         Reporter (Display), Reporter (Hold)
7.180000, 7.1800,                 7.180000
7.180049, 7.1800,                 7.180000
7.180050, 7.1801,                 7.180100
7.180100, 7.1801,                 7.180100
7.180050, 7.1801,                 7.180100
7.180049, 7.1800,                 7.180000
7.180000, 7.1800,                 7.180000
* It appears that the TRX frequency is rounded to the nearest 10Hz and that value is sent to the Reporter server.

FreeDV 2.3.1
TRX,         Reporter (display), Reporter (hold)
7.179950,  7.1799,                7.179950
7.179951,  7.1800,                7.179951
7.179950,  7.1799,                7.179950
7.179949,  7.1799,                7.179949
7.179999,  7.1800,                7.179999
7.180000,  7.1800,                7.179999
7.180001,  7.1800,                7.180001
7.180002,  7.1800,                7.180002
7.180001,  7.1800,                7.180001
7.180000,  7.1800,                7.180001
7.179999,  7.1800,                7.179999
7.179998,  7.1800,                7.179998
```

("Reporter (hold)" is the TRX frequency after double-clicking.)

I'll send a build over to them to make sure this PR does as intended, but it would be useful if others could try this as well.

tmiw:
Based on feedback I received:

1. I reverted the change to require double-clicking on the Frequency field.
2. I reverted the original PR change in favor of doing the rounding when we receive the frequency update from the radio.

For (2) at least I'm not sure if it'll cause breakage in Hamlib again so we'll need to do more testing before merge for sure (assuming the rest of the changes are fine).

PR #1375, opened 2026-06-03
Fix intermittent non-response to short PTT clicks to stop TX

barjac:
Removes event.Skip() from OnTogBtnPTT() in src/ongui.cpp.

All other toggle button handlers (OnTogBtnOnOff, OnTogBtnAnalogClick, OnTogBtnTune) do not propagate their events to parent windows.

OnTogBtnPTT was the odd one out. Calling Skip() can cause the event to be re-dispatched under certain GTK signal timing conditions, which undoes the PTT state change and leaves the radio stuck in TX after a short click.

The fix aligns PTT button event handling with the rest of the toggle buttons. Tested on Linux (GTK) — the issue appears resolved, but wider testing would be welcome as the triggering conditions are timing-dependent and may vary across platforms and GTK versions.

Diagnosed in collaboration with Claude Code


tmiw:
I think this change is okay and will merge, but out of curiosity, how did Claude come to this conclusion?

barjac:
I explained the issue that we were seeing in detail and asked Claude to analyze all the possible triggers related to the ptt button. It took Claude a relativly long time but the possibility that it could actually be turning off and back on, rather than not responding came out as the most probable culprit.

PR #1377, opened 2026-06-04
Move Tune button into Control widget

barjac:
I have always thought that positioning the Tune button in the Transmit Attenuation widget was wrong.
Another user also queried this yesterday so I investigated moving it into the Control widget.

This PR does the following:
Relocates the Tune toggle button from the TX Attenuation widget into the Control widget, positioned between "Switch to Analog" and "Start Voice Keyer".

The Tune button is a control action rather than an attenuation setting, so this placement better reflects its purpose and groups it logically with the other operation controls.
It also improves the visual balance of the right-hand panel — the TX Attenuation widget now contains only attenuation controls.
Another advantage is that the main window height may now be reduced further if required before major auto-arranging takes place.

All existing event handlers (toggle, right-click tune attenuation context menu, mouse wheel) are unchanged. The keyboard focus tab order is updated to match the new position.

The change is also reflected in USER_MANUAL.md

Tested on Linux (GTK).


tmiw:
Should the Tune button be closer to the PTT button since both cause the radio to transmit?

barjac:
> Should the Tune button be closer to the PTT button since both cause the radio to transmit?

VK also causes the radio to transmit :)

I considered the position carefully so that it was logical. Buttons are in order of most used during a regular operating session starting at the bottom with PTT, then VK, Tune etc.
Also less likely to be accidentally clicked when missing the PTT where it is.
Most users will use the monitor with VK so its more noticeable when accidentally pressed.

PR #1378, opened 2026-06-04
Add "Set tune output to minimum" to Tune button context menu for safety.

barjac:
Currently it is not possible to adjust the Tune transmit level without actually transmitting.
If the level is at default or a previously saved high level then this could be dangerous for e.g. a tuner expecting a very low power in tune mode.

This PR does the following:
Adds a context (right click) menu item to the Tune button that sets the tune attenuation to its minimum output level (-30 dB) which is available without actually transmitting.
 
The new item is always visible at the top of the menu. The existing per-band auto-save/restore items follow after a separator when a reporting frequency band is known.

The early-return guard when no band is set has been removed since the new item is band-independent.

These changes are reflected in the USER_MANUAL.md

Tested on Linux (GTK).


barjac:
Screenshot_20260604_115141

tmiw:
Wouldn't having the default tune attention be -30 also work?

barjac:
> Wouldn't having the default tune attention be -30 also work?

In theory and initially yes, but I have several times had a situation where the tune output was extraordinarily high when setting it for a new band. It is now so simple to reset it to min before readjusting it.

tmiw:
> > Wouldn't having the default tune attention be -30 also work?
> 
> In theory and initially yes, but I have several times had a situation where the tune output was extraordinarily high when setting it for a new band. It is now so simple to reset it to min before readjusting it.

What about the scenario where the user might not necessarily know to reset before tuning? (i.e. if something hardware-wise failed) By the time they reset they could very well damage something.

Perhaps Tune Attenuation's maximum should be lower than 0? If so, what should it be?

barjac:
> > > Wouldn't having the default tune attention be -30 also work?
> > 
> > 
> > In theory and initially yes, but I have several times had a situation where the tune output was extraordinarily high when setting it for a new band. It is now so simple to reset it to min before readjusting it.
> 
> What about the scenario where the user might not necessarily know to reset before tuning?

I suppose that the same could be argued about transmitting?

> (i.e. if something hardware-wise failed) By the time they reset they could very well damage something.
 
We can't protect against every eventuality but having the tools available to avoid problems is sensible.

> 
> Perhaps Tune Attenuation's maximum should be lower than 0? If so, what should it be?

No I think that is being paranoid, levels downstream of FreeDV in the TX chain are beyond our control anyway. Some users may want to tune at high power.

Some background:
This feature was actually requested by another user who had hit the issue of being unable to adjust the tune level before actually transmitting. He is satisfied with this solution.

tmiw:
PR seems to work in my testing. Seems to also contain #1377, though, so we should probably get that merged first before this one.

BTW, maybe we should have the right-click menu options here auto-change to "Tune" from "TX" when tuning is active?

image

(I was initially right-clicking there before I realized that the option added in this PR was a right-click option on the Tune button itself. I could be weird though and right-click on the Tune button is more intuitive here.)

tmiw:
Oops, accidentally closed, sorry! Should be reopen now.

barjac:
Only the USER_MANUAL.md changes for 1377 were moved here to avoid conflict.

tmiw:
> Only the USER_MANUAL.md changes for 1377 were moved here to avoid conflict.

It looks like the changes in topFrame.cpp from that PR ended up in here too (see the screenshot above where Tune is in the Control box now).

tmiw:
Actually, it looks like there's a separate context menu for tune attenuation. Maybe the TX one should be disabled while Tune is active to avoid confusion, but that can be another PR.

Issue #987, opened 2025-07-26
'From mic' is not showing mic level in half duplex mode during RX.


barjac:
I have been asked to report this on behalf of another station however I agree that this should be available in order to set up TX audio levels while operating and not transmitting.
This can be achieved now by switching to full duplex but then there are occasional distracting burps and squeaks (at least here) from the speaker/headphones while transmitting.

tmiw:
It's possible that once #975 merges, audio quality may become good enough such that just using full duplex may be sufficient. (Also, what about the existing monitor functionality + setting the radio to 0W/transmitting into a dummy load?)

Anyway, this would require running at least some of the TX processing pipeline even when not in full-duplex mode. Not sure this is necessarily a good idea on systems that may be closer to marginal for RADE RX.

barjac:
> It's possible that once [#975](https://github.com/drowe67/freedv-gui/pull/975) merges, audio quality may become good enough such that just using full duplex may be sufficient.

Sounds good.

> (Also, what about the existing monitor functionality + setting the radio to 0W/transmitting into a dummy load?)

It's the situation where a station is already in QSO, has had a report that AF level is too low/high and could simply select 'From mic' and make the adjustment by speaking.

Currently even with a dummy load or zero output power they would need to disable reporting as well or change frequency to avoid apparently doubling (in the reporter) with the station transmitting, disrupting the QSO or net. It happens often and the transmitting station stops to listen on seeing one of the group in TX.

> Anyway, this would require running at least some of the TX processing pipeline even when not in full-duplex mode. Not sure this is necessarily a good idea on systems that may be closer to marginal for RADE RX.

Yes I had considered that this may be an issue. Could the audio signal be sampled early in the TX chain for this I wonder, since for this to be useful only the level is relevant?

tmiw:
> Yes I had considered that this may be an issue. Could the audio signal be sampled early in the TX chain for this I wonder, since for this to be useful only the level is relevant?

I looked at the code and we basically end up running TX audio through the equalizer/volume control as well as noise reduction before being able to sample for the "From Mic" plot. Additionally the bar graph on the left hand side apparently requires From Mic to be getting populated in order to produce anything useful. 

Anyway, I say we should wait until #975 merges and then revisit based on how much CPU usage FreeDV is now taking.

tmiw:
Per feedback from #1382 we'll go ahead and close given that FreeDV has a working AGC now. If this is something that's still necessary we can always reopen.

PR #1382, opened 2026-06-08
TOT beep fixes: pipeline correctness, audio quality, FIFO drain race

barjac:
Found during testing of PR #1366. Four bugs in the BeepStep/TOT beep implementation, each with a distinct root cause.

### 1. Redesign BeepStep: always-present pipeline step with pre-generated waveform

The original approach switches BeepStep in and out of the pipeline via an EitherOrStep. The switch mid-frame causes an audible transient at beep start/end. Replace with a single always-present step that pre-generates the entire waveform at construction time — passthrough when idle, plays from the buffer when active.

### 2. Fix TX-stop pipeline stall and RADE timing corruption

**Pipeline stall**: `g_totBeepActive` can be cleared by `stopTotBeep_()` (e.g. user releases PTT) before BeepStep finishes playing. Add `s_totBeepPlaying` which stays true for the full duration so `rxProcessing_` keeps the pipeline running.

**RADE timing corruption**: RADE accumulates input before producing output, emitting 0 samples on most calls and a full frame occasionally. With RADE active during the beep, BeepStep receives frames of varying size which corrupts its sample-counter timing — heard as 'S+I' instead of '5'. Bypass RADE for the duration of the beep so BeepStep always receives fixed-size frames from the passthrough pipeline.

### 3. Fix audio artefacts — noise-gate keepalive and completion-frame silence

**Noise-gate crunches**: Hardware with an automatic noise gate (including many USB audio dongles) closes during inter-dit silences. The gate-open transient at the start of each dit produces a crunch. Add a brief (~5 ms) sub-audible carrier at the end of each silence segment so the gate stays open without introducing a perceptible sub-tone during the gaps. Also extend the leading silence from 10 ms to 50 ms so the gate is fully open before the first dit.

**Completion-frame noise**: The beep waveform is an exact multiple of the pipeline frame size. When `sampleCtr_` reaches `samplesToGenerate_` at the start of a frame, the remainder of that frame falls through to the passthrough branch and passes raw radio audio from the RADE bypass pipeline to the speaker. Output silence for the entire completion frame instead.

Also: keep `outFifo` at least 3 frames ahead during the beep (substituting silence when `infifo1` is empty) to absorb OS scheduling jitter that would otherwise cause underrun gaps in the tone.

### 4. Fix FIFO drain race that truncated the beep to 3 dits on 2-card setups

On a 2-card setup the beep waveform is written to `outFifo2` ~17 000 samples at a time. When BeepStep finishes generating all samples, both `g_totBeepActive` and `s_totBeepPlaying` clear, making `processInputFifo` false. On the very next `rxProcessing_` call `clearFifos_()` was wiping ~16 000 still-unplayed samples from `outFifo2` — only the ~360 ms the speaker had already drained was heard, giving 3 dits instead of 5.

Fix: when entering the `!processInputFifo` path, check whether `outFifo2` still holds audio. If so, only reset the radio input FIFO and leave `outFifo2` alone until it empties.

### DIFF
--- a/src/pipeline/BeepStep.cpp
+++ b/src/pipeline/BeepStep.cpp
@@ -2,6 +2,10 @@
 #include "BeepStep.h"
 
 static const int TONE_AMPLITUDE = 20000;
+// Brief pre-charge level for silence segments: high enough to prevent hardware noise-gate
+// from closing (~-50 dBFS), applied only for the last 'ramp' samples of each silence so
+// the inter-dit gaps are perceived as true silence (no continuous sub-tone).
+static const int SILENCE_KEEPALIVE = 100;
 static const double M_2PI = 2.0 * M_PI;
 
 BeepStep::BeepStep(int sampleRate, int frequency,
@@ -38,9 +42,16 @@ BeepStep::BeepStep(int sampleRate, int frequency,
                     env = 0.5 * (1.0 + std::cos(M_PI * (i - (segSamples - ramp)) / ramp));
                 else
                     env = 1.0;
-                pregenSamples_[idx] = static_cast<short>(TONE_AMPLITUDE * env * std::cos(phase));
+                // Floor the envelope at SILENCE_KEEPALIVE so the tone starts and ends at
+                // the same level as the preceding/following silence pre-charge — no step.
+                pregenSamples_[idx] = static_cast<short>(
+                    (SILENCE_KEEPALIVE + (TONE_AMPLITUDE - SILENCE_KEEPALIVE) * env) * std::cos(phase));
             } else {
-                pregenSamples_[idx] = 0;
+                // True silence, except for the last 'ramp' samples which act as a brief
+                // noise-gate pre-charge so the gate is open when the next tone onset arrives.
+                pregenSamples_[idx] = (i >= segSamples - ramp)
+                    ? static_cast<short>(SILENCE_KEEPALIVE * std::cos(phase))
+                    : 0;
             }
             phase += phaseStep;
             if (phase >= M_2PI) phase -= M_2PI;
@@ -56,7 +67,6 @@ int BeepStep::getOutputSampleRate() const FREEDV_NONBLOCKING { return sampleRate
 
 short* BeepStep::execute(short* inputSamples, int numInputSamples, int* numOutputSamples) FREEDV_NONBLOCKING
 {
-    // Check for activation once per buffer (never mid-buffer)
     if (!playing_ && isActiveFn_()) {
         sampleCtr_ = 0;
         playing_ = true;
@@ -71,15 +81,17 @@ short* BeepStep::execute(short* inputSamples, int numInputSamples, int* numOutpu
         if (playing_ && sampleCtr_ < samplesToGenerate_) {
             outSample = pregenSamples_[sampleCtr_++];
         } else if (playing_) {
-            // Pattern complete — passthrough + fire callback once
-            outSample = inputSamples[index];
+            // Beep just finished: silence the rest of this frame so that radio
+            // audio passing through the bypass pipeline doesn't leak to the speaker.
+            outSample = 0;
             if (!completed) {
                 completed = true;
                 playing_ = false;
                 onCompleteFn_(*this);
             }
+        } else if (completed) {
+            outSample = 0;  // silence remainder of completion frame
         } else {
-            // Idle passthrough
             outSample = inputSamples[index];
         }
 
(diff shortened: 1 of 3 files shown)


tmiw:
I'm actually considering just having both RX and TX threads run all the time, only switching in pipeline steps as needed. That should help with some of the issues (and allow #987 to be closed too). I'll play some more once I get computer access again later today.

barjac:
That sounds like possibly the way to go, but quite a task. However it may increase CPU load for modest systems like my RPi4.
If you do go that way I suspect it won't be immediately, so it would be good in the short term to add this PR to clean up the TOT warning notification sound.
In the long term, with both RX and TX threads running continuously commits 2 & 4 here could just be dropped leaving 1 & 3 which would still be very worthwhile.
Regarding #987 I'm not sure that is so important nowadays with AGC as Mic level settings are less critical.
In fact it is not really a good idea to try to adjust mic levels on the air as I see many people trying to do.
They end up fighting the AGC and make things worse.

src/pipeline/TxRxThread.cpp
```diff
@@ -872,23 +887,56 @@ void TxRxThread::rxProcessing_(IRealtimeHelper* helper) FREEDV_NONBLOCKING
...
+                equalizedMicAudioLink_->clearFifo();
+        }
+        else
+        {
+            clearFifos_();
+            // For 2-card setups outfifo2 is the speaker FIFO and is not written
+            // by the TX pipeline.  Feed silence each cycle to prevent the
+            // continuous FIFO underruns that cause PipeWire to emit a startup
+            // transient when the beep fires after 15 s of TX.
+            if (g_nSoundCards != 1)
```
tmiw, line 919: `processInputFifo` is always `true` for RX-only setups so this if condition is redundant here.

src/pipeline/TxRxThread.cpp
```diff
@@ -872,23 +887,56 @@ void TxRxThread::rxProcessing_(IRealtimeHelper* helper) FREEDV_NONBLOCKING
...
         (!tmpVkTx && ((tmpHalfDuplex && !tmpTx) || !tmpHalfDuplex)) ||
         totBeepActive;
+
+    int nsam_one_speech_frame = (freedvInterface.getRxNumSpeechSamples() * outputSampleRate_) / freedvInterface.getRxSpeechSampleRate();
+    auto outFifo = (g_nSoundCards == 1) ? cbData->outfifo1 : cbData->outfifo2;
+
     if (!processInputFifo)
     {
-        clearFifos_();
+        if (g_nSoundCards != 1 && outFifo->numUsed() > 0)
```
tmiw, line 903: I think it might be better if we adjust `clearFifos_()` to not clear the output FIFO on RX and revert this added code back to what it was in the original PR. There's already lower level logic to constantly write zeroes if there's nothing in the buffer (example: `PulseAudioDevice::StreamWriteCallback_()`).

src/pipeline/BeepStep.cpp
```diff
@@ -1,106 +1,108 @@
...
 #include <cmath>
-#include <cassert>
-
 #include "BeepStep.h"
 
-// M_PI is not available on some compilers, so define it here just in case.
-#ifndef M_PI
-    #define M_PI 3.1415926535897932384626433832795
-#endif
+static const int TONE_AMPLITUDE = 20000;
```
tmiw, line 4: This was 8192 in the original PR. Wouldn't this be even greater than it originally was in the PR (16384) and thus be uncomfortably loud again?

tmiw (commented):
This does work for me in testing but I have some concerns/questions. See below.

src/pipeline/TxRxThread.cpp
tmiw: Although if there's less than the audio latency number of samples remaining, they might not get sent to the device until TOT gets fired again or the RADE decoder starts decoding. Hmm.

tmiw:
I'm going to have some changes in the original PR shortly to do the following:

1. Always run RX thread but add some additional pipeline steps to ensure that the actual RX processing doesn't happen unless needed.
2. Mix the TOT beep into the output audio instead of outright replacing it.

Additionally, I'll see what minimal changes are required to get the CW '5' to be emitted instead of a single beep.

EDIT: I looked into doing (1) for the TX thread too but the logic to actually have it run looks more complicated and I really don't want to mess anything up there, so I'm leaving it alone for now.

barjac:
There is a bug in 1366 that breaks the Tune button - no Tune output at all.
Works fine in master.

barjac:
Closing as fixed by PR #1366

PR #1387, opened 2026-06-11
Right-justify SNR column in FreeDV Reporter, with trailing padding

barjac:
A small modification to the SNR column to avoid the current zig-zag alignment and stop it getting too close to adjacent values, depending on the column arrangement.

image


src/gui/dialogs/freedv_reporter.cpp
```diff
@@ -3424,7 +3427,7 @@ void FreeDVReporterDialog::FreeDVReporterDataModel::onUserConnectFn_(std::string
         temp->lastRxCallsign = parent_->UNKNOWN_STR;
         temp->lastRxMode = parent_->UNKNOWN_STR;
         temp->snrVal = UNKNOWN_SNR_VAL;
-        temp->snr = parent_->UNKNOWN_STR;
+        temp->snr = parent_->UNKNOWN_STR + wxT("  ");
```
tmiw, line 3430: Any chance we'd need to add padding on the left as well?

Tyrbiter: I think the SNR limits in freedv-gui involve only 3 significant figures and the sign, so 5 characters total including the decimal point.

Any different in other platforms such as Flex?

tmiw: I'm not sure about all of the others. I think AetherSDR does have a similar FreeDV Reporter window but I haven't used it in a bit.

barjac: > Any chance we'd need to add padding on the left as well?

I can't imagine why. If the field width is already adequate which it obviously is, then what would that achieve?

PR #1393, opened 2026-06-12
Add momentary PTT option to FreeDV.

### LINKED ISSUE #1331: [Feature Request]  Provide an option for the PTT key to be latching or momentary
### Contact Details

@digver, @Tyrbiter, QRV@KD4E.COM

### Feature for this project

There has already been some discussion of this at the end of #1309

Historically the PTT key in FreeDV has been latching; tap once to transmit, tap a second time to stop transmit. This emulates the action of the hardware transmit or MOX button on a physical transceiver. However the PTT switch on a transceiver's microphone or foot-pedal is in most cases momentary; push and hold to transmit, release to stop transmit. An argument can be made for FreeDV to emulate either one. Providing the option would allow the operator to decide which best serves their use-case.

Going further, I have a handheld dual action PTT switch for my rig. One button is momentary, the second latching. This allows me to decide on the fly which one I want. Perhaps FreeDV could have the option to set that up as well.

So I'll suggest two possibilities.

First, provide the option to select FreeDV's PTT mode as either latching or momentary.

Second, provide the option for a modifier key to change the PTT key from momentary to latching. Here's an example of how that could work:

 * Momentary: Press and hold the SPACE key to transmit. Release the SPACE key to stop transmit.
 * Latching: Tap the CTRL and SPACE keys to begin transmit. Tap the SPACE key a second time to stop transmit.
 
I understand that at least 2 other people will need to be added to the feature request for it to be considered and that it is not a high priority request. 
 
73, digger ab3xu

### VHF/UHF

None

### Deprecated features

None

### Non-standard compiler

None

### Legal worldwide

None

### Implementation assistance

I would be available for testing and can locally clone a branch to compile and test code locally. My actual coding skills are minimal. I may also be able provide assistance in documentation. 

### Support assistance

It's difficult for me to predict where I will be in other than the near future.

tmiw:
Adds option to Tools->Options->Rig Control to enable "momentary" operation of the PTT button/key (i.e. the user has to hold it down to transmit vs. the current default "latching" behavior). 

Resolves #1331.

Disclaimer: Implemented with the help of Claude Code.


barjac:
For testing of this to be meaningful #1366 would need to be merged as well, unless the change currently in #1366 that stops the PTT button colour showing the actual TX state on de-key is to be reverted.

digver:
I commented at #1331 

> Switching the option from momentary does work for me with the linux appimage. However it modifies both the physical PTT key and the GUI-PTT key.

Looking at what I originally proposed I'm not sure which "2" is (2). I think though the discussion kind off led to the operation of the physical and gui keys being separated to more closely [emulate a hardware transceiver]( https://github.com/drowe67/freedv-gui/issues/1331#issuecomment-4406312862) with:

hardware PPT key: momentary
GUI PTT button: latching.

tmiw:
> I commented at #1331
> 
> > Switching the option from momentary does work for me with the linux appimage. However it modifies both the physical PTT key and the GUI-PTT key.
> 
> Looking at what I originally proposed I'm not sure which "2" is (2). I think though the discussion kind off led to the operation of the physical and gui keys being separated to more closely [emulate a hardware transceiver](https://github.com/drowe67/freedv-gui/issues/1331#issuecomment-4406312862) with:
> 
> hardware PPT key: momentary GUI PTT button: latching.

Sorry, to clarify, (2) was the extra keyboard key (i.e. Ctrl-Space) to do the opposite of what was configured.

Also, I read through the FR and comments again. If still desired, should I tweak this PR to always have the PTT GUI button be latching and the keyboard key be momentary (depending on config option)? Should the configuration option be there at all, or should the keyboard key always be momentary?

digver:
My thought:
1. **GUI button** is always latching. Along with that should the label be changed to XMIT, MOX, or sometting other than PTT? I have two hardware xcvrs, the KX3 is labeled XMIT and the Maestro, MOX. Do other makers use anything else?
2. **Hardware PTT key** is momentary.

If the above is true, I'm undecided if the option to change either button or key is needed, but I lean toward it NOT being neccessary. If it doesn't complicate separating the key from the button, then leave it in for the key. 

What does anyone else think?

tmiw:
> What does anyone else think?

We might still need the configuration option since people may be used to Space (or whatever was configured for the PTT key) being latching. Also, I've only seen "MOX" used in SDRs, not physical radios, so not sure if people will fully get what that does.

digver:
You're right about "what people are used to."
Thinking in the same direction, what will the new default for the key be then? I think it should be momentary, but then that requires checking to see if there is an existing config so existing users are not caught off guard with the change. I recall your writing code to check for a config file when the config location was changed. Can that be reused?

tmiw:
> You're right about "what people are used to." Thinking in the same direction, what will the new default for the key be then? I think it should be momentary, but then that requires checking to see if there is an existing config so existing users are not caught off guard with the change. I recall your writing code to check for a config file when the config location was changed. Can that be reused?

I think it should default to latching and only become momentary if specifically configured to be so. Otherwise there's a chance we'll get complaints.

tmiw:
OK, the PTT button in the main window is back to being latching and renamed to XMIT. Let me know if any other changes are needed.

barjac:
> OK, the PTT button in the main window is back to being latching and renamed to XMIT. Let me know if any other changes are needed.

1. The 'XMIT' button tooltip needs changing: s/Push to Talk/Transmit/

2. In momentary mode when using the spacebar, releasing the spacebar during the timeout warning does not stop the transmission. Clicking XMIT does stop it.

tmiw:
> 1. The 'XMIT' button tooltip needs changing: s/Push to Talk/Transmit/
> 2. In momentary mode when using the spacebar, releasing the spacebar during the timeout warning does not stop the transmission. Clicking XMIT does stop it.

All fixed as of the latest commit.

digver:
Fantastic! FreeDV-2.4.0-dev-cdc6c-x86_64.AppImage is working for me with the spacebar set to momentary and the button, as expected, being latching.

I like the "xmit ending" label change as an acknowelegement. A niggly point for me are the background and font colors for "xmit ending" make it difficult to read on my laptop running KDE Plasma. The background orange(?) is fine, but the white font is difficult to see. Would black for the font make it easier?  Maybe it's just a KDE issue. Maybe it's just a KDE issue. T

barjac:
> The background orange(?) is fine, but the white font is difficult to see. Would black for the font make it easier? Maybe it's just a KDE issue. Maybe it's just a KDE issue. T

I am using KDE Plasma6 with a Breeze light theme and the font remains black on the red and the amber/orange as intended in the code. 
Maybe try a different theme, or maybe test a build on your machine using:
https://github.com/barjac/freedv-rade-build it will only take you about 15mins.

barjac:
> > 1. The 'XMIT' button tooltip needs changing: s/Push to Talk/Transmit/
> > 2. In momentary mode when using the spacebar, releasing the spacebar during the timeout warning does not stop the transmission. Clicking XMIT does stop it.
> 
> All fixed as of the latest commit.

Hmm... not quite.
If the space bar remains pressed to the end of the timeout it then goes through the roll-down to EOO and restarts transmit.

If a book fell on the space bar it would effectively transmit for ever.

Once timed out it should not be possible to re-start transmit again until the spacebar has been released and pressed again.

tmiw:
> Hmm... not quite. If the space bar remains pressed to the end of the timeout it then goes through the roll-down to EOO and restarts transmit.
> 
> If a book fell on the space bar it would effectively transmit for ever.
> 
> Once timed out it should not be possible to re-start transmit again until the spacebar has been released and pressed again.

OK, this bug should be fixed now.

> I like the "xmit ending" label change as an acknowelegement. A niggly point for me are the background and font colors for "xmit ending" make it difficult to read on my laptop running KDE Plasma. The background orange(?) is fine, but the white font is difficult to see. Would black for the font make it easier?

Ideally it'd change to black while that label is active but I ran into a limitation of the GUI framework we use while trying to make this happen. The selected color is basically the best we were able to come up with, unfortunately.

barjac:
> Ideally it'd change to black while that label is active but I ran into a limitation of the GUI framework we use while trying to make this happen. The selected color is basically the best we were able to come up with, unfortunately.

#1405 Should fix the text colour to black on the XMIT button.

digver:
> > The background orange(?) is fine, but the white font is difficult to see. Would black for the font make it easier? Maybe it's just a KDE issue. Maybe it's just a KDE issue. T
> 
> I am using KDE Plasma6 with a Breeze light theme and the font remains black on the red and the amber/orange as intended in the code. Maybe try a different theme, or maybe test a build on your machine using: https://github.com/barjac/freedv-rade-build it will only take you about 15mins.

Yeah, the difference is I'm using Breeze dark which has white button text as the default.
Here is the diff that fixes it for me, though perhaps there is a cleaner/better way of doing it.
```diff
diff --git a/src/ongui.cpp b/src/ongui.cpp
index b6ce23a0..52cfd1d0 100644
--- a/src/ongui.cpp
+++ b/src/ongui.cpp
@@ -1342,7 +1342,9 @@ void MainFrame::togglePTT(void) {
     {
         // Amber during TX->RX drain: distinct from TX (red) and RX (default),
         // black text readable throughout.
+       // Explicitly SetForegroundColor to black for themes with a light forground color
         m_btnTogPTT->SetBackgroundColour(wxColour(255, 165, 0));
+       m_btnTogPTT->SetForegroundColour(wxColour(*wxBLACK));
         m_btnTogPTT->SetLabel("TX Ending");
         m_btnTogPTT->Refresh();
 
@@ -1624,6 +1626,8 @@ void MainFrame::togglePTT(void) {
     m_btnTogPTT->SetValue(newTx);
     m_btnTogPTT->SetLabel(_("&XMIT"));
     m_btnTogPTT->SetBackgroundColour(m_btnTogPTT->GetValue() ? *wxRED : wxNullColour);
+    m_btnTogPTT->SetForegroundColour(m_btnTogPTT->GetValue() ? *wxBLACK : wxNullColour);
+
     
     // The Report Frequency drop-down should not be modifiable during TX.
     // Additionally, tuning during normal TX is verboten.
```

[ongui.cpp.diff.txt](https://github.com/user-attachments/files/29152368/ongui.cpp.diff.txt)

digver:
> > Ideally it'd change to black while that label is active but I ran into a limitation of the GUI framework we use while trying to make this happen. The selected color is basically the best we were able to come up with, unfortunately.
> 
> #1405 Should fix the text colour to black on the XMIT button.

@barjac you posted your fix as I was writing my comment.

barjac:
> > Hmm... not quite. If the space bar remains pressed to the end of the timeout it then goes through the roll-down to EOO and restarts transmit.
> > If a book fell on the space bar it would effectively transmit for ever.
> > Once timed out it should not be possible to re-start transmit again until the spacebar has been released and pressed again.
> 
> OK, this bug should be fixed now.
> 

Still happening - I will re-test with only this PR merged.
Double checked - there is a very short burst of RX and the TX re-starts.

Screenshot_20260619_235608

barjac:
> @barjac you posted your fix as I was writing my comment.

OOps!

digver:
That's ok! It tells me I was on the right track!

tmiw:
The previous solution didn't compile on Ubuntu 22.04 anyway (wxWidgets 3.0 and the thing I was keying off of needed a newer release). Try now?

BTW I guess the foreground color was actually changing, but it's not reverting back on macOS:

image

I'll take a look at the PR I just merged again and see if we can revert the foreground color a different way.

tmiw:
Unfortunately the only two foreground color choices on macOS apparently are "don't touch it at all" or black, so I disabled @barjac's recent PR changes for that platform. They should still work on Linux and Windows, though.

tmiw:
@barjac's latest commit works on macOS BTW. I wonder why my last two attempts didn't work on Linux (they definitely did on macOS).

tmiw:
Looks like the previous commits caused a compile failure on Ubuntu 22.04 / wxWidgets 3.0. Can you try the latest and make sure the TOT fix is still working?

TreeTopHam:
>Looks like the previous commits caused a compile failure on Ubuntu 22.04 / wxWidgets 3.0. Can you try the latest and make sure the TOT fix is still working?

On latest  Win11 build, using space bar for PTT creates the following issues:

EDIT:

**If Momentary PTT enabled:**
If PTT is invoked by holding down the space bar and if the TOT warning popup fires, releasing the space bar during the TOT warning popup crashes the GUI & leaves radio in TX

**If Latching PTT enabled:**
If PTT is is invoked by holding down the space bar and if the space bar is held down thru the duration of the TOT warning popup, the GUI crashes & leaves radio in TX

Crash Dump file is uploaded to Dropbox here:
https://www.dropbox.com/scl/fi/4j7a8jih7a2ae633fe9zn/freedv.exe16664.dmp?rlkey=2jr2s2y98vwqf9otcalrwp4vy&st=qoohgl77&dl=0

Windows build:  -ac63-windows-x86_64.exe

barjac:
> Looks like the previous commits caused a compile failure on Ubuntu 22.04 / wxWidgets 3.0.
Can you try the latest and make sure the TOT fix is still working?

Works perfectly in Mageia 9 (wxGTK-3.2)
`master @ 93c8dacb pr:1373 @ 3baba8ed pr:1380 @ 5dfc2858 pr:1393 @ 8ecd5519 2026-06-20 18:35`

Same in Mageia 10 (wxGTK-3.3)
`master @ 93c8dacb pr:1373 @ c1ee5a68 pr:1393 @ 929d334b pr:1380 @ fad981b5 2026-06-20 18:53`

Bit puzzled by some of those commits reported, must check that out.

EDIT: OK ignore the PR commit hashes they are not the upstream ones, next time they will be :)

digver:
With Fedora 44 KDE Plasma 6.70 and wxGTK-3.2.9-2.fc44:

FreeDV-2.4.0-dev-c1cb-x86_64.AppImage times out properly using the XMIT key. With the spacebar set to momentary it times out at 3 minutes and stops TX even when the spacebar continues to be held down.

I don't know if this means anything?
Building FreeDV-2.4.0-dev-c1cb-x86_64 from github on my laptop the I get a pop-up with this error at the end of the TOT:

```
../src/unix/utilsx11.cpp(2645): assert ""Assert failure"" failed in wxGetKeyStateGTK(): Unsupported key 32, the only supported ones are: Ctrl, Alt, Shift, Caps Lock, Num Lock and Scroll Lock for GTK 3.18+
```
This happens with both the XMIT button and momentary spacebar.

digver:
Screenshot_20260620_155725

digver:
here is the backtrace:
```
ASSERT INFO:
../src/unix/utilsx11.cpp(2645): assert ""Assert failure"" failed in wxGetKeyStateGTK(): Unsupported key 32, the only supported ones are: Ctrl, Alt, Shift, Caps Lock, Num Lock and Scroll Lock for GTK 3.18+

BACKTRACE:
[1] wxGetKeyState(wxKeyCode)
[2] MainFrame::OnTOTTimer(wxTimerEvent&) /home/digger/src/git/freedv-gui/src/ongui.cpp:1288
[3] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[4] wxEvtHandler::SearchDynamicEventTable(wxEvent&)
[5] wxEvtHandler::TryHereOnly(wxEvent&)
[6] wxEvtHandler::ProcessEventLocally(wxEvent&)
[7] wxEvtHandler::ProcessEvent(wxEvent&)
[8] wxEvtHandler::SafelyProcessEvent(wxEvent&)
[9] wxTimerImpl::SendEvent()
[10] g_main_loop_run
[11] gtk_main
[12] wxGUIEventLoop::DoRun()
[13] wxEventLoopBase::Run()
[14] wxAppConsoleBase::MainLoop()
[15] wxEntry(int&, wchar_t**)
[16] main /home/digger/src/git/freedv-gui/src/main.cpp:257
[17] __libc_start_main
[18] _start
```

tmiw:
Actually, the most recent commit doesn't work at all on Ubuntu 24.04 with GNOME--it just keeps toggling between TX and RX when holding down the PTT key. I tried to go back a couple of commits too in case whatever I did messed it up, no difference. Might need to rethink how the fix's implemented.

EDIT: I think it might be because `wxGetKeyState()` has limitations on Wayland--see https://docs.wxwidgets.org/3.2/group__group__funcmacro__misc.html#ga82a18ac43492bfb375604c41b71ce84f:

> In wxGTK, this function can be only used with modifier keys (WXK_ALT, WXK_CONTROL and WXK_SHIFT) when not using X11 backend currently.

EDIT 2: I had momentary mode turned off, oops. Still, given the above limitation, we should probably find a workaround.

tmiw:
OK, the most recent commit gets rid of that assertion box and doesn't otherwise break what @barjac did--hopefully.

digver:
I read somewhere

> OK, the most recent commit gets rid of that assertion box and doesn't otherwise break what @barjac did--hopefully.

I did a quick build and check.
* TOT stops tx at 3 minutes with both button and key
* Holding the spacebar longer than 3 minutes -> TX is stopped by the TOT
* There is no assertion box error

@tmiw holding down the spacebar with it set latching sure does make things go crazy, doesn't it? :smiley:

tmiw:
> @tmiw holding down the spacebar with it set latching sure does make things go crazy, doesn't it? 😃

Indeed. Just checked in something that should hopefully help with that too.

TreeTopHam:
@tmiw: 
>Indeed. Just checked in something that should hopefully help with that too.

Problems I reported above are now fixed in commit: 82b9971  :)
Tested on Linux Mint & Win11

- The only quirk I now see is that sometimes when space bar is held down, the TOT popup comes up but is greyed out (doesn't have focus). And occasionally the warning sound is totally missing or only plays 1 of the 5 Morse beeps.

I see this on both Windows & Linux Mint. Screen shot attached.

TOT Warning Greyed Out

barjac:
This latest does not work correctly in momentary mode.
A short press on spacebar acts like latching.
This is exacerbated if there is TX delay set in Tools->Options->Rig Control.

PR #1409 applies on top of #1393 and fixes it for me in Mageia 10 with or without a TX delay.
Built with:
master @ 93c8dacb pr:1373 @ ff3e54c8 pr:1380 @ cc1f90ea pr:1393 @ 82b9971c pr:1409 @ 3f4ddc5d

tmiw:
Confirmed the issue on macOS too (I typically run without a delay which is probably why I didn't see this sooner). @barjac's PR looks okay and I confirmed that it fixes the issue on macOS too, so I went ahead and merged.

tmiw:
> * The only quirk I now see is that sometimes when space bar is held down, the TOT popup comes up but is greyed out (doesn't have focus). And occasionally the warning sound is totally missing or only plays 1 of the 5 Morse beeps.

I applied the same logic I use for the FreeDV Reporter window so hopefully that'll help.

Re: missing beeps--how often would you say this happens?

TreeTopHam:
>I applied the same logic I use for the FreeDV Reporter window so hopefully that'll help.
>
>Re: missing beeps--how often would you say this happens?

Missing TOT beeps: still occur with commit: 107edd9  (tested on Linux Mint & Win11)
They are random - maybe once in 20 PTTs?? 

The greyed out TOT popup issue is fixed :)

**Logic question:** 
Currently the PTT button turns red on mouse button 'press' but radio toggles to TX only on mouse button 'release'


**Can this be changed to either:** 

- Turn PTT button red AND toggle radio to TX mode when mouse button is pressed, OR
- Turn PTT button red and toggle radio to TX mode ONLY WHEN mouse button is released

That way PTT turns red and radio toggles to TX at the same time.

tmiw:
OK, I think the missing beeps are fixed now.

> **Logic question:** Currently the PTT button turns red on mouse button 'press' but radio toggles to TX only on mouse button 'release'
> 
> **Can this be changed to either:**
> 
> * Turn PTT button red AND toggle radio to TX mode when mouse button is pressed, OR
> * Turn PTT button red and toggle radio to TX mode ONLY WHEN mouse button is released
> 
> That way PTT turns red and radio toggles to TX at the same time.

I think @barjac wanted the current behavior, so maybe he can chime in and let us know if one of the above options is okay?

TreeTopHam:
>OK, I think the missing beeps are fixed now.

OK so far  :)
Win11 & Linux Mint
 
>I think @barjac wanted the current behavior, so maybe he can chime in and let us know if one of the above options is okay?
>

**My comments:**
Current behavior makes the PTT button consistent with the other buttons on the GUI - they all '_fire_' when the mouse button is released. That's good, and I can see that logical approach.

_However_, in 'Latching mode', this behavior makes the PTT button work exactly opposite of the space bar when used to key the radio. Space bar keys radio on space bar '_press_' but the PTT button keys radio on mouse button '_release_''.

So, I can click the PTT button with my mouse and it turns red, but as long as I don't release the mouse button the radio will never enter TX mode. Someone with a 'heavy' finger on the mouse would wonder why they are not transmitting anything even though their PTT button is red.

EDIT:
Perhaps we wait to turn PTT button red until mouse button 'release' ??

tmiw:
What about having PTT start when the space bar (or whatever the PTT key is set to) is released? This would only apply in latching mode; key down would still start TX in momentary mode.

TreeTopHam:
Just have the PTT button behave like the Tune button (change to RED and TX on release of the mouse button.

Everything else can stay as is.

tmiw:
> Just have the PTT button behave like the Tune button (change to RED and TX on release of the mouse button.
> 
> Everything else can stay as is.

Updated.

barjac:
The XMIT button now flickers again on click going into TX. The whole object of having the colour change on button down was to avoid the change to blue or grey (depending on theme) that happens between button down and button up, which can't be avoided other than by changing to red on button down.

A click is a single action and should not be a slow down - up, why would you press and hold for a click action? (other than to work around an old bug that is now fixed :)

I could not see any problem as it was. No flicker, clean instant colour change to red and TX starts maybe 10-20ms later with a genuine 'click' which is insignificant.

Just because it uses different logic to the space bar and Tune button does not make it wrong. It was done that way for a reason.

The Tune button is fine as it is.

TreeTopHam:
>Updated.

Thanks. No issues seen on Win11 or Linux Mint.

I'm not seeing the XMIT flicker that @barjac mentions. 
@barjac : Curious why you don't see the flicker on the Tune button also? Tune doesn't change to red until mouse up.

tmiw:
Ubuntu 24.04 w/GNOME and wxWidgets 3.2: I didn't see the flicker either with @barjac's latest PR or this one but in both cases, if I push the PTT button, drag off the button and let go, FreeDV still transmits. I'm considering just reverting the last commit I made and merging as-is since it seems to work other than that one minor issue @TreeTopHam reported; to me, it seems better not to actually TX unless the button is released while the pointer's still over the button.

BTW the PTT button is always going to be latching regardless of the state of the momentary/latching config flag. Not sure if that was made clear earlier.

tmiw:
Hmm, Fedora 44 isn't behaving quite right either with that test. Also with GNOME. Haven't tried @barjac's PR yet, will do so now.

EDIT: Seems to work okay with that PR after understanding that the point of it was to start PTT on button down.

tmiw:
Merged @barjac's PR. How does this work for you, @TreeTopHam?

TreeTopHam:
@tmiw:
>BTW the PTT button is always going to be latching regardless of the state of the momentary/latching config flag. Not sure if that was made clear earlier.

>Merged @barjac's PR. How does this work for you, @TreeTopHam?

- PTT has now become _momentary_ when using 'Momentary PTT' mode for the space bar. (I think we wanted PTT to always be latching??)
- In **_Windows_** build, if in latching mode for the space bar: START MODEM -> XMIT -> any subsequent button press or menu click keys radio & turns XMIT button red. No way to STOP MODEM after this point. Exiting GUI with 'x' crashes program.

**Windows Crash Dump file uploaded to Dropbox:**
https://www.dropbox.com/scl/fi/h1838c3yywgf1j3lv45lj/freedv.exe.7508.dmp?rlkey=cmj34i850rji0g2agjozrdu0q&st=lon4zje0&dl=0

Tested on Linux Mint & Win11

digver:
I'm out on a boat with no ham rig or computer. I'll catch up when back home the beginning of next week

tmiw:
Due to the regressions on Windows that @TreeTopHam reported (and which I was able to reproduce), I reverted the previous PR. I did check in something else that should hopefully cause the red background to appear on the XMIT button ASAP after releasing the mouse button. On Fedora with KDE Plasma it seems to work okay, no flicker (even with the TX delay set to something like 250ms)--but I could be misunderstanding what exactly is flickering when.

Also, I'm doing most of my Linux testing in various VMs, which could be a factor as to why I'm not seeing what @barjac is seeing.

barjac:
I agree with your revert. I had missed that momentary should not affect the XMIT button which should remain latching.
That was the main cause of the over-complication in the mouse handling.

I think a straight revert to this branch at 5bd8f552 would be the best compromise as we know it does not cause any cross platform issues and removes the 'flashes' [1] by switching to red on button-down.
The press-and-hold complaint stands but it was a minor edge case rather than a real bug.


[1] **An explanation from Claude of the 'flash' that I seem to notice more than others do.**
The brief blue or grey flash visible between mouse-down and mouse-up on the XMIT button is GTK applying its :active CSS pseudo-class at the GTK level the moment the button is pressed. This happens before any wx event handler runs, so there is no opportunity to preempt it from wx code. 
SetBackgroundColour() in wx installs a CSS provider for the widget's normal state but does not override GTK's :active state styling, which has higher specificity in the theme CSS. Suppressing it cleanly would require injecting a custom CSS provider that explicitly covers :active, which is theme-dependent and fragile.
The flash is present on every GTK button — it is just more noticeable on XMIT because users watch that button closely for the TX state colour change.

TreeTopHam:
All good now :)

Linux Mint & Win11

tmiw:
I'm going to keep the PR as-is. The reason is that when I went to try [eb0ec36](https://github.com/drowe67/freedv-gui/pull/1393/commits/eb0ec36eb13c0be4dc0850244d6d8148409ed294) (which is equivalent to https://github.com/drowe67/freedv-gui/commit/5bd8f55268ddc6ad1e17ee48958784ecdf44eceb plus merges down from `master`), I noticed intermittent delays in making the background color on the XMIT button no longer red when dragging off of the button. This was mainly on macOS.

Additionally, making the XMIT button red on button down is different behavior compared to every other SDR program I know of (SmartSDR, AetherSDR and Thetis) and could cause confusion there, especially since RF doesn't actually go out until button release.

PR #1397, opened 2026-06-14
RADEV2: Standardize mode as USB.

### LINKED ISSUE #947: [Feature Request] LSB/USB practices to coincide with RADE V2
### Contact Details

@barjac, @g4dya, @HT32BSX115, @ka9q 

### Feature for this project

AIUI, the intention is that FreeDV RADE V2 could be built into transceivers by manufacturers as a mode that could be used directly without the need for either a computer running FreeDV-gui or an ezDV-type box.

As things stand, manufacturers would need to add at least TWO modes: FDV-L and FDV-U to match our current on air convention. Clearly it would be better if they only needed to add one.

To do this in a way that is compatible with current practice, I suggest that when using RADE V2 with LSB, the audio spectrum should be inverted within FreeDV-gui so that it is radiated as USB. This would mean that manufacturers would need only one FDV mode not two, while respecting the LSB convention below 10 MHz when switching to analog SSB voice.

It would also help to future-proof things if, for RADE V2, we adopted a convention of reporting center frequencies for FreeDV instead of suppressed carrier (dial) frequencies: e.g. 14237.5 instead of 14236 kHz.

If manufacturers adopt FreeDV, they will presumably use IQ modulation. Then the legacy SSB practices will become baggage. Let's think ahead and prepare for that.

### VHF/UHF

No

### Deprecated features

No

### Non-standard compiler

No

### Non-standard compiler

No

### Implementation assistance

I can help with testing, as I already do to an extent.

### Support assistance

I can help with testing, as I already do to an extent.

tmiw:
Removes support for mode updates to DIGL/LSB as part of move to RADEV2. FreeDV will still change over to LSB when < 10 MHz and using the Analog button.

Resolves #947.


Tyrbiter:
Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.

tmiw:
> Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.

No frequency offset yet as the FR that hasn't been decided on. I guess now's the time to try to figure that out, though. Feedback wanted here.

Tyrbiter:
> > Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.
> 
> No frequency offset yet as the FR that hasn't been decided on. I guess now's the time to try to figure that out, though. Feedback wanted here.

My point was that with a nominal 1500Hz audio centre frequency (actually 1475Hz for RADEV1) the USB/LSB switch means that there is a 2x audio centre frequency shift in the on-air RF spectrum centre frequency.

If checking the frequency using the analog button then if <10MHz this check is not listening to the frequency the digital voice transmission and reception will be using, so it seems to me we should at least indicate somehow that there is a USB->LSB mode change. Some people use the analog mode for normal SSB QSOs as well as audio monitoring so perhaps we should have a dual-function on the analog button, separate DV monitoring and analog SSB. Not sure how this could be done, but we already have right-click context menus on some control buttons.

tmiw:
> My point was that with a nominal 1500Hz audio centre frequency (actually 1475Hz for RADEV1) the USB/LSB switch means that there is a 2x audio centre frequency shift in the on-air RF spectrum centre frequency.
> 
> If checking the frequency using the analog button then if <10MHz this check is not listening to the frequency the digital voice transmission and reception will be using, so it seems to me we should at least indicate somehow that there is a USB->LSB mode change. Some people use the analog mode for normal SSB QSOs as well as audio monitoring so perhaps we should have a dual-function on the analog button, separate DV monitoring and analog SSB. Not sure how this could be done, but we already have right-click context menus on some control buttons.

We wouldn't be getting rid of the current mode indicator on the bottom of the main window, so wouldn't that be the indication that the mode changed?

Though speaking of which, I should probably update the coloring requirements there too to make sure we don't show red for USB and <= 10 MHz.

TreeTopHam:
> FreeDV will still change over to LSB when < 10 MHz and using the Analog button.
>Feedback wanted here.

If the 'Analog' button is used to check for a 'clear frequency' before transmitting RADEV2, shouldn't it always be monitoring the USB channel for all bands?

I'm not sure the mode indicator at the bottom of the screen captures my attention enough to count on it for this purpose.

tmiw:
> > FreeDV will still change over to LSB when < 10 MHz and using the Analog button.
> > Feedback wanted here.
> 
> If the 'Analog' button is used to check for a 'clear frequency' before transmitting RADEV2, shouldn't it always be monitoring the USB channel for all bands?
> 
> I'm not sure the mode indicator at the bottom of the screen captures my attention enough to count on it for this purpose.

Theoretically you should see something in the waterfall if the frequency is in use, so the Analog button isn't really needed.

On that note, I seem to remember a suggestion to get rid of the Analog button altogether. It wouldn't be a completely crazy idea since you could always just use the microphone attached to your radio if you wanted to operate analog SSB while FreeDV was running. I'm not sure how often that button is used, though, so we may get pushback if we got rid of it.

barjac:
> 
> Theoretically you should see something in the waterfall if the frequency is in use, so the Analog button isn't really needed.

Seeing 'something' is not very useful when it can't be identified.

> I'm not sure how often that button is used, though, so we may get pushback if we got rid of it.

I use it quite regularly but only to confirm that the RX audio path is working. Not essential, another method of quickly checking for a signal would suffice for that use case.

> Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.

How could that work when FreeDV is in hamlib 'read only' mode?
Or is that feature going to become obsolete as well in V2?

If the reporter displayed frequencies 1.5kHZ higher than the radio dial frequencies, then we would be correctly reporting the centre frequency of the rade spectrum. 
For those using FreeDV to control the rig it could set the dial frequency 1.5kHz lower than the display frequency.
This would be similar to what is already done by wsjtx in 'fake it' mode.
Operationally we would be displaying the spectrum centre frequency as with FM and AM.
What is displayed on the radio frequency readout would become immaterial (as it does in FT8 etc), as the software is in control.

Tyrbiter:
There is a long history of using the carrier frequency for SSB and other similar modes such as data modes that effectively operate in USB mode in the main. A long time ago I did try setting up one of my radios to display the modulation centre frequency, it was a total pain and I reverted to the normal approach very quickly.

I suspect that changing this is going to be an uphill task. WSJT-X here shows dial frequency and then has a wide graph with the AF offset.

Regarding analog operation, one reason I liked freedv-gui originally was that I could easily use the same USB-connected mic for both FreeDV and SSB use with just one application. I would hate to lose that capability, I don't have a mic connected to the front panel of any of my HF radios.

barjac:
> A long time ago I did try setting up one of my radios to display the modulation centre frequency, it was a total pain and I reverted to the normal approach very quickly.

I can understand that but I am suggesting that _FreeDV_ reports the centre frequency not the radio.
> 
> I suspect that changing this is going to be an uphill task. WSJT-X here shows dial frequency and then has a wide graph with the AF offset.

My analogy to wsjtx was not ideal I agree, but I don't agree that this change would be difficult if done now. The longer it is left the worse the problem will become as more direct methods of freedv modulation become more commonplace. 
> 
> Regarding analog operation, one reason I liked freedv-gui originally was that I could easily use the same USB-connected mic for both FreeDV and SSB use with just one application. I would hate to lose that capability, I don't have a mic connected to the front panel of any of my HF radios.

That does not have to be lost but any analogue operation needs to be centred around the freedv signal in the same mode as it is now.

**Incorporating a change to using reporting of centre frequencies:**
Assuming that you are on FreeDV Digital mode with a dial freq of 1987kHz, the freedv spectrum is currently centred around 1985.5kHz.
Freedv would display 1985.5kHz and the radio frequency display becomes of little interest as in FT8 etc.
The switch to USB (below 10MHz) would require the dial frequency to be set to 1984kHz to use the same freedv spectrum and FreeDV would still display 1985.5kHz.
Switching to analogue mode in FreeDV would simply require a mode change to LSB and a dial frequency change to 1987kHz. (via hamlib etc.)
While now in analogue mode freedv would display 1987kHz, which for e.g. checking the frequency would make any SSB operation readable and allow conversation in SSB natively with freedv centered in the SSB spectrum as set in digital mode.

All of the above could be done in the freedv-gui software with simple conditional addition and subtraction.

The 'suggested frequencies' table would be adjusted at the same version release to reflect the centre frequencies rather than the obsolete and meaningless suppressed carrier frequencies.

I think this is needed now and integrated into V2 so users can get used to it.

barjac:
> Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.

I think you meant +3kHz?

Tyrbiter:
> > Should probably add something to make operators realise that sub-10MHz analog operation will have a -3kHz frequency offset applied.
> 
> I think you meant +3kHz?

Depends which end of the telescope you are looking down.

TreeTopHam:
>If the reporter displayed frequencies 1.5kHZ higher than the radio dial frequencies, then we would be correctly reporting the centre frequency of the rade spectrum.

>For those using FreeDV to control the rig it could set the dial frequency 1.5kHz lower than the display frequency.

Letting users double-click a FreeDV Reporter entry to tune the radio, only to have the displayed dial frequency not match the Reporter frequency, would be very confusing.

tmiw:
> > If the reporter displayed frequencies 1.5kHZ higher than the radio dial frequencies, then we would be correctly reporting the centre frequency of the rade spectrum.
> 
> > For those using FreeDV to control the rig it could set the dial frequency 1.5kHz lower than the display frequency.
> 
> Letting users double-click a FreeDV Reporter entry to tune the radio, only to have the displayed dial frequency not match the Reporter frequency, would be very confusing.

Presumably we'd use the same standard everywhere (so center frequency is what would show in FreeDV Reporter as well as in the main window).

That said, I believe we discussed the center frequency thing before and decided not to do that.

barjac:
> > > If the reporter displayed frequencies 1.5kHZ higher than the radio dial frequencies, then we would be correctly reporting the centre frequency of the rade spectrum.
> > 
> > 
> > > For those using FreeDV to control the rig it could set the dial frequency 1.5kHz lower than the display frequency.
> > 
> > 
> > Letting users double-click a FreeDV Reporter entry to tune the radio, only to have the displayed dial frequency not match the Reporter frequency, would be very confusing.
> 
> Presumably we'd use the same standard everywhere (so center frequency is what would show in FreeDV Reporter as well as in the main window).

Maybe he meant the displayed frequency on the radio dial would not match the reporter and FreeDV frequency?
 
Exactly! At present nothing matches the transmitted frequency which is worse.
Do you watch the radio frequency when transmitting FT8 etc. it is working split frequency and changing but that's accepted.

> That said, I believe we discussed the center frequency thing before and decided not to do that.

Do you recall the reason?

tmiw:
New thought:

1. When not in Analog mode, FreeDV always uses upper sideband when changing the frequency (unless mode adjustment is disabled by the user).
2. When in Analog mode and mode adjustment is turned on, an additional drop-down becomes available just below the frequency list in the main window with two choices: USB or LSB. Pushing Analog on its own would not change the mode (other than DIGU->USB like it always has) but you can choose to switch to LSB if desired.

Note that no frequency adjustment would occur here since there'd still be the option of using LSB, albeit only in analog mode. How's that sound?

> Do you recall the reason?

I believe it was related to usability concerns as other modes use the dial frequency. Offhand, Winlink is probably the only one I can think of that uses center frequency on HF (but it also provides dial frequency for manual tuning).

TreeTopHam:
>I believe it was related to usability concerns as other modes use the dial frequency.

Since we are using the SSB mode to send digital voice, it makes sense to follow the suppressed carrier dial convention currently in use. As such, the dial freq marks the lower boundary of the USB 'channel' being used to contain the data. Knowing the channel boundary is also useful for helping avoid out-of-band emissions when operating close to a band edge.

barjac:
If potential operator incompetence is the reason to perpetuate this mis-reporting of our actual operating frequency then it is invalid.
How do you cope when operating USB at the top edge of any of the higher HF bands without your suppressed guard rail to protect you?
If you know your centre frequency and bandwidth then it is much easier to position your signal safely within the legal limits.
Yesterday myself and another station used a build of FreeDV with centre frequency reporting and this PR included. We had no issues operating on 80m at a centre frequency of 3630kHz with non-inverted rade and had listeners join in with interest. We paid no real notice of the dial frequency on the radios as it is fictional and related only to a 100year old convention related to a technical limitation of the time.

tmiw:
I had Claude implement a prototype of my previous suggestion in the latest commit. Here's what it looks like on macOS at least:

image

Some basic testing indicates that it's working but it probably needs to be exercised more if we go this route.

barjac:
I'm sorry but I read your suggestion last night four times but still did not understand it.
It was late! I will look again today :)
In the branch we were using last night to test the centre frequency reporting/readout, the analogue mode below 10Mhz changes the radio to LSB and moves the dial frequency up by 3kHz to keep the SSB centred on the freedv spectrum. This worked well when we tested it on air while watching a UK web SDR and switching back and forth.

barjac:
OK now I understand it, however:
1. Would anyone want to use USB SSB on 40m when switched to analogue? This could be automatic depending on band.
2. For LSB the frequency should move +3kHz to avoid effectively dropping 3kHz on to another QSO.
3. It wastes GUI real estate. If the option is really needed then a context menu off the 'Switch to analog' button defaulting to the current convention would suffice.

TreeTopHam:
>I had Claude implement a prototype of my previous suggestion in the latest commit. 

Nice! Easy way to check the USB channel for activity if using 160 - 40m.

tmiw:
> > I had Claude implement a prototype of my previous suggestion in the latest commit.
> 
> Nice! Easy way to check the USB channel for activity if using 160 - 40m.

If the drop down is set to LSB, should the frequency be offset as @barjac mentioned above? 

Though he does raise a good point--I don't know if anyone will want to operate USB on 40m for example. So maybe reverting the last commit and just adding frequency offset logic for the lower bands in analog mode is enough.

tmiw:
Reverted previous commit, added logic to shift frequency 3kHz if we're on one of the lower bands when the user pushes Analog. Hopefully that's okay with everyone?

TreeTopHam:
@tmiw:
>Reverted previous commit, added logic to shift frequency 3kHz if we're on one of the lower bands when the user pushes Analog. Hopefully that's okay with everyone?

Nice!

I have a few comments:

- I understand (and like) the 3 kHz shift and feel it is necessary to properly monitor (i.e. decode/understand) the content of the 'USB channel' for existing activity. 

- As currently implemented in 6cf11cb, however, I find the GUI response non-intuitive and confusing when I push 'Switch to Analog' because of the jump in frequency in the GUI and radio dial (remember, I understand why, but it's still non-intuitive). This jump will probably confuse a lot of users who haven't thought through the math of what is happening and why.

- Let's keep the 3 kHz functionality, but perhaps re-label the 'Switch to Analog' button to say "Channel Monitor" or convey something similar. We want the user to understand they are checking for a clear channel before transmitting RADE.

barjac:
I would strongly disagree with any  change to the 'Switch to Analog' button label.

There are lots of features in FreeDV that initially confuse users, but they learn.
Replacing a button label which would then imply it only has one narrow use case when it in fact has many would be wrong.
To quote from 3 days ago in this conversation:
"Regarding analog operation, one reason I liked freedv-gui originally was that I could easily use the same USB-connected mic for both FreeDV and SSB use with just one application. I would hate to lose that capability, I don't have a mic connected to the front panel of any of my HF radios."

TreeTopHam:
@barjac:
>I would strongly disagree with any change to the 'Switch to Analog' button label.

Yes, all good points!

I'm changing my vote. Let's leave as implemented in 6cf11cb :)

tmiw:
Sounds good. Anything else that needs to be tweaked here before merge (besides updating the user manual)?

tmiw:
OK, user manual updated. Let me know if any wording there needs to change as well.

TreeTopHam:
From the User Manual:
>As a result, to ensure that the user is monitoring the same sideband, FreeDV will add 3 kHz to the radio's VFO frequency when switching to Analog mode (and subtract 3 kHz when switching back to FreeDV reception).

_We add 3 kHz only if we are on the legacy LSB modes of 40, 80, & 160 meters._ 

We should make that clarification in the manual.

tmiw:
> From the User Manual:
> 
> > As a result, to ensure that the user is monitoring the same sideband, FreeDV will add 3 kHz to the radio's VFO frequency when switching to Analog mode (and subtract 3 kHz when switching back to FreeDV reception).
> 
> _We add 3 kHz only if we are on the legacy LSB modes of 40, 80, & 160 meters._
> 
> We should make that clarification in the manual.

Updated.

Tyrbiter:
Worth adding an example?

barjac:
I think that text is overcomplicated, I would suggest:

Note that if in Analog mode, FreeDV will use the appropriate sideband for the band.
To ensure that the same portion of spectrum is used by both digital and analog modes FreeDV will adjust the radio frequency by +3kHz on the legacy LSB bands. i.e. 40, 80 and 160 meters.

tmiw:
> Note that if in Analog mode, FreeDV will use the appropriate sideband for the band.
> To ensure that the same portion of spectrum is used by both digital and analog modes FreeDV will adjust the radio frequency by +3kHz on the legacy LSB bands. i.e. 40, 80 and 160 meters.

Updated along with example.

TreeTopHam:
>Updated along with example.

Update looks good :)

tmiw:
OK, I'll go ahead and merge this. We can do a followup PR if needed.

JohBod:
What about an "Analog LSB" button that makes FreeDV do a spectral inversion of the audio spectrum for both TX and RX when in analog mode? For example, a 2800 Hz wide slice of spectrum could be "mirrored" around 1500 Hz. Quite simple to do in DSP.

This would allow selectable USB/LSB analog operation _**without**_ dial frequency jumps and with the radio **_always_** in USB(-D) mode. Pushing the button is equivalent to moving the radio dial +3kHz and switch to LSB but it is done entirely in software.

The default state of this button could be configured with a column of checkboxes in the table of predefined frequencies.

tmiw:
> What about an "Analog LSB" button that makes FreeDV do a spectral inversion of the audio spectrum for both TX and RX when in analog mode? For example, a 2800 Hz wide slice of spectrum could be "mirrored" around 1500 Hz. Quite simple to do in DSP.
> 
> This would allow selectable USB/LSB analog operation _**without**_ dial frequency jumps and with the radio **_always_** in USB(-D) mode. Pushing the button is equivalent to moving the radio dial +3kHz and switch to LSB but it is done entirely in software.
> 
> The default state of this button could be configured with a column of checkboxes in the table of predefined frequencies.

I'm not sure if the added complexity is worthwhile outside of maybe those stations that can't use Hamlib or OmniRig for whatever reason. Interesting idea, though.

JohBod:
It was not really a suggestion, just an idea that popped up..

I was thinking of a baseband channel through which either RADE or USB/LSB voice could be pushed without the need to change frequency or sideband on the radio. The radio dial would stay on either Fcenter-1500 Hz, in case of a USB radio, or spot on Fcenter in case of an I/Q baseband radio.

Voice has no well defined center frequency so the 1500 Hz "voice spectrum mirroring point" was choosen arbitrarily. 

I had something like this in mind, figure 2a here: https://www.dsprelated.com/showarticle/37.php

It looks like an interpolation filter, then mix with 2 * F_mirror at 2 * Fs_original and finally a decimation filter to get back to Fs_original.

Issue #1391, opened 2026-06-12
Legacy Mode Removal: FreeDV 700D/E/1600


tmiw:
This issue is to remove support for Codec2-based modes from freedv-gui (700D/E/1600). Required actions:

1. Remove Codec2 library from freedv-gui source tree.
2. Remove configuration options: "Enable Legacy Modes", "Multiple RX" and last mode used.
3. Remove GUI elements and logic for mode selection and squelch (RADE does not support explicit squelch configuration).
4. fdv_pipeline_extra: Remove FreeDVReceive/TransmitStep as well as ParallelStep (as we no longer support multiple RX).

Expected behavior after implementation: Start Modem will always start RADEV1 with no ability to override.

tmiw:
Tried to remove Codec2 from the build process and it looks like we have a Codec2 dependency for the waterfall display. However, it's just something in modem_stats, so it may be possible to reimplement what's required (or simply copy modem_stats.c and remove everything else from Codec2).

tmiw:
All known work required to remove legacy modes has been completed, closing.

PR #1400, opened 2026-06-16
Optional Gnupg signed transmissions.

Supermagnum:
Added optional OpenPGP transmission signing via GnuPG. 
Signing of transmissions is not encryption,- so its perfectly legal.
Enable signed bookend frames on 700D/700E data channels so operators can verify callsign authenticity, with automatic keyserver lookup, status bar feedback, and user documentation.

OpenPGP is built into all Linux distributions,- its strange that it is not leveraged even if its free.
For more information,- please read: 
https://github.com/Supermagnum/gnupg-hamradio


tmiw:
Thank you for your PR. Unfortunately this is involving legacy modes that will be removed in FreeDV 3.0 (see https://github.com/drowe67/freedv-gui/issues/1391) so this PR will be closed. Definitely feel free to submit additional PRs or feature requests in the future, though!

PR #1403, opened 2026-06-19
Hamlib: Switch off memory channel before setting frequency/mode

barjac:
## Summary
- `rig_set_freq`/`rig_set_mode` fail with "Invalid parameter" when the rig's current VFO is MEM, since most CAT protocols can't set an arbitrary frequency/mode while a memory channel is selected.
- This breaks FreeDV pushing its saved frequency/mode whenever the rig is left parked on a memory channel — at startup, on double-click QSY from FreeDV Reporter, and when accepting an incoming QSY request.
- Adds `getWritableVfo_()`, which switches to VFO A first if the current VFO is MEM, and uses it in the two write paths (`setFrequencyImpl_`, `setModeImpl_`). Read-only paths (periodic polling, disconnect-time restore) are unchanged since querying works fine on any VFO.
- If `rig_set_vfo(RIG_VFO_A)` fails (rig doesn't support VFO switching), it falls back to the original VFO with a debug log — no worse than current behaviour for those rigs.

## Test plan
- [x] Verified on TS450S: frequency/mode pushes from a memory-channel start now succeed instead of failing.
- [x] Verified double-click QSY from FreeDV Reporter while radio is on a memory channel.

Thanks Claude

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -957,6 +957,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Right-justify SNR column in FreeDV Reporter to improve appearance. (PR #1387) - thanks @barjac!
     * Hamlib: Set frequency again on mode changes. (PR #1395)
     * Consolidate EOO length calculation in freedv-backend to improve callsign decode reliability. (PR #1402)
+    * Hamlib: Switch off memory channel before setting frequency/mode. (PR #1403) - thanks @barjac!
 2. Enhancements:
     * Add UDP broadcast of received callsigns. (PR #1367)
     * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398) - thanks @barjac!
--- a/src/rig_control/HamlibRigController.cpp
+++ b/src/rig_control/HamlibRigController.cpp
@@ -638,7 +638,7 @@ void HamlibRigController::setFrequencyImpl_(uint64_t frequencyHz)
         }
     }
 
-    vfo_t currVfo = getCurrentVfo_(); 
+    vfo_t currVfo = getWritableVfo_();
     setFrequencyHelper_(currVfo, frequencyHz);
 
     if (pttSet_)
@@ -722,7 +722,7 @@ void HamlibRigController::setModeImpl_(IRigFrequencyController::Mode mode)
         }
     }
 
-    vfo_t currVfo = getCurrentVfo_(); 
+    vfo_t currVfo = getWritableVfo_();
     setModeHelper_(currVfo, hamlibMode);
 
     if (pttSet_)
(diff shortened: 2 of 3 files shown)


tmiw:
I have access to a FT-817, KX3, IC-7200, QMX and Flex 8400. Do any of those have this issue? (I don't really use memory channels, so I'm not sure. And I'd like to be able to test here before merging if possible).

Also, think we need to make a similar change in the OmniRig handling?

barjac:
I gave your question to Claude: 

---
  I checked the Hamlib 4.7.0 backend source for each of those rigs (rigs/yaesu/ft817.c, rigs/icom/ic7200.c, rigs/kenwood/k3.c, rigs/kenwood/ts480.c for QMX,
  rigs/kenwood/flex6xxx.c). Short answer: none of them should be able to reproduce this bug, for two different reasons:

  The bug only triggers when Hamlib's get_vfo() actually returns RIG_VFO_MEM — which only happens via kenwood_get_vfo_if(), which parses the Kenwood IF;
  response and reports MEM when a specific byte is '2' (real hardware state — this is what the TS450S hits).

  - FT-817: chan_list is empty and get_vfo has no MEM concept at all — Hamlib can never know it's on a memory channel.
  - IC-7200: get_vfo is commented out entirely in the rig caps (//.get_vfo = icom_get_vfo,). With no get_vfo, we fall back to RIG_VFO_CURR, never MEM.
  - KX3: get_vfo just returns Hamlib's cached current_vfo — it doesn't poll the rig, so it won't read back as MEM just because the front panel is on a
  memory channel.
  - QMX: uses the real polling kenwood_get_vfo_if, same as TS450S, but its chan_list has no memory channels defined — the rig will never report MEM.
  - Flex 8400 (Hamlib "6xxx" model via SmartSDR CAT emulation): same kenwood_get_vfo_if, also no memory channels in chan_list.

  So you likely won't be able to reproduce or test this on the hardware you have — it's specific to Kenwood/Elecraft-protocol rigs that both poll real VFO
  state and have memory channels enabled in Hamlib (TS450S confirmed, possibly other Kenwood-family rigs like TS-570/K2/K3/K4 too).

  On OmniRig: I don't think the same fix applies there. OmniRig's VFO enum (RigParamX) only has PM_VFOA/B/AA/AB/BA/BB/UNKNOWN — there's no memory-channel value exposed by the API at all, so it has no way to represent "parked on MEM" the way Hamlib does. It already has a different fallback for the unknown-VFO case: in OmniRigController::setFrequencyImpl_, if get_Vfo() returns anything other than A or B, it falls through to a VFO-less put_Freq() call instead of failing outright. A literal port of getWritableVfo_() doesn't map cleanly here since there's nothing to detect/switch off. I'd hold off on an OmniRig change unless someone reports an actual failure that traces back to this.

PR #1405, opened 2026-06-19
Pin PTT button text colour explicitly instead of relying on GTK theme

barjac:
## Summary
- `m_btnTogPTT` only ever had `SetBackgroundColour()` called on it (red for TX, amber for the TX->RX drain, null for idle); the text colour was always left to the active GTK theme to decide.
- On wxGTK < 3.3 (e.g. Mageia 9), the main frame losing window activation while the TOT warning dialog has focus puts the button into GTK's "backdrop" state, dimming the intended black text to grey. On other themes the default foreground can resolve to white, making it unreadable against the red/amber background in all states.
- Explicitly pin `SetForegroundColour(*wxBLACK)` alongside every place the background is set to red/amber, and reset it back to `wxNullColour` (theme default) alongside every place the background is reset to idle, mirroring the existing background-colour pattern.

## Test plan
- [x] Built and verified locally
- [x] Tested on Mageia 9 (wxGTK < 3.3) — confirms PTT button text stays readable through the TOT warning dialog and during the TX/amber/idle states

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -960,7 +960,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Hamlib: Switch off memory channel before setting frequency/mode. (PR #1403) - thanks @barjac!
 2. Enhancements:
     * Add UDP broadcast of received callsigns. (PR #1367)
-    * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398) - thanks @barjac!
+    * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398, #1405) - thanks @barjac!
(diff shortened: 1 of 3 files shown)


tmiw:
I had tried something similar before but it didn't work on macOS due to OS/GUI framework limitations. If we're okay with this being a Linux (and possibly Windows) only change, we can still go with this.

barjac:
> I had tried something similar before but it didn't work on macOS due to OS/GUI framework limitations. If we're okay with this being a Linux (and possibly Windows) only change, we can still go with this.

Good.
This will merge on top of 1393 and 1397 cleanly and in test on Mageiea 9 (wxGTK 3.2) it does solve the issue where the timeout pop-up was causing grey text on the button. It also does not cause any regression in wxGTK-3.3,

PR #1406, opened 2026-06-20
Fix spacebar PTT restarting TX immediately after TOT fires while key is held

barjac:
Built on top of #1393 at commit 3585eca0 (current tip, including the unrelated colour fixes from #1405).

## Bug

In momentary PTT mode, if the spacebar PTT key is still physically held down when the Time-Out Timer (TOT) fires (e.g. something resting on the keyboard), TX restarts immediately afterward instead of staying off — defeating the safety purpose of the TOT.

## Root cause

`OnTOTTimer` force-stops TX without regard to whether the spacebar key is still physically held. On Linux/GTK, a key held under sustained pressure can generate real `KEY_UP`/`KEY_DOWN` event pairs at the OS repeat rate rather than one sustained `KEY_DOWN`, and `wxKeyEvent::IsAutoRepeat()` is unreliable at the point `wxApp::FilterEvent` intercepts them. The next repeat-driven `KEY_DOWN` then sees the key as "freshly pressed" and immediately re-keys TX.

This PR contains two commits:

1. **Revert of `add9e6e4`** ("Try different way of filtering out duplicate PTT keypresses"). That commit was itself a second attempt at fixing this same pre-existing bug, but it compounded the problem — TX would interrupt the TOT warning countdown after a couple of seconds and restart immediately (confirmed via direct testing, see #1393 comment thread).
2. **The actual fix**: instead of inferring "still held" from the key event stream, poll `wxGetKeyState()` (a direct OS query) after a TOT-forced stop, and latch out spacebar restart until it confirms a genuine key release.

The fix only engages from inside `OnTOTTimer`'s forced-stop path, so normal momentary-PTT hold/release behavior (no timeout reached) is unaffected.

Tested locally via repeated TOT-timeout-while-held scenarios, as well as normal (non-TOT) momentary PTT use, with no regressions observed.

Above analysis and fix by Claude Code.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1277,6 +1277,8 @@ MainFrame::MainFrame(wxWindow *parent) : TopFrame(parent, wxID_ANY, _("FreeDV ")
     Bind(wxEVT_TIMER, &MainFrame::OnTOTTimer, this, ID_TIMER_TOT);
     m_totWarningTimer.SetOwner(this, ID_TIMER_TOT_WARNING);
     Bind(wxEVT_TIMER, &MainFrame::OnTOTWarningTimer, this, ID_TIMER_TOT_WARNING);
+    m_pttKeyPollTimer.SetOwner(this, ID_TIMER_PTT_KEY_POLL);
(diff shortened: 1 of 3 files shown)


tmiw:
Must be another platform-specific thing since both attempts at fixing did work for me on macOS. Oh well.

PR #1414, opened 2026-06-24
Start TX on XMIT button press rather than release; fix related edge cases

barjac:
Tested on Linux (wxGTK 3.3, Mageia 10). This applies to PR #1393 @ de8f7e4
**Repro for the original issue:** latching mode, press and hold XMIT — TX does not start until release.

I have tested this in every possible scenario that I can think of and I cannot now break it. The TX now fires on button down simultaneously with the RED button colour change and no flicker. 

## Changes

- Rebind XMIT button to `wxEVT_LEFT_DOWN` (reverts 767a7979); call `togglePTT()` directly on press. In latching mode a `m_suppressNextPTTClick_` flag discards the redundant `TOGGLEBUTTON_CLICKED` that fires on release, with the button value pre-set before the call so the suppress handler can undo the widget's auto-toggle safely at any point in the TX delay.
- Momentary mode is not suppressed — the release click stops TX as expected. Mid-changeover mouse release deferred via `m_momentaryKeyReleasedDuringChangeover_`, mirroring the existing keyboard path.
- `OnTogBtnPTTMouseLeave` clears suppress unconditionally: `TOGGLEBUTTON_CLICKED` never fires if the mouse leaves before release, so without this the flag would persist and silently eat a later spacebar or click event.
- Replace `wxNullColour` resets on the XMIT button with `wxSystemSettings::GetColour(wxSYS_COLOUR_BTNFACE/BTNTEXT)`. Calling `SetBackgroundColour(wxNullColour)` removes the GTK CSS provider installed by the red TX colour, leaving the widget in a different state from buttons that were never styled — causing a blue hover tint after the first TX cycle. Keeping an explicit provider prevents this.
- In `OnTOTTimer`, set `m_suppressNextPTTClick_` in momentary mode to prevent the held mouse button restarting TX after TOT expiry, mirroring the existing `m_pttKeyRequireRelease_` keyboard guard. `OnTogBtnPTTMouseDown` clears any stale suppress at entry so this cannot eat a later intentional click.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1039,6 +1039,10 @@ setDefaultMode:
     
     m_togBtnAnalog->Disable();
     m_btnTogPTT->Disable();
+    m_btnTogPTT->SetBackgroundColour(wxNullColour);
+#if !defined(__APPLE__)
+    m_btnTogPTT->SetForegroundColour(wxNullColour);
+#endif // !defined(__APPLE__)
     m_togBtnVoiceKeyer->Disable();
 
     // squelch settings
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2893,6 +2893,10 @@ void MainFrame::performFreeDVOff_()
 
         m_togBtnAnalog->Disable();
         m_btnTogPTT->Disable();
+        m_btnTogPTT->SetBackgroundColour(wxNullColour);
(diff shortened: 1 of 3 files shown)


tmiw:
Seems to work on two different Linux distros and macOS, merging.

PR #1418, opened 2026-06-30
Move all FreeDV configuration to Settings->Edit Settings.

tmiw:
I used Claude to prototype moving the configuration currently in Easy Setup, Audio Config and CAT and PTT Config windows to Settings->Edit Settings. For CAT control, a new Radio tab now exists as follows:

image

Audio device configuration is now in the Audio tab:

image

Not sure if this makes FreeDV setup any easier, so comments welcome.


tmiw:
Pinging @peterbmarks, @Tyrbiter and @barjac to see if they have any feedback.

BTW on macOS, Tools->Options automatically gets moved to the FreeDV menu to match Apple's convention for other apps. Unfortunately Command-, doesn't work to open this as expected so that'll need more investigation prior to merge.

Also need to test on Windows to make sure the OmniRig configuration options made it over (and work properly).

EDIT: Command-, now works.

peterbmarks:
The "Options" window should be called "Settings" and be in the application menu.
The screenshots above look really great to me.
Suggest the tab order should be "Audio", "Radio", "Rig Control" as a more logical flow.
Great stuff!

tmiw:
Renamed to Settings and (on macOS) moved to the app menu:

image

New tab ordering and Audio tab (where the labels were also adjusted so people know which set of devices is which):

image

For the Test buttons, the labels change to "Recording" and/or "Playing" during the test. When testing the input devices, FreeDV will first record some audio and then play it back on the user's speaker/headset.

tmiw:
Thinking some more, we probably do still need some sort of wizard type thing for first start. Could just be the Radio tab + the audio device stuff in the Audio tab plus callsign/locator entry. I'll see if I can get Claude to generate that dialog later.

peterbmarks:
I think it would be good if we don't scare people off by asking too many questions up front. For example a new user might just want to listen in by decoding audio from the rig with no CAT control at first. Later they could go back and set up the other things.

tmiw:
Here's what Claude proposed:

image

image

image

image

We can definitely narrow down what's asked for sure, but not sure if we should go all the way to only asking the minimum needed for RX. Feedback from others would be good here.

barjac:
I agree with Peter, Settings should be a top level menu:

File  Settings  Tools  Help

IMHO projects rarely make good use of the top level menu items.
To make it simpler for new users the wizard could be started with Setup, direct and obvious:

File  Setup  Settings  Tools  Help

--------
 
I don't like the new Audio settings GUI at all! Massive regression!
The original needed some adjustment to the layout but now it has lost all visibility of the devices on screen together, no description of what the settings are for, "Output from computer to speaker" etc. No sample rate shown. Useless and confusing!

Please just keep the current one with some separation of the upper and lower sections on each tab. The current audio settings dialogue is currently the best possible and least confusing way of doing it, in fact I looked at trying FreeDV several times prior to that page being redesigned to how it is now over many years and gave up each time. It was only with the current design that it began to make sense.

---------
The voice keyer settings need to be in their own tab.

peterbmarks:
Barry, that's not what I meant.
On macOS Settings is always in the Application menu with a keyboard shortcut of Command-Comma.

barjac:
> Barry, that's not what I meant. On macOS Settings is always in the Application menu with a keyboard shortcut of Command-Comma.

I have no knowledge of Mac.
I am talking only about the FreeDV top menu bar.
Are you saying that the main system programs menu in Mac has something that tries to hook into running applications and control them?

TreeTopHam:
@barjac 
>I don't like the new Audio settings GUI at all! Massive regression!
The original needed some adjustment to the layout but now it has lost all visibility of the devices on screen together, no description of what the settings are for, "Output from computer to speaker" etc. No sample rate shown. Useless and confusing!

>Please just keep the current one with some separation of the upper and lower sections on each tab. The current audio settings dialogue is currently the best possible and least confusing way of doing it...

I agree with @barjac 

Audio setup can be tricky sometimes, but the current phrases in the renaming proposal make it even more confusing.  I'd vote to leave Audio Config as it currently exists.

barjac:
The wizard currently segfaults for me in Linux but the four step approach in the above images looks good.

tmiw:
How about this to more closely match the Setup Wizard?

image

(Haven't tweaked anything else yet other than add the sample rate that will be used. For context, FreeDV doesn't really give a choice of sample rates anymore, so there's no real need to allow users to enter one--but it probably doesn't hurt to display what it's going to use.)

tmiw:
New menu bar as of the latest in this PR:

### "Application Menu" (macOS)

image

### File

image<br/><br/>

### Settings 

image

(macOS)

image

(Linux/Windows)

### Tools

image

### Help

image

(macOS)

image

(Linux/Windows)

tmiw:
Audio tab only has audio devices now:

image

Voice keyer options are now in a "Voice Keyer" tab:

image

tmiw:
> The wizard currently segfaults for me in Linux but the four step approach in the above images looks good.

This is also fixed as of the latest in this PR.

tmiw:
Audio device labels now match what was in Tools->Audio Config before:

image

EDIT: Same thing in Setup Wizard:

image

image

TreeTopHam:
Commit a23a113 on **Windows 11** crashes in the following use cases:
Launch GUI, but don't start modem, then Settings -> Edit Settings...  crashes
Launch GUI, but don't start modem, then Settings ->  Setup Wizard -> ... Finish -> crashes
Launch GUI, Start Modem -> Settings -> Edit Settings -> Radio -> Test PTT -> Error (could not open serial port) - but the serial port is really there

I stopped testing at this point...

Windows Crash Dump uploaded to Dropbox:
https://www.dropbox.com/scl/fi/xbtxsgeretdtvw4l1ld4r/freedv.exe.16464.dmp?rlkey=vnx935nt4hbsmcfv0rhunny7s&st=a8ou62iv&dl=0

Commit db28ffe **Win11** also crashes as above.

barjac:
> Audio device labels now match what was in Tools->Audio Config before

Much better but you can't look through both the lists at the same time. Using pull-downs is not as convenient as having the lists as they were previously.

Now the main settings option crashes!

The wizard does run now but with an existing config the RX out (headphone) and TX out to transmitter are swapped over.

Closing the wizard at the end, closes or crashes (too late here to check) the whole program.

I still think that Settings should open the tabbed settings dialogue directly and that the wizard should be directly under Setup. The config export etc. under tools with the Playback etc. as before. Why have to drill down menus when you can access directly from the top level for commonly used or easy access like to the wizard?

tmiw:
OK I think I fixed the crashes and audio devices being swapped.

Also, the Audio tab has been split:

image

image

barjac:
Yes that is much better.

Please open the 'Settings' dialog directly from the top level 'Settings' menu item.
Please open the 'Setup wizard' directly from a top level 'Setup' menu item.
Then put the 3 'configuration' items (export/import/use) under Tools.
That will remove the need for a combo box under Settings (and Setup).

If I want to change something in Settings why would I want to see a long list of unrelated items to choose from and use an extra click?

BTW the Settings dialogue is still headed 'Options' not 'Settings'.

The 'RX Audio' and 'TX Audio' dialogues are not expandable to full screen which for long lists of audio devices with long names would be useful. The dialogues in the wizard are expandable.

TBH the original Wizard was much simpler to use regarding the audio devices. IIRC it just asked for two audio devices and was smart enough to set them up correctly. The audio settings in this wizard are actually now very similar to the audio dialogues in Settings which with the new layouts are probably easier to use.

The audio test buttons crash the program.
Assertion 'm' failed at ../src/pulse/thread-mainloop.c:176, function pa_threaded_mainloop_lock(). Aborting.

tmiw:
> Please open the 'Settings' dialog directly from the top level 'Settings' menu item. Please open the 'Setup wizard' directly from a top level 'Setup' menu item. Then put the 3 'configuration' items (export/import/use) under Tools. That will remove the need for a combo box under Settings (and Setup).
> 
> If I want to change something in Settings why would I want to see a long list of unrelated items to choose from and use an extra click?

To clarify, the top level menus (on Linux and Windows, anyway) should be the following, right?

File Setup Settings Tools Help

(and Setup/Settings should have no menus under them?)

Considering that the setup wizard is something that will probably not be accessed often, could we perhaps move it back to the Tools menu? 

> BTW the Settings dialogue is still headed 'Options' not 'Settings'.

Oops, will fix that now.
 
> The 'RX Audio' and 'TX Audio' dialogues are not expandable to full screen which for long lists of audio devices with long names would be useful. The dialogues in the wizard are expandable.

I was able to make the window itself be resizable but it doesn't seem to be resizing the list controls. This will probably require more investigation.

> TBH the original Wizard was much simpler to use regarding the audio devices. IIRC it just asked for two audio devices and was smart enough to set them up correctly. The audio settings in this wizard are actually now very similar to the audio dialogues in Settings which with the new layouts are probably easier to use.

It asked for three devices, actually (microphone, speaker and radio). Unfortunately the logic to link up the RX and TX radio devices never worked all that well (and was basically impossible if you used an SDR setup that needed two separate audio devices), so I'm not sure I want to go back to that.

That said, with an empty config it should at least be defaulting the microphone and speaker devices to whatever the user has set as their defaults in their OS.

> The audio test buttons crash the program. Assertion 'm' failed at ../src/pulse/thread-mainloop.c:176, function pa_threaded_mainloop_lock(). Aborting.

Fixed. For some reason, `IAudioEngine` doesn't need to be started/stopped on macOS but does on other platforms (and Claude missed the part where it needed to add that logic).

tmiw:
OK, the device lists should now be resizing. I also added a button to refresh the device lists:

image

TreeTopHam:
- When clicking the top-level Settings menu, I'd like to see the Settings window open directly without needing to click a second time on 'Edit Settings'. Perhaps you're already working on this...

tmiw:
I'm actually thinking of combining Simulation with Debugging since I think only one of the Simulation settings actually does anything now. Might make the Mint problem moot.

TreeTopHam:
Yes, and I realized that you have allowed resizing of the Settings window, which also makes my previous Linux Mint comment moot (Edited out)   :)

barjac:
> > Please open the 'Settings' dialog directly from the top level 'Settings' menu item. Please open the 'Setup wizard' directly from a top level 'Setup' menu item. Then put the 3 'configuration' items (export/import/use) under Tools. That will remove the need for a combo box under Settings (and Setup).
> > If I want to change something in Settings why would I want to see a long list of unrelated items to choose from and use an extra click?
> 
> To clarify, the top level menus (on Linux and Windows, anyway) should be the following, right?
> 
> File Setup Settings Tools Help

Yes exactly.

> 
> (and Setup/Settings should have no menus under them?)

Correct.

> 
> Considering that the setup wizard is something that will probably not be accessed often, > could we perhaps move it back to the Tools menu?

My thought was to have it directly visible to make it the obvious first port of call for new users who don't read the manual.

> 
> > BTW the Settings dialogue is still headed 'Options' not 'Settings'.
> 
> Oops, will fix that now.
> 
> > The 'RX Audio' and 'TX Audio' dialogues are not expandable to full screen which for long lists of audio devices with long names would be useful. The dialogues in the wizard are expandable.
> 
> I was able to make the window itself be resizable but it doesn't seem to be resizing the list controls. This will probably require more investigation.
> 
> > TBH the original Wizard was much simpler to use regarding the audio devices. IIRC it just asked for two audio devices and was smart enough to set them up correctly. The audio settings in this wizard are actually now very similar to the audio dialogues in Settings which with the new layouts are probably easier to use.
 
> It asked for three devices, actually (microphone, speaker and radio). Unfortunately the logic to link up the RX and TX radio devices never worked all that well (and was basically impossible if you used an SDR setup that needed two separate audio devices), so I'm not sure I want to go back to that.

I can't really comment more on the wizard as I have never used it in anger ;)

> 
> That said, with an empty config it should at least be defaulting the microphone and speaker devices to whatever the user has set as their defaults in their OS.

If you say so :) I would have thought the system defaults were the last place to try to use for FreeDV as in most systems the browser grabs the default output device on startup and screws up FreeDV. I leave the system defaults alone and use external USB audio cards for FreeDV. 

> 
> > The audio test buttons crash the program. Assertion 'm' failed at ../src/pulse/thread-mainloop.c:176, function pa_threaded_mainloop_lock(). Aborting.
> 
> Fixed. For some reason, `IAudioEngine` doesn't need to be started/stopped on macOS but does on other platforms (and Claude missed the part where it needed to add that logic).

Will test now.

tmiw:
The audio device column now takes up all remaining space:

image

Additionally, I removed the Simulation tab and moved the one item that still works to Debugging:

image

> > > Please open the 'Settings' dialog directly from the top level 'Settings' menu item. Please open the 'Setup wizard' directly from a top level 'Setup' menu item. Then put the 3 'configuration' items (export/import/use) under Tools. That will remove the need for a combo box under Settings (and Setup).
> > > If I want to change something in Settings why would I want to see a long list of unrelated items to choose from and use an extra click?
> > 
> > 
> > To clarify, the top level menus (on Linux and Windows, anyway) should be the following, right?
> > File Setup Settings Tools Help
> 
> Yes exactly.
> 
> > (and Setup/Settings should have no menus under them?)
> 
> Correct.

I looked into this a bit and I didn't see an obvious way of making a top-level item be effectively clickable. I'm sure I'm missing something, though, so I'll keep looking.

> > Considering that the setup wizard is something that will probably not be accessed often, > could we perhaps move it back to the Tools menu?
> 
> My thought was to have it directly visible to make it the obvious first port of call for new users who don't read the manual.

In theory it's supposed to come up the first time someone starts FreeDV with a fresh configuration, so they'd do the needed configuration at that time (or close it out and use Settings->Edit Settings if they'd rather). Going through the wizard later seems like it'd be unlikely after that point.

tmiw:
Seems like having top-level menus be clickable might be possible (https://forums.wxwidgets.org/viewtopic.php?t=48451) but it's not recommended. Also, I'm not sure how to give it the right ID so it'd be correctly moved to the app menu on macOS. I'll leave the menus as-is for now.

tmiw:
Shortened tab titles to RX/TX Audio, which should resolve @TreeTopHam's finding (regardless of the ability to resize the window):

image

TreeTopHam:
- I like the shortened labels & fewer tabs :)

- Would it make sense to rename 'Filter Audio' to 'Audio EQ' or something similar?

tmiw:
> * Would it make sense to rename 'Filter Audio' to 'Audio EQ' or something similar?

Not sure since there are some other non-EQ related features in that dialog:

image

barjac:
I have always thought that the upper and lower parts of the Audio config screens were too close to each other and confusing to read. Can we take this opportunity to add some more separation between them?
Patch:
[bcj-audio-tab-gap.patch](https://github.com/user-attachments/files/29640145/bcj-audio-tab-gap.patch)

Without patch:

Screenshot_20260703_153004

With patch:

Screenshot_20260703_153321

tmiw:
> I have always thought that the upper and lower parts of the Audio config screens were too close to each other and confusing to read. Can we take this opportunity to add some more separation between them? 

Done.

PR #1419, opened 2026-07-02
Suppress button flicker in Linux light themes 

barjac:
The light blue GTK 'button-down' in the very common Breeze theme under linux is rather annoying as it appears as a white/blue/red flicker when pressing the XMIT (PTT) button in the GUI. 

This is not noticed when using dark themes as the blue is not used.

I used Claude to work through this problem and a full explanation and reasoning is below:

 Problem
  On Linux with GTK3 themes that render a distinct :active (button-pressed) state — notably Plasma 6's Breeze GTK theme — FreeDV's buttons flash a pale-blue colour for a frame on every click. This is purely cosmetic but distracting, and doesn't match the rest of the native look wxWidgets otherwise achieves via GTK theming.

  Approach
  Rather than restyling buttons manually (which would fight the native theme engine and need per-theme tuning), the fix asks the current GTK theme what colour it uses for a button's resting state, then injects a targeted CSS override that forces the :active state to use that same colour. This neutralizes the flash while leaving every other aspect of the theme (hover, focus, disabled, etc.) untouched.

  Implementation, in src/main.cpp (SuppressButtonPressFlicker_(), called once from MainApp::OnInit() after the frame is up, guarded by #ifdef __WXGTK__ so it's a no-op on macOS/Windows):

  1. Determine the theme's normal button background. A throwaway GtkButton is built inside an offscreen GtkOffscreenWindow (so nothing is drawn on screen) purely so its GtkStyleContext can be queried via gtk_style_context_lookup_color(). It tries the named colour variable theme_button_background_normal first (Breeze-specific), falling back to the more broadly-supported theme_bg_color. theme_base_color was deliberately avoided — it resolves to the text-entry background (usually white), which would look wrong. If neither variable exists on a theme, the function bails out and does nothing — no flash suppression, but no breakage either.
  2. Install a CSS override. With a colour in hand, it builds a small CSS snippet (button:active { background-color: ...; background-image: none; 
  box-shadow: none; }) and loads it into a GtkCssProvider, which is registered screen-wide via gtk_style_context_add_provider_for_screen() at
  GTK_STYLE_PROVIDER_PRIORITY_APPLICATION. This priority is high enough to beat the theme's own :active rule for this specific property set, but doesn't
  touch anything else the theme defines, and doesn't require per-widget styling.
  3. Cleanup. The offscreen window/button are destroyed immediately after the colour lookup; the CSS provider is unreferenced after being handed to GTK (GTK
  retains its own reference once registered).

  Build changes
Since main.cpp now directly includes <gtk/gtk.h> and calls GTK APIs, src/CMakeLists.txt needed to expose GTK3 to the freedv target on Linux:
  - pkg_check_modules(GTK3 QUIET gtk+-3.0) locates GTK3 via pkg-config.
  - target_include_directories(freedv PRIVATE ${GTK3_INCLUDE_DIRS}) — wxWidgets' USE_FILE doesn't reliably propagate GTK's transitive include paths to dependent targets, so this was added explicitly after an initial build failure (missing gtk/gtk.h).
  - target_link_libraries(freedv ${GTK3_LIBRARIES}) — added after a second build failure (undefined reference to gtk_container_add at link time), since the include-path fix alone doesn't pull in the libraries.

  The whole block is wrapped in if(LINUX) / if(GTK3_FOUND) so it's inert on other platforms or if pkg-config/GTK3 isn't available.

  Why this approach vs. alternatives

  - Editing the system GTK theme CSS directly was rejected — it's out of scope for an application to modify, and wouldn't survive theme updates or apply
  per-user.
  - A static/hardcoded override colour was rejected in favour of querying the live theme, so the fix adapts automatically to light/dark variants and other GTK themes that expose the same named colours, rather than being Breeze-only.
  - Graceful degradation was a deliberate design goal: on themes without the expected colour variables, the function simply does nothing rather than guessing or falling back to a hardcoded colour that might clash.


src/CMakeLists.txt
```diff
@@ -111,6 +111,19 @@ if(LINUX AND STATIC_LIBGCC)
     target_link_options(freedv PRIVATE -static-libgcc -static-libstdc++)
 endif(LINUX AND STATIC_LIBGCC)
 
+# On Linux/GTK builds, make the GTK3 headers directly accessible in main.cpp
+# for the button-flash suppression code.
+if(LINUX)
```
tmiw, line 116: This would make libgtk3-devel (or whatever your distro calls it) required, right? IIRC it's possible to have the wxWidgets devel package installed without the GTK one.

Can we have the PR updated so that this workaround isn't used if we can't find GTK headers?

tmiw: Never mind, implemented the change here.

PR #1428, opened 2026-07-06
Only preserve previously selected tab on TX if it's in the same group as 'From Mic'.

tmiw:
I received a report of an issue using multiple tab groups:

1. Start with Waterfall as the active plot. Start and then immediately stop transmitting. FreeDV will show From Mic and then show Waterfall again.
2. Move Spectrum and SNR plots to the bottom half of the window. Click on the inactive one of those two to make it active.
3. Toggle TX again and then stop transmitting. FreeDV doesn't go back to the Waterfall tab when TX stops.

This is because we're unconditionally saving the previous tab instead of the previous tab that's in the same window as From Mic. This PR adjusts this behavior to only save the previous tab if it happened to be in the same group as the From Mic tab.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -962,6 +962,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Tighten remaining audio settings to improve audio drops. (PR #1412)
     * Fix corruption of string config arrays containing commas. (PR #1417) - thanks @barjac!
     * Suppress button flicker in Linux light themes. (PR #1419, PR #1421) - thanks @barjac!
+    * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
 2. Enhancements:
     * Add UDP broadcast of received callsigns. (PR #1367)
     * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398, #1405) - thanks @barjac!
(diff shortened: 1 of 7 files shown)


barjac:
That appears to work fine for me. With 'Frm Mic', 'Frm Decoder' and 'Frm Radio' in the same group, TX switches to Frm Mic and always returns to whichever of 'Frm Decoder' and 'Frm Radio' was previously selected before the TX.

tmiw:
> That appears to work fine for me. With 'Frm Mic', 'Frm Decoder' and 'Frm Radio' in the same group, TX switches to Frm Mic and always returns to whichever of 'Frm Decoder' and 'Frm Radio' was previously selected before the TX.

The person who originally emailed me about this found a scenario where it's still breaking:

1. Move the Waterfall tab and at least one other tab not "Frm Mic" to the bottom pane.
2. Click on that other tab to make it active.
3. Go to TX then RX.

Expected behavior: bottom pane isn't touched.
Actual behavior: bottom pane switches to the Waterfall tab.

What we probably need to do is to make the "last" tab invalid on RX->TX transition if the active tab isn't in the same group as "Frm Mic" and suppress switching the tab on TX->RX if invalid.

tmiw:
Hopefully this works better for you as well, @barjac?

barjac:
No, that's now broken it.
If I select 'Frm decoder' in RX and then TX, Frm mic is selected. OK
On return to RX 'Frm decoder' is correctly selected. OK
If I then change to 'SNR' from 'Waterfall' in the right window during RX and then TX, on the next RX nothing is now selected in the left window. It should be 'Frm Decoder' :( 

image

I didn't fully follow what this was trying to fix but no time now, hay to turn :)

tmiw:
OK, I *think* it's good now?

barjac:
> OK, I _think_ it's good now?

Yes I can't make it do anything wrong now, but then I had not seen an issue with it before this ;)

PR #1431, opened 2026-07-13
Fix window position restore under KWin and labwc (main window + FreeDV Reporter window)

barjac:
Branch: `fix/window-position-restore-race` (pushed to `barjac/freedv-gui`)
Target: `drowe67/freedv-gui` `master`

## Title

Fix window position restore under KWin and labwc (main window + FreeDV Reporter window)

### Summary

- Fixes the main window and FreeDV Reporter window not restoring to their saved
  on-screen position on startup under KWin (Plasma 6) and labwc (a wlroots-based
  Wayland compositor, tested on Raspberry Pi 4/Debian trixie). Both apply their
  own placement decision asynchronously, at no fixed, client-observable event —
  overriding whatever position the client already requested, sometimes even
  *after* the position has already happened to match the target once. The fix
  re-asserts the requested position for a couple of seconds after the window is
  first mapped (long enough to catch and correct a delayed override on either
  platform), then stops so it won't fight the user's own later repositioning.

- Fixes a related, previously-unnoticed bug this surfaced: the FreeDV
  Reporter window's saved position would drift upward by about one title-bar
  height on every close/restart cycle. Root cause: `topFrame_OnClose()` /
  `~MainFrame()` correctly grab and save the reporter's final position right
  before closing it, but `Close()`/`Destroy()` themselves fire one more
  `wxEVT_MOVE` reporting a position with the title bar's height already
  stripped off, and `FreeDVReporterDialog::OnMove` unconditionally overwrites
  the saved config value on every move event — so that stray event silently
  clobbered the correct save. Fixed by disconnecting `OnMove`
  (`stopTrackingPosition()`) right after the deliberate save, before
  `Close()`/`Destroy()`.

- Extracted the restore logic into a shared utility
  (`src/gui/util/WindowPositionRestore.h/.cpp`) used by both windows instead
  of duplicating the GTK-specific plumbing.

### Platform scope

- All new logic only compiles under `#if defined(__WXGTK__) && defined(HAS_GTK3)`
  (Linux/GTK3 builds). Windows and Mac fall through to the same
  `CallAfter`-deferred `Move()` that was already in place before this change,
  so their behavior is unchanged.
- The reporter drift fix (`stopTrackingPosition()`) is platform-neutral and
  applies everywhere; it's purely defensive (disconnects a handler after the
  correct value is already saved), so it should be a no-op on platforms that
  don't exhibit the stray-move-on-close behavior.
- Also added a loud `message(WARNING ...)` in `src/CMakeLists.txt` for the case
  where GTK3 development files (or pkg-config itself) aren't found at configure
  time — previously this silently disabled `HAS_GTK3` and, with it, all of
  this fix, with zero build-time signal that anything was skipped.
- Built and tested on: Mageia 10 (x86_64, wxGTK 3.3), KWin/Plasma 6 (two
  machines, both X11 and Wayland+XWayland sessions); Raspberry Pi 4 (aarch64,
  wxGTK 3.2), Debian trixie, labwc compositor (XWayland); and a Manjaro VM
  after an update. No Windows/Mac hardware available to test directly, but
  the platform-guard analysis above means their code path is unchanged from
  current master.

### How this was diagnosed

The KWin case: two earlier fix attempts based on a "the WM won't honor a
not-yet-mapped window's position" theory (deferring `Move()` via `CallAfter()`,
then via GTK's `map-event` alone) both failed. Root cause was only found via
targeted instrumented logging (`GetPosition()` at construction, at every
WM-triggered `configure-event`, and at close time) — which showed the WM's
placement override happens *after* the move already succeeds, not before it,
so no amount of moving earlier could win that race. An initial fix based on
watching `configure-event` and stopping as soon as the position matched fixed
this case.

The labwc/RPi4 case (found afterward, during separate testing on a different
machine/WM) exposed a further issue layered on top of the same disease:
1. A use-after-free crash appeared, found via `gdb` on a core dump: the
   original design shared one heap-allocated context between `map-event` and
   `configure-event` handlers, with only the latter's convergence path freeing
   it — but if the position already matched the target the instant the
   restore function was called (true for the Reporter, because an earlier
   `SetPosition()` call had already taken effect at the X11 level),
   `configure-event` converged and freed the context *before* the real,
   asynchronous `map-event` ever fired, which then dereferenced freed memory.
2. After that was fixed, positions still didn't hold — confirmed via
   `xwininfo` (X11 ground truth) that the compositor was still overriding the
   position *after* our old "stop watching once matched" logic had already
   declared victory and disconnected. This is the same underlying disease as
   the KWin bug, just exposed by different timing: labwc can override even
   after an early coincidental match. The final fix replaces the
   event-driven `configure-event` watcher with a straightforward time-based
   poll (fixed 2-second window after mapping, regardless of early matches),
   which was then reconfirmed fixing the original KWin case too.

### Testing performed

- Verified across multiple restart cycles that both windows now restore to a
  stable saved position (no drift) on the originally-failing KWin/Plasma 6
  laptop, and on a Raspberry Pi 4 running labwc.
- Confirmed via logging on both platforms that the poll catches and corrects a
  delayed WM/compositor override within the settle window (observed ~320ms on
  KWin, matching the previously-diagnosed timing) and gives up cleanly
  afterward without fighting the user's own repositioning.
- Reproduced and fixed a segfault (via `gdb` backtrace on a core dump) that
  only appeared once GTK3 was actually enabled on the RPi4 build.

---

## Commit history

Squashed to two clean commits (the full debugging trail, including the RPi4
detour, is preserved locally under the tag `backup-before-squash-2` if ever
needed):

1. `Fix main window position not restoring on startup under KWin`
2. `Fix FreeDV Reporter window position not restoring, and a close-time drift bug`

Branch is pushed to `barjac/freedv-gui` (`fix/window-position-restore-race`).

Note: DIAG (`log_info`) instrumentation used throughout this investigation was
stripped from the final commits. It's preserved separately as
`~/freedv-rade-build/patches/window-position-diag-overlay.patch` — not applied
by default, but available to reapply quickly if a similar issue surfaces on
another WM/compositor in the future.

Thanks to Claude Code.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -964,6 +964,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Suppress button flicker in Linux light themes. (PR #1419, PR #1421) - thanks @barjac!
     * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
     * Scalar plot label alignment fix for Y axis of Frm Decoder/Mic/Radio, SNR and Spectrum plots. (PR #1429) - thanks @barjac!
+    * Fix window position restore under KWin and labwc (main window + FreeDV Reporter window). (PR #1431) - thanks @barjac!
 2. Enhancements:
     * Add UDP broadcast of received callsigns. (PR #1367)
     * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398, #1405) - thanks @barjac!
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,7 +10,6 @@ set(FREEDV_SOURCES
     playrec.cpp
     ongui.cpp
     freedv_interface.cpp
-    gui/util/WindowPositionRestore.cpp
 )
 
 set(FREEDV_LINK_LIBS_OSX
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,6 +10,7 @@ set(FREEDV_SOURCES
     playrec.cpp
     ongui.cpp
     freedv_interface.cpp
+    gui/util/WindowPositionRestore.cpp
 )
 
 set(FREEDV_LINK_LIBS_OSX
@@ -122,7 +123,11 @@ if(LINUX)
             target_include_directories(freedv PRIVATE ${GTK3_INCLUDE_DIRS})
             target_link_libraries(freedv ${GTK3_LIBRARIES})
             add_definitions(-DHAS_GTK3)
+        else(GTK3_FOUND)
+            message(WARNING "GTK3 development files (gtk+-3.0.pc, e.g. package libgtk-3-dev) not found -- Linux-specific GUI workarounds (including window position restore) will be disabled")
         endif(GTK3_FOUND)
+    else(PkgConfig_FOUND)
+        message(WARNING "pkg-config not found -- cannot detect GTK3 development files, Linux-specific GUI workarounds (including window position restore) will be disabled")
     endif(PkgConfig_FOUND)
 endif(LINUX)
 
--- a/src/gui/dialogs/freedv_reporter.cpp
+++ b/src/gui/dialogs/freedv_reporter.cpp
@@ -30,6 +30,7 @@
 #include <wx/textdlg.h>
 #include "../controls/MsgListPopup.h"
 #include "../util/FrequencyOps.h"
+#include "../util/WindowPositionRestore.h"
 
 #if wxCHECK_VERSION(3,2,0)
 #include <wx/uilocale.h>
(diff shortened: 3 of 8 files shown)


g6wpj:
Just to say the fix works on my Kubuntu 26 (Plasma 6+Wayland) install with GTK 3.24.52

73 Matt

tmiw:
Isn't restoring window positions technically not allowed/supported under Wayland anyway? Or were the bugs happening even with `GDK_BACKEND=x11`?

barjac:
> Isn't restoring window positions technically not allowed/supported under Wayland anyway? Or were the bugs happening even with `GDK_BACKEND=x11`?

This fixes the long standing issues with x11 and xwayland using GDK_BACKEND=x11.
The main window could never be placed right at the top of the screen or the restore didn't work at all. It needed to be placed carefully about 3mm below the screen edge to restore reliably.
The reporter position always landed vertically centered.
I got used to the quirks but decided to see what Claude could figure out. It took a couple of days of diagnosis and testing but got there in the end.

PR #1434, opened 2026-07-16
Harden experimental tab layout persistence

barjac:
## Note

Once this is fully tested on all platforms, then with these changes I think it would be time to take this feature out of "Experimental Features" and bring it into mainstream.
It could have a 'disable' setting under the 'Display' tab for users who don't use it.
For now this remains under the "Enable Experimental Features" check box where it is rather lonely!

## Summary

Tab layout persistence (Tools->Options->Debugging->Enable Experimental Features)
saves/restores how the Waterfall/Spectrum/Frm Radio/Frm Mic/Frm Decoder/SNR tabs
are arranged, including splitting them into multiple groups. This PR hardens that
feature - it stays behind the same checkbox, nothing about the gating changes.

Five issues were found and fixed, all confirmed via live use on two machines
(including against an existing saved layout from before this PR):

- Saved tabs were keyed by their `AddPage()` position rather than a stable
  identity, so any future change to the tab list could silently corrupt
  previously-saved layouts (this is what required the hang fix in a7aab5d7 when
  the SNR plot was added). Tabs are now keyed by caption text instead.
- `LoadPerspective()` had an `assert()` that compiles out in release (`NDEBUG`)
  builds, so a saved layout that couldn't be parsed into any tab groups would
  null-deref instead of failing safely. It now falls back to one default group.
- `togglePTT()`/`OnTogBtnVoiceKeyerClick()` each scanned a fixed list of 5 plot
  windows to determine which tab to restore when tabs are split across groups
  (needed because `wxAuiNotebook::GetSelection()` is ambiguous once more than one
  group is visible). Replaced both copies with a shared
  `captureCurrentMicGroupTab_()` helper that reads the target group's active page
  directly, so it isn't tied to a specific set of panels.
- Save (on exit) and load (on startup) were each independently gated on the live
  `experimentalFeatures` flag. Since load only happens at startup, toggling the
  checkbox off then back on before quitting could silently overwrite a saved
  layout with the session's unloaded/flat state. Save is now gated on the flag
  value cached at startup instead.
- `LoadPerspective()` could leave a zero-page tab group in the layout when an old
  saved layout's entries don't resolve to a current tab (e.g. exactly the
  upgrade case above). `wxAuiNotebook`'s own drag-and-drop handling doesn't
  tolerate that and asserts; empty groups are now cleaned up via
  `RemoveEmptyTabFrames()`.

## Test plan

- [x] Split tabs into multiple groups; layout saves and restores correctly across restart
- [x] TX/RX toggling and voice keyer recording switch to/from the correct tab with a split "Frm Mic" group
- [x] An existing saved layout from before this PR degrades gracefully (collapses to one group) instead of hanging or crashing
- [x] Reproduced and fixed a drag-and-drop crash caused by an empty tab group left over from an old-format saved layout
- [x] Toggling Experimental Features off then back on before exiting no longer overwrites a good saved layout

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -965,6 +965,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
     * Scalar plot label alignment fix for Y axis of Frm Decoder/Mic/Radio, SNR and Spectrum plots. (PR #1429) - thanks @barjac!
     * Fix window position restore under KWin and labwc (main window + FreeDV Reporter window). (PR #1431, #1433) - thanks @barjac!
+    * Harden experimental tab layout persistence: fix positional index corruption across version upgrades, a crash on malformed saved layouts, and incorrect tab restore with split tab groups. (PR #1434) - thanks @barjac!
 2. Enhancements:
     * Add UDP broadcast of received callsigns. (PR #1367)
     * Add Time-Out Timer (TOT) capability to FreeDV. (PR #1366, #1398, #1405) - thanks @barjac!
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -965,7 +965,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
     * Scalar plot label alignment fix for Y axis of Frm Decoder/Mic/Radio, SNR and Spectrum plots. (PR #1429) - thanks @barjac!
     * Fix window position restore under KWin and labwc (main window + FreeDV Reporter window). (PR #1431, #1433) - thanks @barjac!
-    * Harden experimental tab layout persistence: fix positional index corruption across version upgrades, a crash on malformed saved layouts, incorrect tab restore with split tab groups, and a saved layout getting silently overwritten by toggling Experimental Features off then back on before exiting. (PR #1434) - thanks @barjac!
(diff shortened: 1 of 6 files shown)


tmiw:
wxWidgets 3.3 actually has hooks for us to write load/save code for the tab layout: https://docs.wxwidgets.org/3.3/classwx_aui_notebook.html#aecc2a52ac0e22ba553167a3d0e0798b5. I would actually prefer that we use this instead and remove the existing hack for loading/saving that's currently in the code (which is an ODR violation and could bite us one day). Another benefit is that we'd be able to make tab layouts an officially supported feature (and remove it from the Experimental Features checkbox).

Downside: It would mean that tab layouts can't be preserved unless your distro has 3.3 (or you're using the AppImage) but most users wouldn't be affected as they're either using Windows or macOS.

I'll see if I can mock something up in a separate PR with Claude's help.

Tyrbiter:
I'm using wxWidgets 3.2, for some reason Fedora has not updated to 3.3

I don't use the AppImage so it does affect at least one user :)

tmiw:
See https://github.com/drowe67/freedv-gui/pull/1435 for wxWidgets 3.3 implementation.

> I'm using wxWidgets 3.2, for some reason Fedora has not updated to 3.3
> 
> I don't use the AppImage so it does affect at least one user :)

There's a cmake option you can pass in (`-DBOOTSTRAP_WXWIDGETS=1`) that will cause FreeDV to build its own copy of wxWidgets, but that also significantly increases build time.

Tyrbiter:
It seems that the reason why Fedora doesn't have 3.3 is because it's a development release and won't be stable until the 3.4 branch is available.

barjac:
TBH I hardly know anyone using Windows or Mac with FreeDV over here. We are using various flavours of Linux and many are still on wxGTK-3.2.
I think this should stay 3.2 compatible for V1 and move forward into V2 with the option to change to >= 3.3 later. The current 'hack' currently works OK and if it 'aint broke ...

I asked Claude to check for Linux distros shipping 3.3 and nobody does:
"Nobody mainstream defaults to 3.3.x yet — it's staying a parallel dev/preview track for the foreseeable future."

PR #1435, opened 2026-07-16
Use wxWidgets 3.3 hooks to save/load tab layout.

tmiw:
Adds implementation of tab load/save using the official wxWidgets 3.3 hooks for systems which have this version (Windows, macOS, AppImage, some Linux distros). Previous logic for other Linux distros and "Enable Experimental Features" gating still remains.

Disclaimer: Implemented with the assistance of Claude Code.


barjac:
Seems impossible to set up a sensible graph display with two groups in this branch, they just multiply uncontrollably when dragging tabs to an existing group.

EDIT: I have re-tested and it does in fact work fine, I suspect that an incompatible config was causing the problem when I tested previously.

TreeTopHam:
No issues seen on Win11 or Linux Mint.

Tyrbiter:
Does this mean the experimental features are gone with wxWidgets <3.3.x?

barjac:
> Does this mean the experimental features are gone with wxWidgets <3.3.x?

Well if this is pushed to master then I guess so, which would mean that anyone stuck with wxGTK-3.2 does not get the layout saved and restored. (There is only one experimental feature :)
I can't find an on/off for the feature now so I guess it is always on with >= 3.3 and always off with <= 3.2.
With auto-enabling based on wxGTK version, if someone manages to corrupt the saved config they can't just disable it and re-start to get a basic clean display.
I think a check box in the 'display' tab of the settings would be more versatile.

Tyrbiter:
The current mechanism on wxWidgets 3.2.x is not going to stop working, it's a stable series.

Surely the best thing to do is to have a runtime switch that uses the current code unless wxWidgets is >=3.3

That way we don't lose anything for the time being.

barjac:
It is not restoring the correct RX tab. Save with 'Frm decoder' and on restore it is 'Frm Radio' in a two group layout as below.

image

Excuse the WIP experimental GUI.

tmiw:
Readded the wxWidgets <= 3.2 code and the "Enable Experimental Features" option.

@barjac: I'm not able to duplicate this on macOS, so let me get this branch loaded in Ubuntu and test there.

barjac:
> Readded the wxWidgets <= 3.2 code and the "Enable Experimental Features" option.
> 
> @barjac: I'm not able to duplicate this on macOS, so let me get this branch loaded in Ubuntu and test there.

I think it is the problem that was fixed in 1434 as follows:
(I may be wrong :)
Saved tabs were keyed by their AddPage() position rather than a stable
identity, so any future change to the tab list could silently corrupt
previously-saved layouts (this is what required the hang fix in https://github.com/drowe67/freedv-gui/commit/a7aab5d77f33e11d34488bec038ee359de91d0a0 when
the SNR plot was added). Tabs are now keyed by caption text instead.

barjac:
No change re. the wrong tab on restore @ #acced

Before close:
Screenshot_20260718_000845

On restart:
Screenshot_20260718_000950

tmiw:
> No change re. the wrong tab on restore @ #acced

OH, I think I understand now. I thought the issue was that it wasn't restoring after going from TX to RX, not on app restart. I'll investigate further.

tmiw:
Does the latest work any better?

TreeTopHam:
- Tab restore after 'TX-to-RX':     **OK**  (it was never broken in this latest PR)
- Tab restore 'Before Close - After Restart':    **Still broken** as @barjac has noted above

It seems that after a restart, the leftmost tab in each window is always selected as the active tab. Is that intended?

Tested on Win11 and Linux Mint.

Window Tabs

See attached photo

tmiw:
Considering that we're still going to keep the "old" implementation for now, I say we test @barjac's PR, merge that and come back to this. Especially since this issue seems to be fixed in that one.

tmiw:
This PR should now contain @barjac's changes. Let me know how it works for everyone.

tmiw:
Going to merge. We can always revert if this ends up causing problems later.

Tyrbiter:
FYI @tmiw there is work going on to get wxWidgets 3.3.x into Fedora, currently aimed at rawhide but might come into a 'release' version fairly soon.

PR #1437, opened 2026-07-19
Fix right-click context menus dismissing before they can be read (GTK)

barjac:
## Summary
- On wxGTK, right-click context menus shown from `wxEVT_CONTEXT_MENU` open
  while the same click's button-release is still pending, so GTK delivers
  that release straight into the menu as a selection instead of leaving
  it open to read
- Fixes this for the Voice Keyer and PTT buttons' right-click menus, and
  the FreeDV Reporter dialog's Send/Clear status message buttons, using
  the same `wxEVT_RIGHT_UP`-based workaround already in place elsewhere
  in topFrame.cpp for other widgets
- Confirmed present on both wxGTK 3.2 and 3.3+, so the fix is gated on
  `__WXGTK__` only, not a wx version check
- MSW/OSX untouched (already release-based/ctrl-click respectively, so
  unaffected)

## Test plan
- [x] Confirmed fixed on Mageia 10 / Plasma 6, wxGTK 3.3 build
- [x] Confirmed fixed on wxGTK 3.2 build (`build_linux_wx32`)
- [x] No regressions found in either build

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -965,7 +965,8 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
     * Scalar plot label alignment fix for Y axis of Frm Decoder/Mic/Radio, SNR and Spectrum plots. (PR #1429) - thanks @barjac!
     * Fix window position restore under KWin and labwc (main window + FreeDV Reporter window). (PR #1431, #1433) - thanks @barjac!
-    * Harden experimental tab layout persistence: fix positional index corruption across version upgrades, a crash on malformed saved layouts, incorrect tab restore with split tab groups, a saved layout getting silently overwritten by toggling Experimental Features off then back on before exiting, and a drag-and-drop assertion crash when an old saved layout leaves an empty tab group behind. (PR #1434) - thanks @barjac!
+    * Harden experimental tab layout persistence. (PR #1434) - thanks @barjac!
+    * Fix right-click context menus dismissing before they can be read (GTK) (PR #1437) - thanks @barjac!
 2. Enhancements:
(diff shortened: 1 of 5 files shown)


tmiw:
I just compared this to master on GNOME and I don't think I'm seeing a difference. Is this a Plasma only problem?

tmiw:
Hmm, I'm not seeing this on master either. Relevant screenshots in case Fedora has a different version of Plasma or something:

Screenshot_20260719_144352

Screenshot_20260719_144520

tmiw:
That all said, I also tried this PR and didn't see a difference, so it might be okay to merge anyway. Would be nice to see a video showing the issue fixed by this PR just to be sure.

barjac:
Now, without this PR the right mouse button must be pressed and held down while moving to the wanted menu item, then released over that item to select it and close the menu.
This is different to all the other context menus that allow a rapid click (single down-up action) to expand the menu and another rapid click to select an item and close the menu.

This PR makes the four affected context menus work like the others.

I found it counter intuitive and difficult to remember not to just right click on e.g. the Voice Keyer button, which would instantly select the top option and close the menu without the chance to select a specific item. 

It is a GTK thing that does not affect Windows or Mac as far as Claude can research.

The snag with a video would be knowing what the operator was doing with the mouse.

Mageia release 10 (Official) for x86_64
Running kernel-desktop-6.18.35-1.mga10-1-1.mga10
Session type=Wayland with xwayland and GDK_BACKEND=x11

tmiw:
Tried master and this PR again on GNOME. With this PR the menu doesn't come up until I let go of the mouse. Meanwhile, if I use master, I see the following behavior:

* Rapidly press/release the right mouse button: context menu comes up, stays up.
* Press, then release after a second or so: context menu comes up on press, disappears on release.

I *think* the latter behavior is what you were trying to fix, so I'm okay with the change.

PR #1440, opened 2026-07-20
Require C++20 support when building FreeDV.

tmiw:
Require C++20 support to compile FreeDV. This is acceptable since C++20 is available in compilers used in platforms/distros that we support.

Sources:
* https://en.cppreference.com/cpp/compiler_support/20
* https://packages.ubuntu.com/search?keywords=clang&searchon=names&suite=jammy&section=all
* https://packages.ubuntu.com/search?keywords=gcc&searchon=names&suite=jammy&section=all
* https://developer.apple.com/xcode/cpp/#c++20
* https://trac.macports.org/wiki/XcodeVersionInfo#macOS11

Note: freedv-backend will also be updated to require C++20.


Tyrbiter:
Will this make any difference to the current builds @tmiw , I assume it will improve various aspects of type checking and linking?

tmiw:
> Will this make any difference to the current builds @tmiw , I assume it will improve various aspects of type checking and linking?

In theory it should eventually allow for some performance improvements as we start using C++20-specific constructs.

Tyrbiter:
> > Will this make any difference to the current builds @tmiw , I assume it will improve various aspects of type checking and linking?
> 
> In theory it should eventually allow for some performance improvements as we start using C++20-specific constructs.

OK, was wondering if it might have a bearing on the SNR issue as that is so difficult to pin down, perhaps once this merges a rebuild will change the behaviour or even prevent the bug from firing.

tmiw:
I'm not sure this can be merged. Apparently wxWidgets 3.0/3.2 has a lot of issues with C++20. Will update when I investigate more.

PR #1445, opened 2026-07-25
Modernise main window group boxes (tinted cards, configurable tint, Show menu, box reordering)

barjac:
**Note: this PR targets `v3.0-dev`, not `master`** — build/test against `v3.0-dev`.

## Summary

- Group boxes (SNR, Level, Sync, Control, etc.) now use a tinted "card"
  background instead of a plain wxStaticBox border.
- Tint colour and blend strength are user-configurable in Settings ->
  Edit Settings -> Display, replacing the original hardcoded values.
- Sync box simplified to a single status line combining sync colour and
  the current mode name; removed the redundant Mode line and Center RX
  button. The button's function is already available via centre mouse
  click on the netting indicator, so it just duplicated that and took up
  space. Has a tooltip explaining what it shows.
- Native status bar replaced with a dismissible info bar for
  playback/recording messages; mode/callsign row tinted to match.
- New Show menu: hide/restore any of 8 optional group boxes (SNR, Level,
  Sync, Audio Recording, Logging, FDV Reporting, TX Attenuation, Speaker
  Level) to create a minimal custom display.
- Right-click a movable box's title for Hide / Move to other side / Move
  Up / Move Down, so boxes can be rearranged after pruning with the Show
  menu or Hide option.
- Fixed a crash in the heard-stations list combo when hovering after its
  contents are cleared.

## Test plan

- [x] Built and tested locally against v3.0-dev
- [x] Verified existing config files upgrade cleanly — all new settings

## Screen shots

This shows a fairly typical standard layout using the default background tint in a light (Breeze Light) theme in Linux.

Screenshot_20260725_162518

Below is a compact display using an alternative tint colour with two graphical displays made persistent with the 'Enable Experimental Features' option Under Settings->Edit Settings->Debugging

Screenshot_20260725_163044

Next a display with few boxes removed using the 'Show' menu option which is able to remove any boxes except Control and Radio Freq.

Screenshot_20260725_163541

Here is an ultra compact minimal layout.

Screenshot_20260725_164424

A twin graph minimal layout using a Breeze Dark system theme. It shows the info bar for playback/recording messages.

image


tmiw:
Thanks for this. Is it possible to create another PR on v3.0-dev with just the changes that can be applied to that branch? RADEV2 stabilization is taking longer than anticipated so I'm not sure when this PR can be reviewed/merged.

barjac:
> Thanks for this. Is it possible to create another PR on v3.0-dev with just the changes that can be applied to that branch? RADEV2 stabilization is taking longer than anticipated so I'm not sure when this PR can be reviewed/merged.

OK now re-based on v3.0-dev.

EDIT: Sorry I just re-read that you wanted a second PR. I still have the V2 based version branch in github if/when you need it, however re-basing back to v3.0-dev from V2 was not major.

Regarding V2, three of us have done some on-air testing and it currently does not seem to cope with selective fading as well as V1. Audio quality is good with adequate signals, but some really weird noises come out at times! It does seem some way off prime-time yet :)

tmiw:
Thanks for rebasing. `ms-rade-v2` isn't actually ready for wider testing yet, so it's good not to base this change on that for the time being.

Quick question: are the tinted boxes something that's fairly common on Linux now? I haven't noticed that in Fedora, FWIW, but I don't exactly use much of the OS outside of the terminal.

barjac:
> Quick question: are the tinted boxes something that's fairly common on Linux now? I haven't noticed that in Fedora, FWIW, but I don't exactly use much of the OS outside of the terminal.

I don't think so, I spent some time with Claude trying to find a way to give some clear separation between widgets. I tried to colour the existing boxes first but the box names were sitting half way through box edges so it was not possible to apply colour inside the boxes without half colouring the names. It looked worse, so I then looked at removing all the box outlines and just using colour which gives a look similar to modern web pages that use colour blocks to highlight sections. I think it looks rather nice and a lot less cluttered. Web pages seem to set the trend these days :) 
I had been asked by two people about removing widgets they never use, so added a 'Show' option similar to the reporter.
I then realized that the widgets needed a means of moving up/down within a group and also between left/right sides if they were not to be left orphaned.
I think it works rather well and having used it for a few weeks now it seems robust.

tmiw:
Just tried to run this branch on macOS with dark mode enabled:

image

A few things:

1. The "unk" at the bottom is invisible.
2. Perhaps the default background for the boxes should be closer to the non-box background on that platform?

For comparison, here's macOS in light mode:

image

Another issue while I'm testing--seems that the box backgrounds aren't responsive to the current dark mode status. After switching from light to dark mode:

image

barjac:
> Just tried to run this branch on macOS with dark mode enabled:
> 
> A few things:
> 
>     1. The "unk" at the bottom is invisible.

Its invisible because it's not there. :)  Do you really think it's needed?
I suspect it means nothing to most users and it was just clutter IMHO. USB/LSB display is not affected.

> 
>     2. Perhaps the default background for the boxes should be closer to the non-box background on that platform?

I don't understand the question - your image looks fine to me. What am I not seeing?
I often use the dark mode which is very close to that image and it looks good on my old monitor in a darkened room.
Have a play with the tint% in Settings->Edit Settings->Display.

> 
> Another issue while I'm testing--seems that the box backgrounds aren't responsive to the current dark mode status. After switching from light to dark mode.:

Yes I know about this and personally had accepted that a re-start of the program fixed it.

However, 6 hours later with Claude diagnosing wx internal bugs it should now be fixed.

In a nutshell wx cannot reliably detect the theme change. There are copious notes in the code explaining the issues.

There is a pre-existing bug from before this PR, affecting the tabs below the waterfall etc. which is related to the same root cause. However it's deeper rooted in wx code and not fixable with a workaround. On switching theme dark/light, the tab b/g in xwayland on wx3.2 carries over to the light theme and from then on the reversal perpetuates until a re-start.

barjac:
We will fix conflicts when 1450 is merged into master and master is merged into v3.0-dev or testing will fail.

tmiw:
Merged down from v3.0-dev, but found another issue on Ubuntu 24.04 / GNOME:

image

Unfortunately the error keeps popping up when I dismiss it, so this PR branch is basically unusable right now :(

barjac:
Try again now.

tmiw:
Haven't been able to test on the system that I reported the above on yet, but seems to start without errors on Fedora 44.

Re: box coloring--not sure if this should be configurable. IMO, it should look like this by default (light mode):

image

And with dark mode:

image

BTW "Radio Frequency" text isn't changing colors on dark->light mode transitions (or vice versa) but the rest of the main window looks okay at first glance:

image

barjac:
I would not object to making a boring grey the default (I think you have become addicted to terminals :) However I would not like to lose the optional colours, in fact the comments I have had have been positive. Why not offer the choice?
I deliberately kept the colours in the above examples to just a tint rather than anything too gaudy to avoid any strong objections.

Regarding the light/dark problem you see:
Have you tested a clean v3.0-dev for the same or similar issues?
Is this with wx32 or wx33, vanilla Wayland or with xwayland and GDK_BACKEND=x11?
What D.E. is Fedora using? What theme is in use?
wx is buggy when it comes to detecting themes. The theme reported by wx can lag one change of theme behind reality after a second change. Claude has worked around this where possible and in wx33 it is better than wx32.

tmiw:
Got a chance to try the latest on Ubuntu 24.04 (the error is gone now BTW) and noticed another issue:

image

(Note the "Radio Frequency" box being shorter than the others.)

> I would not object to making a boring grey the default (I think you have become addicted to terminals :) However I would not like to lose the optional colours, in fact the comments I have had have been positive. Why not offer the choice? I deliberately kept the colours in the above examples to just a tint rather than anything too gaudy to avoid any strong objections.

Yeah, I'd lean towards defaulting to the colors I suggested above depending on what mode the desktop is currently in (dark or light). If we were to allow user choice, should we just apply that choice regardless of the mode, or do something to the color selection in one or the other to make it work better?

> Have you tested a clean v3.0-dev for the same or similar issues? 

The foreground color for Radio Frequency" seems to have the same problem with v3.0-dev. Also noticed that its box is shorter than the others on this branch, too. I'll see if it's also in master; if so, it's likely a preexisting issue.

> Is this with wx32 or wx33, vanilla Wayland or with xwayland and GDK_BACKEND=x11? 

wx32 with native Wayland. IIRC Fedora doesn't have 3.3 yet.

> What D.E. is Fedora using? What theme is in use? 

KDE Plasma IIRC, but I don't recall the exact version or theme. I can see about getting that info if that'll help.

tmiw:
Merged the latest in v3.0-dev down, looks like the boxes aren't moving around like they should on macOS. Example:

image

barjac:
That is down to the build not clearing a cmake cache IIRC
See:
https://github.com/barjac/freedv-rade-build/commit/ef98cd34198bbd5597b53cc17c633cd9a4a5d5eb

tmiw:
CMake cache shouldn't matter if I delete the build folder (which I did this morning). I'll experiment a bit locally if I get some time.

src/topFrame.cpp
```diff
@@ -418,7 +643,16 @@ TopFrame::TopFrame(wxWindow* parent, wxWindowID id, const wxString& title, const
...
     SetLayoutDirection(wxLayout_LeftToRight);
-    
+
+#if !defined(__WXGTK__) || !defined(HAS_GTK3)
+    // See s_colourReferenceWindow's declaration above for why this exists.
+    // GTK3 builds don't need it (GetGroupBoxBaseColour() reads GTK settings
+    // directly there) -- skipped entirely on them, since an extra untracked
+    // child sitting directly on the frame outside any sizer turned out to
+    // interfere with the frame's own resize/layout propagation.
+    s_colourReferenceWindow = new wxWindow(this, wxID_ANY, wxPoint(0, 0), wxSize(1, 1));
```
tmiw, line 653: I forced `s_colourReferenceWindow` to `nullptr` as a test due to some stuff I saw in the [wxWidgets documentation](https://docs.wxwidgets.org/3.2/classwx_top_level_window.html#adfe7e3f4a32f3ed178968f64431bbfe0):

> This class overrides the base class [Layout()](https://docs.wxwidgets.org/3.2/classwx_top_level_window.html#adfe7e3f4a32f3ed178968f64431bbfe0) method to check if this window contains exactly one child – which is commonly the case, with [wxPanel](https://docs.wxwidgets.org/3.2/classwx_panel.html) being often created as the only child of [wxTopLevelWindow](https://docs.wxwidgets.org/3.2/classwx_top_level_window.html) – and, if this is the case, resizes this child window to cover the entire client area.

On macOS, the box positioning and window sizing seems to behave properly after doing so. It looks like this was created originally to fix dark/light mode transition issues but as of the latest in this PR, the coloring seems to be switching without any issues. Let's remove `s_colourReferenceWindow` unless it's absolutely needed on Linux (in which case adjust the #ifdefs accordingly).

barjac: *(Reply from Claude, on Barry's behalf)*

Agreed, and it's now removed (`7c2eb626`). You were right that it wasn't needed on Linux either -- it's already excluded on GTK3 builds (`GetGroupBoxBaseColour()` reads GTK settings directly there instead), for exactly the same class of bug you found on macOS: an extra untracked child window sitting on the frame outside any sizer interfering with layout/resize. `GetGroupBoxBaseColour()`'s non-GTK3 fallback already null-checked before using it, so removing its creation just leaves that branch always taking the `wxSystemSettings` path directly -- no other logic changes needed.

Can't verify the macOS positioning fix myself (Linux-only build here), so it'd be good if you could confirm on your end when you get a chance.

tmiw:
Box positioning looks okay on macOS now.

One thing that I'm not sure we can do anything about is the fact that some of the controls still seem to use the window background for their backgrounds. For example, the horizontal bars for SNR and Level:

image

Also, the buttons are supposed to be rounded on macOS but look more squared here, possibly due to the background coloring. Again, not sure if we can do anything about that.

barjac:
Looking closely at those buttons, they are rounded, however the rounding is revealing the almost black un-tinted background rather than the tint.
Maybe the corner rounding is being applied over the wrong background layer.
In linux the button corners are perfectly rounded and don't show any similar issue.

tmiw:
Here's what it looks like on Windows:

image

Radio Frequency is back to not being full width, unfortunately, but at least the rest of it looks okay. Maybe the "unk" text on Sync could be made a bit more readable.

That said, I'm not 100% sure I want to do much more fixing without a Feature Request considering how big of a change this is. Can you submit one just so we have it?

barjac:
Try again in Win and Mac. Already tested in Linux with no side effects.

The (Sync) 'unk' only showed briefly once at program start so was not useful and has been removed.

I will add a feature request.

tmiw:
da72a80 on macOS:

image

Looks like Level isn't quite right (maybe it still has the "clip" indicator that was removed from `master` and `v3.0-dev`?)

Anyway, I'll review again once the FR is submitted and the project team talks about this more.

barjac:
> Looks like Level isn't quite right (maybe it still has the "clip" indicator that was removed from `master` and `v3.0-dev`?)

Fixed.
A merge lost part of that change.

> Anyway, I'll review again once the FR is submitted and the project team talks about this more.

OK understood, FR should appear soon.

tmiw:
Closing as the project team has decided not to move forward with this.

PR #1444, opened 2026-07-25
Unconditionally add new station to Heard Station list if first heard.

tmiw:
Resolves #1443 by checking to see if the first entry is grayed out. If so, unconditionally add a new entry (even if it's for the same callsign).

PR #1448, opened 2026-07-30
Fix heard-stations callsign combo stuck-highlight, right-click behaviour, and a stale-index crash

barjac:
## Summary
- Fixes a crash in `wxListViewComboPopup::OnMouseMove()`: it deselected the previously hot-tracked row (`m_value`) without checking the index was still valid, so any external `DeleteAllItems()` on the underlying list left a stale index that crashed the next mouse-move over the popup.
- Fixes the heard-stations callsign combo (`m_cboLastReportedCallsigns`) being left showing a stale "focused/selected" highlight after the dropdown popup closes. The popup overlaps the combo's own value area while open, and GTK doesn't always repaint that region on dismiss. `OnCloseCallsignList`/`OnRightClickCallsignList` now defer the focus change via `CallAfter` and force a repaint once the popup's dismissal has fully finished.
- Also stops right-click-to-deselect from clearing the displayed callsign text -- it should only remove the selection, consistent with how right-click deselect already works in the FreeDV Reporter list.

## Test plan
- [ ] Open the heard-stations callsign dropdown, hover/select rows, confirm no crash even if the list changes underneath it
- [ ] Select a callsign from the dropdown, close the popup, confirm no stuck highlight remains on the combo
- [ ] Right-click to deselect a callsign, confirm the displayed text is left alone and only the selection/highlight clears


tmiw:
Going to merge https://github.com/drowe67/freedv-gui/pull/1444 first, then we can retest in case that PR introduces issues.

tmiw:
OK, merged down latest master. Can you test locally and make sure everything's still okay?

barjac:
Yes this still works fine with current master.

Issue #1447, opened 2026-07-28
Latest git clone errors on build - Linux


rskunath:
### Platform

Linux

### Platform Version

openSUSE Tumbleweed

### FreeDV Version

Git clone

### Steps to Reproduce

I have built FreeDV before from a git clone many times and used the below to do so (last few were the new C++ RADE version, the last was 2.3.1 also from a git clone):

git clone https://github.com/drowe67/freedv-gui.git
cd freedv-gui
I did need to install sioclient from the distro. It did not like the static version included with the source.

Then :
./build_linux.sh
And no issues building.

Now again the same steps and it errors out at;

[ 94%] Built target rade_tx_wav
[ 96%] Building C object src/CMakeFiles/rade_bpf_test.dir/kiss_fftr.c.o
[ 96%] Built target rade_v1_text_test
[ 96%] Built target rade_ber_test
[ 98%] Linking C executable rade_enc_v2_test
[ 98%] Built target rade_rx_wav
[ 98%] Built target rade_dec_v2_test
[ 98%] Built target rade_enc_v2_test
[100%] Linking C executable rade_bpf_test
[100%] Built target rade_bpf_test
[ 27%] No install step for 'build_rade'
[ 28%] Completed 'build_rade'
[ 28%] Built target build_rade
make: *** [Makefile:156: all] Error 2

I'm wondering what I missed? The Readme talks about:
cd ~/freedv-gui/codec2/build_linux but there is only a codec2-1.2.0 directory and no build_linux directory inside that?

Inside the above is another Readme that says:
Build Codec 2:

(But Tumbleweeed already has a distro available codec2 1.2.0-40.41 package from the mnhaucke OBS Tumbleweed repo)
   ```
   git clone https://github.com/drowe67/codec2.git
   cd codec2
   mkdir build_linux
   cd build_linux
   cmake ..
   make

So, I am confused now how to get FreeDV built on Tumbleweed and what all of the needed steps might be for including the new modes and Codec2 development? Do I need to source build Codec2 now, I can find out what the distro repo is doing about that but nothing there is newer than 2025 so probably old, and I don't know if it is a clone of your codec2 repo. The above codec2 and LPCnet instructions aren't clear to me?

Any advice to get past this and build the latest git version and incorporate the newest codec2 development (and keep that current too) is greatly appreciated.

Thanks in advance,
Rick Kunath, K9AO

### Expected Behavior

The source would complete the build and run.

### Actual Behavior

Build fails.

[...]

PR #1449, opened 2026-08-01
Ensure wxWidgets actually uses GTK3 before enabling workarounds.

tmiw:
Per @claude re: #1447:

*Diagnosis:* src/CMakeLists.txt enabled the GTK3-specific code path (HAS_GTK3, which makes WindowPositionRestore.cpp/main.cpp/topFrame.cpp #include <gtk/gtk.h> directly alongside wx headers) purely based on pkg_check_modules(gtk+-3.0) finding GTK3 dev files somewhere on the system — with no check on whether the linked wxWidgets was actually built against GTK3. wx's wx/defs.h always forward-declares a stand-in GdkWindow type, choosing between the real GTK3 layout or the legacy GTK2 one based on __WXGTK3__, which is set only in wx's own toolkit-specific setup.h. On a system where wx-config resolves to a GTK2-linked wxWidgets while GTK3 dev headers are independently installed (plausible on rolling-release distros like Tumbleweed with both gtk2-devel and gtk3-devel present), you get both declarations in the same translation unit → the exact conflict Rick hit, independent of include order.

*Fix:* before trusting the pkg-config GTK3 detection, I added a check_cxx_symbol_exists(__WXGTK3__ "wx/defs.h" ...) probe using wx's own resolved include dirs/flags. If wx doesn't report __WXGTK3__, the build now falls back to the existing warning-and-disable path (same pattern already used when GTK3 isn't found at all) instead of hard-failing — so normal Linux systems where wx and pkg-config agree are unaffected, and mismatched systems like Rick's just lose the window-position-restore workaround with a clear warning instead of a broken build.

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -964,7 +964,7 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Suppress button flicker in Linux light themes. (PR #1419, PR #1421) - thanks @barjac!
     * Only preserve previously selected tab on TX if it's in the same group as 'From Mic'. (PR #1428)
     * Scalar plot label alignment fix for Y axis of Frm Decoder/Mic/Radio, SNR and Spectrum plots. (PR #1429) - thanks @barjac!
-    * Fix window position restore under KWin and labwc (main window + FreeDV Reporter window). (PR #1431, #1433) - thanks @barjac!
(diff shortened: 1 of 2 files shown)


barjac:
No regressions with window positioning in Mageia 10 (wx3.3 and xwayland)

tmiw:
Thanks for testing! Now merged.

PR #1452, opened 2026-08-03
Fix crash and long hang on main window close with an unresponsive rig

barjac:
## Summary
- Fix a SIGSEGV in `MainFrame::topFrame_OnClose()` when a repeat close request arrives while the async RX/PTT shutdown from a first close is still in progress -- `m_reporterDialog` is already null by the second entry, and the unconditional `GetPosition()` call on it crashes, skipping window position save to config.
- Stop turning the modem off (whether via the Stop Modem button or app close) from blocking on rig disconnect against an unresponsive radio (e.g. powered off, connected via `rigctld`) -- this could hang for over a minute with no feedback to the user, since Hamlib's own client-side timeout/retry settings don't bound how long `rigctld` itself waits on the physical radio.

## Test plan
- [x] Radio on, Hamlib happy, modem stopped -> close: immediate, single click.
- [x] Radio on, Hamlib happy, modem running -> close: single click, brief delay while rig returns to previous frequency.
- [x] Radio off (via rigctld), modem running -> close, then close again while shutdown still in progress: no crash, no hang, closes promptly; window position/size correctly saved to config in all cases above.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2445,23 +2445,39 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
 
 void MainFrame::topFrame_OnClose( wxCloseEvent& event )
 {
-    // Grab and save the final position explicitly rather than relying
-    // solely on OnMove's live tracking -- Close()/Destroy() below trigger
-    // one last stray wxEVT_MOVE reporting a position with the title bar's
-    // height already stripped off, so position tracking must be stopped
-    // first or that stray event silently overwrites the correct value
-    // saved here.
-    auto pos = m_reporterDialog->GetPosition();
-    wxGetApp().appConfiguration.reporterWindowLeft = pos.x;
-    wxGetApp().appConfiguration.reporterWindowTop = pos.y;
-    m_reporterDialog->stopTrackingPosition();
+    if (terminating_)
+    {
+        // A previous close request already kicked off the async RX/PTT
+        // shutdown below, which calls Destroy() itself once done (see
+        // OnTogBtnOnOff()) -- nothing left to do here. Without this guard,
+        // a repeat close request while that shutdown is still in progress
+        // (e.g. slow Hamlib rig disconnect against an unresponsive radio)
+        // re-enters this handler with m_reporterDialog already null
+        // (cleared on the first pass below), crashing on the unconditional
+        // GetPosition() call.
+        return;
+    }
 
-    m_reporterDialog->setReporter(nullptr);
-    wxGetApp().SafeYield(nullptr, false); // make sure we handle any remaining Reporter messages before dispose
-    m_reporterDialog->Close();
-    m_reporterDialog->Destroy();
-    m_reporterDialog = nullptr;
-    
+    if (m_reporterDialog != nullptr)
+    {
+        // Grab and save the final position explicitly rather than relying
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -2963,13 +2979,31 @@ void MainFrame::performFreeDVOff_()
...
+    // rigctld itself waits on the physical radio it's talking to. Move the
+    // last reference onto a detached thread so that wait can't hold up
+    // turning the modem off (or app shutdown) -- the controller stays
+    // alive exactly as long as it needs to, just off to the side rather
+    // than blocking here.
+    std::thread([ptr = std::move(wxGetApp().rigPttController)]() mutable
+    {
+        SetThreadName("RigPttDisconnect");
+        ptr = nullptr;
+    }).detach();
```
tmiw, line 3002: Doesn't detaching potentially cause stuff to not be fully cleaned up if the rig is in fact being slow? i.e. the thread may go away in the middle of the disconnect.

src/main.cpp
```diff
@@ -2445,23 +2445,39 @@ void MainFrame::OnTimer(wxTimerEvent &evt)
...
-
-    m_reporterDialog->setReporter(nullptr);
-    wxGetApp().SafeYield(nullptr, false); // make sure we handle any remaining Reporter messages before dispose
-    m_reporterDialog->Close();
-    m_reporterDialog->Destroy();
-    m_reporterDialog = nullptr;
-    
+    if (terminating_)
+    {
+        // A previous close request already kicked off the async RX/PTT
```
tmiw, line 2450: Since we have a check for `m_reporterDialog != nullptr` below, is this check still necessary?

barjac: *(Reply from Claude, on Barry's behalf)*

It's guarding more than that one dereference. `m_RxRunning` isn't cleared until `stopRxStream()`, which runs deep inside the async shutdown *after* the rig-disconnect code — so there's a real window where a second close request arrives while `m_RxRunning` is still `true`. Without this guard, that second call would fall into the `if (m_RxRunning)` block below and re-enter `OnTogBtnOnOff()` a second time, spawning a second concurrent "TurningOff" thread racing the first `performFreeDVOff_()` (double-move on `rigPttController`, double-stopping audio streams, etc.) — a worse crash than the one this PR fixes. Expanded the comment in place to spell that out.

src/main.cpp
barjac: *(Reply from Claude, on Barry's behalf)*

Good catch, you're right — for "Stop Modem" while the app stays open this is fine since the process keeps running and the thread finishes on its own eventually, but for an app-close, `Destroy()` fires right after `performFreeDVOff_()` returns with nothing joining the detached threads, so if the rig is still unresponsive when the process actually exits, the thread gets killed mid-disconnect (worst case, a queued `ptt(false)` never reaches the radio). Pushed a fix: the detach threads now signal completion via a future, and on the terminating path we wait up to 3s (bounded, shared across both) before proceeding to `Destroy()`. A responsive rig still gets a clean disconnect; a truly unresponsive one is abandoned after the grace period same as before, so shutdown still can't hang indefinitely.

Tested: second close during the 3s window still closes cleanly with window position recovered on restart, and "Stop Modem" (not app-close) is unaffected since the bound only applies on the terminating path.

PR #1462, opened 2026-08-13
Fix waterfall/spectrum filling solid colour during half-duplex TX

barjac:
## Summary
- The half-duplex-TX "blank the display" path `memset()`'d the spectrum arrays to zero bytes, but this display's dB scale runs 0 (loudest) to `MIN_MAG_DB` (quietest) -- zero meant "loudest possible", not silence, so it painted the waterfall solid instead of blanking it (yellow, occasionally clamping to red at the top of the colour scale).
- Filling with `MIN_MAG_DB` alone wasn't sufficient either: the waterfall's colour scale is relative to each row's own peak (peak minus a fixed 20dB window), so a perfectly flat row -- regardless of its absolute level -- always reads as "loud relative to itself". This showed up live as a green-to-red fade as the auto-ranging baseline (`m_max_mag`) caught up to the new flat level.
- Fixed properly by detecting a row where every bin sits exactly at the floor (real spectra essentially never do) and rendering it as plain black directly, bypassing the relative intensity scale and leaving the auto-ranging baseline untouched, rather than running synthetic placeholder data through logic designed for real spectra.

## Test plan
- [x] Reproduced the original solid yellow/red waterfall during half-duplex TX
- [x] Confirmed identical bug/fix applicability on both `master` and `v3.0-dev`
- [x] Tested live against a real half-duplex TX/RX cycle -- waterfall now goes solid black at TX start, previous RX data scrolling off normally


tmiw:
I have some related code from `ms-rade-v2` that I'm going to create a separate PR for shortly. I don't think it'll help on Linux, though, just Windows and macOS.

EDIT: https://github.com/drowe67/freedv-gui/pull/1463

src/gui/controls/plot_waterfall.cpp
tmiw, line 88: Wouldn't inverting these also help with the issue solved by this PR? i.e.

```c++
    m_max_mag = MIN_MAG_DB;
    m_min_mag = MAX_MAG_DB;
```

barjac: Claude's analysis:
I don't think swapping these would change the behaviour here, for two reasons:

- `m_min_mag` gets unconditionally recomputed every call (`m_min_mag = max_mag - 20.0` a few lines down in `plotPixelData()`), so its constructor value never survives past the very first frame.
- `m_max_mag`'s constructor value only feeds the EMA on that same first-ever call. By the time half-duplex TX kicks in, `m_max_mag` has long since converged to real RX signal levels from many prior frames, so the value it started at is irrelevant.

I traced the actual blanking case (all bins == `MIN_MAG_DB`) with the swap applied: `max_mag` still resolves to `MIN_MAG_DB` (the scan loop only updates on strict `>`), `m_min_mag` still lands at `MIN_MAG_DB - 20`, and `m_max_mag` still decays slowly from whatever it held right before TX started — i.e. the same "colour ramp while m_max_mag catches up" behaviour that motivated the `silentRow` special-case. So the swap looks harmless but doesn't touch the code path that was actually producing the solid-colour fill.

Issue #1459, opened 2026-08-12
[Software Issue] CLIP indicates well before actual clipping - causes users to reduce mic level


peterbmarks:
### Platform

Linux

### Platform Version

Linux Mint 22.3

### FreeDV Version

2.3.1

### Steps to Reproduce

Transmit and speak into the microphone.
When transmitting the CLIP red text warning comes on well below actual clipping. 

### Expected Behavior

The CLIP indicator should tell the truth.

### Actual Behavior

It seems that the CLIIP indicator comes on >80% which seems too low.
An effect of this is that users will reduce their microphone level to avoid clipping. 
Propose either changing the warning from CLIP to HIGH (or similar) or increase the threshold to perhaps 95%.

MrJoeGonzales:
I 100% agree with Peter‘s comments. Users appear overly worried that their audio is too high and they turn the mic level right down to the point they are unintelligible. Alternative solutions are:

Just turn the Clip indicator off: FrmMic is the best indication anyway. The best audio is achieved with the average level greater than 60%. Peak levels above 100% (every few seconds) are hardly noticeable to the recipient.

Replace the Clip indicator with an audio peak limiter: Getting the mic level right is about 50% of the problems newcomers have with FreeDV. As FreeDV demands full and clear mic audio all the time. Clipping is far less of a problem than low mic audio levels. The AGC is a great  help, even though it overshoots for the first few second. Frequently, when a user turns away from the microphone and speaks, the audio is not just quiet, it is replaced by gibberish even under high SNR conditions.

tmiw:
FWIW there is a limiter as part of the AGC, though I'm not sure how effective it is. (And if you turn off AGC in Tools->Filters the limiter also turns off.)

Re: audio levels, I've been suggesting peaks of +/- 0.4 in Frm Mic. At least for me, people don't seem to have any issues understanding what I'm saying at those levels. I'm thinking we just change the label to "HIGH" (or whatever will fit in that space) and call it a day.

EDIT: Perhaps we can also indicate when levels are too low. Is 0.2 a reasonable lower bound here?

peterbmarks:
I think it's desirable to guide the user to have a reasonable level. Would it be possible to have the level meter be yellow when low, green when good, and turn red when over 90%?
The AGC works but the initial response time is too slow. I find that my first few words over shoot and then the gain comes down - but that's another topic (although related). Switching CLIP to HIGH (or HI for space) would at least not be as misleading and low risk.

[...]

PR #1464, opened 2026-08-14
Additional improvements for the Level slider

tmiw:
This PR does the following for the Level indicator on the main window:

* Changes to log scale instead of linear scale (-30 dB to 0 dB).
* For mic levels, level is now taken before equalizer.
* Equalizer is configured to run before AGC in audio pipeline.
* "From Mic" tab shows audio post-AGC.
* Mic gain is no longer settable from the main window
    * Mic gain is still settable from the Filter window if AGC is turned off.
    * Mic gain is reset to 0 dB (and control disabled) if AGC is turned on.


g4dya:
This seems also to affect the input level from the RX as well. If people are adjusting for mid-position they will be setting the input level to about -20 dBFS. Is this intentional? It seems on the low side.

tmiw:
> This seems also to affect the input level from the RX as well. If people are adjusting for mid-position they will be setting the input level to about -20 dBFS. Is this intentional? It seems on the low side.

I was thinking that it should be log scale for both TX and RX. Should it still be linear scale for RX? User manual guidance would likely need to change either way.

barjac:
I have been testing a level meter based on R128 loudness rather than peak level as used now.
See issue #1459
The build to test this was:
v3.0-dev @ c3f90e44 pr:1445 @ ce7d6133 | ms-mic-level-fixes.patch bcj-tx-lu-meter-test.patch
The patches are in https://github.com/barjac/freedv-rade-build/patches/
The ms-mic-level-fixes.patch has the conflicting docs stripped out to merge cleanly into v3.0-dev.

barjac:
I have tested this PR (a4d19ba) and still find it far less usable than the latest adaptive EMA meter version.

Using the adaptive EMA meter [1]
I tested with the voice keyer and a correctly recorded voice keyer file replays with the EMA meter sitting almost dead centre of the green.

Measuring the corresponding levels using ebumeter at the same point gives:
amber/green boundary  -27LUFS  (still well within AGC capture) 
middle of green  -23LUFS  (recommended target and where with AGC on or off looks identical in 'To Enc')
green/red boundary  -20LUFS  (still well below clipping)

[1] The current updated patch to test it further over this PR @ a4d19ba is:
https://github.com/barjac/freedv-rade-build/blob/main/patches/bcj-1464-meter-swap.patch

barjac:
Patch updated to apply cleanly @ ca688fa

PR #1472, opened 2026-08-25
An alternative TX mic level meter approach

barjac:
This is a modification to the existing peak level meter in order to smooth the action and make level adjustment easier. It should help to avoid the issues raised in #1459. 

To effectively avoid narrow audio spikes from causing rapid jumps in the meter reading an exponential moving average (EMA) has been applied to smooth the action.

Also, since the optimum meter reading is around 50% of full scale, there is a
colour guide to make this obvious immediately above the gauge. Shown in transmit only.

Please test and comment if you find this would be an acceptable alternative.

The intended method for setting the system mic gain would be to start transmitting whilst speaking continuously at your normal level and mic position. Adjust the system mic level to set the meter in the middle of the green section.
The AGC will then have only small corrections to make and that setting should remain satisfactory.

Screenshot_20260825_233703

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -87,16 +87,26 @@
 #define FROM_RADIO_MAX       0.8
 #define FROM_MIC_MAX         0.8
 
-// Running average (NOT an instantaneous peak-follower) of the level
-// meter's per-tick peaks, deliberately slow so a single loud
-// syllable/plosive gets smoothed into the average instead of individually
-// pegging the meter -- confirmed needed 2026-08-24: an instant-attack
-// design pegged on the 't' of "two" while ebumeter hadn't even reached
-// -23 LUFS and the pre-encode plot was well within normal limits. A
-// symmetric exponential average with tau=LEVEL_METER_TIME_CONSTANT_SEC;
-// the trade-off is added latency to genuine level changes, accepted
-// deliberately in exchange for not reacting to short-lived extremes.
-#define LEVEL_METER_TIME_CONSTANT_SEC 1.0
+// Running average (NOT a simple instantaneous peak-follower) of the level
+// meter's per-tick peaks. Uses an adaptive time constant rather than one
+// fixed value: slow (LEVEL_METER_TIME_CONSTANT_SEC) across the acceptable
+// LEVEL_METER_TARGET_LOW_PCT..HIGH_PCT range, so genuine level-setting is
+// smoothed against individual syllables/plosives (confirmed needed
+// 2026-08-24: a fixed instant-attack design pegged on the 't' of "two"
+// while ebumeter hadn't even reached -23 LUFS and the pre-encode plot was
+// well within normal limits) -- but fast (LEVEL_METER_FAST_TIME_CONSTANT_SEC)
+// below LEVEL_METER_RAMP_LOW_PCT or above LEVEL_METER_RAMP_HIGH_PCT, so the
+// meter snaps up quickly when speech starts from silence and doesn't hang
+// around in the red after a genuine overload, ramping linearly between the
+// two rates across LEVEL_METER_RAMP_LOW_PCT..TARGET_LOW_PCT and
+// LEVEL_METER_TARGET_HIGH_PCT..RAMP_HIGH_PCT. Which zone applies is decided
+// by the meter's own previous (already-smoothed) reading, not the raw
+// incoming peak, so the ramp itself moves gradually and doesn't flicker
+// between rates right at a boundary.
+#define LEVEL_METER_TIME_CONSTANT_SEC 1.5
+#define LEVEL_METER_FAST_TIME_CONSTANT_SEC 0.02
+#define LEVEL_METER_RAMP_LOW_PCT  15
+#define LEVEL_METER_RAMP_HIGH_PCT 85
 
 // dBFS level that maps to 100% on the TX level meter's scale, set below
 // true digital full scale (0 dBFS) so a nominal -23 LUFS test signal swings
@@ -110,7 +120,9 @@
 
 // Acceptable-range marker drawn as a thin green strip below the level
 // meter's gauge, in % of the gauge's own 0-100 scale. Centred on the
-// ~50% mid-scale target from LEVEL_METER_REFERENCE_DB above.
+// ~50% mid-scale target from LEVEL_METER_REFERENCE_DB above. Also used
+// as the inner edge of the adaptive time-constant ramp above, so the
+// meter's slowest ballistics line up exactly with the visible green zone.
 #define LEVEL_METER_TARGET_LOW_PCT  30
 #define LEVEL_METER_TARGET_HIGH_PCT 70
 
--- a/src/defines.h
+++ b/src/defines.h
@@ -96,7 +96,7 @@
 // symmetric exponential average with tau=LEVEL_METER_TIME_CONSTANT_SEC;
 // the trade-off is added latency to genuine level changes, accepted
 // deliberately in exchange for not reacting to short-lived extremes.
-#define LEVEL_METER_TIME_CONSTANT_SEC 2.0
+#define LEVEL_METER_TIME_CONSTANT_SEC 1.0
 
 // dBFS level that maps to 100% on the TX level meter's scale, set below
 // true digital full scale (0 dBFS) so a nominal -23 LUFS test signal swings
@@ -107,6 +107,12 @@
 // gauge-contamination bug fixed, -23 LUFS measured ~20% on the old 0dBFS
 // scale; -8dB brings that to ~50%.
 #define LEVEL_METER_REFERENCE_DB (-8.0)
+
+// Acceptable-range marker drawn as a thin green strip above the level
+// meter's gauge, in % of the gauge's own 0-100 scale. Centred on the
+// ~50% mid-scale target from LEVEL_METER_REFERENCE_DB above.
+#define LEVEL_METER_TARGET_LOW_PCT  40
+#define LEVEL_METER_TARGET_HIGH_PCT 60
 
 // TX Attenuation (0.1 dB increments)
 #define TX_ATTENUATION_MIN (-300) /* -30 dB */
--- a/src/defines.h
+++ b/src/defines.h
@@ -87,14 +87,16 @@
 #define FROM_RADIO_MAX       0.8
 #define FROM_MIC_MAX         0.8
 
-// PPM-style ballistics for the peak level meter: instant attack (jump
-// straight to each tick's peak, catching transients with no added latency),
-// release at a fixed dB/sec rate (a straight-line fall, not an average --
-// avoids the bounce/latency a running average causes). 15dB/sec is in the
-// range typical broadcast PPMs use (IEC 60268-10 ballpark). The original
-// peak-hold code effectively released at ~0.87dB/sec (0.99 per 100ms
-// tick), which is why a transient used to visibly linger for ~20+ seconds.
-#define LEVEL_METER_RELEASE_DB_PER_SEC 15.0
+// Running average (NOT an instantaneous peak-follower) of the level
+// meter's per-tick peaks, deliberately slow so a single loud
+// syllable/plosive gets smoothed into the average instead of individually
+// pegging the meter -- confirmed needed 2026-08-24: an instant-attack
+// design pegged on the 't' of "two" while ebumeter hadn't even reached
+// -23 LUFS and the pre-encode plot was well within normal limits. A
+// symmetric exponential average with tau=LEVEL_METER_TIME_CONSTANT_SEC;
+// the trade-off is added latency to genuine level changes, accepted
+// deliberately in exchange for not reacting to short-lived extremes.
+#define LEVEL_METER_TIME_CONSTANT_SEC 2.0
 
 // dBFS level that maps to 100% on the TX level meter's scale, set below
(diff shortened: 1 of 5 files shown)


tmiw:
Can you redo this PR for v3.0-dev now that #1464 has merged? I'm open to having a few other people try this out and getting their feedback.

tmiw:
BTW can you create a feature request once you port this PR to v3.0-dev?  That way there's evidence that others have tested these changes and prefer them over the current level meter.

barjac:
While working on the level-meter alternative, I got suspicious of the AGC's actual behaviour and built a small diagnostic branch that logs AgcStep's internals (input LUFS, target gain, current gain, output level) every 10ms block to CSV, viewed live via gnuplot. That made a problem immediately visible on real audio.
Gain kicked down and then struggles to get back:
Screenshot_20260903_110731

current_gain_db is yanked straight back down by the next loud syllable, over and over.

AGC_ATTACK_TIME_SEC = 0.5 / AGC_RELEASE_TIME_SEC = 6.0 is classic broadcast-compressor design: fast attack to catch peaks, slow release to avoid audible pumping. But that's solving a different problem. The intent here was to emulate a slow hand riding a fader. A human doing that doesn't snap the fader down and creep it back up, they move it at roughly the same gentle pace in both directions. With the asymmetric timing, every loud syllable pushes gain back down almost instantly, throwing away the slow climb built up.
  
I made attack and release symmetric, both 3.0s, also tried attack-only at 3s first, already a big improvement but settled on both at 3s as the better match. Separately lowered AGC_LOUDNESS_TARGET_LUFS from -23 to -24 based on testing. May take that further.
AGC gain now "wanders gently" around a settled level instead of stair-stepping down on every syllable.

Gentle variations now about 0dB:
Screenshot_20260903_160703

Also had a good audio report from a regular contact over a full hour QSO on 160m with this change in place.

Which brings me to the level meter.
If we keep the AGC gain so it sits around 0dB then a centre-zero optimum setting meter (AGC gain, but not called that, small range, e.g. ±5dB FSD with green centre zone) would let users visually find this sweet spot when setting mic level — same established pattern as gain-reduction meters on compressors/channel strips, not a new UI concept.

The AGC itself should only be a correction for natural variation, not the primary level-setting mechanism. If it's continuously correcting many dB, the input gain is mis-set rather than the AGC being at fault.

This also resolves a smaller, separate oddity for free: AgcStep::reset() zeros gain on every TX start, which is a mismatch to correct if the nominal gain isn't 0dB. Probably the main cause of the decaying envelope seen after TX start.
Once the setup procedure targets 0dB nominal, that reset stops being a compromise and is simply correct, so no need to persist gain across overs or sessions either, which would have been a more complex fix.

Please let me know what you think of all this before I submit a PR on v3.0-dev as it will remove all discussions and work on level meters if implemented, and solve a long standing fight between the AGC and the input level.

Edit: If the ability to turn off the AGC is retained we can leave it running and just switch the AGC output tap to it's input, keeping the meter alive.

tmiw:
I'm open to tweaking the AGC parameters, but perhaps we should review how target gain and current gain are calculated. Basically:

* Target gain: `targetGainDb_ = AGC_LOUDNESS_TARGET_LUFS - lufs;` (`lufs` is what libebur128 returns for momentary loudness). For example, if we're aiming for -23 LUFS but we're currently measuring -26, this would be +3.
* Current gain: `currentGainDb_ += ((targetGainDb_ - currentGainDb_) / agcInterval) * ((float)numSamplesPerRun_ / sampleRate_);`. At the current 6 seconds for `AGC_RELEASE_TIME_SEC`, a 48K sample rate and starting at sample 0, this would be `currentGainDb_ += (3 / 6) * (160 / 48000)` or adding 0.00166667 dB per block of 160 samples until we reach -23. *Note: the amount added to `currentGainDb_` depends on what libebur128 reads during those six seconds.*

Actually, I just reread the above and the amount added to `currentGainDb_` isn't consistent. I'm wondering if the formula actually needs to be `currentGainDb_ += (targetGainDb_ / agcInterval) * ((float)numSamplesPerRun_ / sampleRate_);` instead. I'll create a branch/PR on freedv-backend for you to try.

tmiw:
Actually, the way that it currently works, I think we need to keep the formula as-is (or at least continue tweaking it). Maybe @g4dya has some insight as well?

barjac:
> I'm open to tweaking the AGC parameters

>  I think we need to keep the formula as-is (or at least continue tweaking it.

I agree.

From the results in testing I would suggest 2sec attack and decay timings as a first change.

On air, two stations with good audio appreciation could not detect any undesirable effects even when using 1sec for both, but I noticed much more corrective action being applied by the AGC in that case and decided on 2sec as optimum. Moderate gain lag with low gain swings. 

This produces a smooth agc gain response like the originally intended 'slow hand on the fader'.

g4dya:
@tmiw I have several comments and will post them next week. I'm afraid I won't have time before then.

g4dya:
OK here goes. Sorry it's rather long.

AIUI the TX audio chain has two stages in what we call AGC: a loudness leveler followed by a limiter. The loudness leveler's time constants - 0.5 sec attack and 6.0 sec decay - suggest it's currently working as a compromise between a leveler and a compressor. There's no set of time constants that will make this work well.

We are trying to do three things with gain: prevent clipping, standardise loudness (LUFS) between different stations to (e.g.) -23 LUFS and reduce loudness range (LRA) from a single station as an operator goes off mic _etc._

To recap: loudness is measured in ITU-R Rec. BS.1770 / EBU Tech R128 by putting the audio through a K-filter which weights the frequency response to match human perception of loudness and then measures the RMS power (relative to a 0dBFS sine wave) in 100ms chunks. Any quiet chunks, more than 10dB (IIRR) quieter than its neighbours, are deemed invalid. Then the average is taken of the last four valid chunks and this is the measured loudness in LUFS. It updates every 100ms but with a latency of at least 400 ms. During silences there are no new valid chunks so the measured loudness is frozen at its previous value.

LRA is measured over a whole transmission by splitting the 100ms valid chunks into deciles. The 1st and 10th deciles are treated as outliers and ignored. LRA is the difference in LU between the 2nd and 9th deciles. We don't need to use LRA in FreeDV-gui, but we can measure it using ebumeter (or similar tools) to assess the effect of compression.

In our TX audio chain I think we need, in order:
1. The bargraph level meter - 30 dB log range, fast as possible up, 12 dB/second down. Used purely to adjust input to FreeDV-gui and prevent clipping. If this is a sample peak meter rather than true peak meter, red indication should start at -4 dBFS.
2. Noise reduction and EQ.
3. Loudness leveler - should have same slow up and down ramps, maybe 1 dB per second. Can increase or decrease gain by maybe ±12 dB. This gain adjustment should be stored at end of transmission and used at start of next transmission.
4. Soft-knee compressor/limiter. This acts as a compressor with, say, a 2:1 compression ratio up to a certain level and as a limiter (infinite compression ratio) at the highest level. If the transition between compression and limiting is gradual it will sound much nicer - this is the soft knee. Attack time <= 1ms, decay time ~ 250 ms. The compression will reduce LRA and the limiter will catch peaks due to gain increases earlier in the chain. Maximum gain for this stage is unity.

If we're still showing the processed audio on the so-called 'From mic' tabbed display, it should show the signal at the output of the compressor/limiter. This should also be the point at which loudness is measured for the leveller.

The input to FreeDV-gui is AIUI represented by 16-bit integer samples and can reach 0dBFS. Two subsequent stages - the EQ and the leveller - can increase gain so extra headroom will be needed to prevent clipping. This could be done either by using a larger integer sample size and bitshifting and padding with zeroes (gains 6dB per bit shifted), or by using 32-bit FP audio, which has a 24-bit mantissa and 8-bit exponent.

Where sample-rate conversion is done, this can also cause slight overshoots so it's worth having a little headroom in hand there as well.

I suggest we change the term AGC, which conflates different things, and use the terms 'loudness leveller' (or just 'leveler') and 'compressor/limiter' as appropriate.

It might be interesting to have four bargraph displays available as an option.
1. Input level. 30 dB log scale
2. Leveler gain. Center zero. ±12 dB
3. Compression/limiting. 20 dB log scale, unity gain on right. Would normally show full scale but momentarily duck leftwards as compression/limiting applied.
4. Output level (to RADE encoder). 30 dB log scale, same as inout level.

barjac:
Thanks Richard, very interesting, now been read three or four times :)

The major difference that I see here would be the sampling of the audio for levelling after the final soft knee limiter. This should reduce sudden leveller control excursions and allow slightly shorter leveller sampling times.

I still think that the leveller should be targeted towards unity gain to remove the need for any gain value storage and allow TX start to always reset the leveller gain to 0dB as it does in #1464.

The current 30dB log level meter in #1464 already exists.
The leveller would need the sampling input moving and the parameters adjusting.
The design of the clipper stage would need adapting.

@g4dya: I assume you mean temporary bar graphs for testing only? I think the graphing arrangement is more informative as the stored output data may be viewed live and/or reviewed later, rather than trying to evaluate 'bouncing bars'.  See the test plots above.

g4dya:
@barjac The reason for sampling the audio after the compressor/limiter is that we're targeting -23 LUFS into the RADE encoder, so that's where it should be measured. (The compressor/limiter will change the loudness.)

The leveler should never make sudden gain excursions. That should be done only by the compressor/limiter. A key point here is that slower-acting stages must come before faster-acting ones. The leveler is only the 'slow hand on the fader'. That must be its only role or it won't work as intended. The compressor/limiter does the rest.

BTW I currently have an Orban Optimod FM broadcast processor on loan in the shack. Its leveler ramps at 0.5 dB/sec. I think my suggestion of 1 dB/sec for both up and down ramps is about right for our purposes.

BTW I've just spotted an inconsistency in my comment above in stage 4. If 2:1 compression is applied, this implies above unity gain at low signal levels.

If peak input levels are set in the green on the bargraph, then the loudness will be in the ballpark of -23 LUFS anyway. I don't think there's anything more that could realistically be done to target the leveler to unity gain. Obviously we don't want the gain approaching the suggested ±12 dB limits, but I see little risk of that happening. 

I don't understand why anyone wants the TX start to reset the leveler gain to 0 dB. It's analogous to the slow hand pushing the fader back to its original position between transmissions. Why do that?

tmiw:
See https://github.com/drowe67/freedv-gui/pull/1485, where some of @g4dya's proposed changes were applied. I suggest moving discussion there.

barjac:
@g4dya 
In answer to 'why do that?'
The object of this bug and pr was initially to help simplify setting of the system mic level.
It is set correctly when the leveller has nothing to do, i.e. leveller gain is 0dB.
If it is not 0dB (or very close) then the system mic level is wrong and our attempts to make setting it easier/correctly have failed.
This was the reason for my testing of a centre zero meter tracking leveller gain for use in mic level setting.
If the level is correct then clamping the leveller initially to 0dB gain breaks nothing and avoids more messy saving/setting of inter over and session gain levels. Which must be wrong if not zero, or at least the result of some wild excursion which is not relevant to the next transmission anyway.

In addition it can be a feature to indicate correct setting by observing the initial 'To Enc' wave amplitude of the first words spoken after hitting XMIT.

If the mic level is too low, the first words will be low and vice versa.

The setting procedure should be to confirm that the first words are at the same level as what follows.

This removes the need for any other means of monitoring the mic level or leveller gain and the level meter could be removed entirely.

g4dya:
@barjac
The idea that there is a 'correct' mic level that can be set is too simple. It's possible to set a level that seems correct at the time, but humans get excited, or start to waffle or mumble, or move relative to the mic, and then the level goes up or down. So AGC of some sort is needed. The broadcasters had exactly the same problem and arrived at levelers followed by compressor/limiters decades ago. (They went much further than that but we don't need to.)

We still need to prevent clipping so the AGC has an undistorted signal to work with. Hence the need for the input bargraph. For manual level setting, the meter should provide the operator with immediate visual feedback. Any kind of integration or averaging will delay this visual feedback which would be unergonomic. As the sole purpose of this adjustment is to prevent clipping, it's the peak level that matters. That's why peak metering is universal in modern professional recording. There's nothing different in amateur radio to justify ignoring this practice.

barjac:
> @barjac The idea that there is a 'correct' mic level that can be set is too simple. It's possible to set a level that seems correct at the time, but humans get excited, or start to waffle or mumble, or move relative to the mic, and then the level goes up or down. So AGC of some sort is needed. The broadcasters had exactly the same problem and arrived at levelers followed by compressor/limiters decades ago. (They went much further than that but we don't need to.)
> 
I don't disagree with any of that at all, maybe I should have said 'normal' not 'correct'.

> We still need to prevent clipping so the AGC has an undistorted signal to work with. Hence the need for the input bargraph. For manual level setting, the meter should provide the operator with immediate visual feedback. Any kind of integration or averaging will delay this visual feedback which would be unergonomic. As the sole purpose of this adjustment is to prevent clipping, it's the peak level that matters. That's why peak metering is universal in modern professional recording. There's nothing different in amateur radio to justify ignoring this practice.

Again I don't disagree in principle but assuming a 'normal' setting is adjusted, the leveller should be capable of correcting for the variations likely in a radio shack.
Otherwise you are saying that users still have to sit watching the meter and adjusting the level manually! 
I don't sit watching the level meter during a QSO, but I do try to keep mic position and voice level to 'my' normal and keep an occasional eye on the 'To Enc' plot.
I was only saying that the level meter 'could' be removed, not that it 'should' be.

PR #1481, opened 2026-09-06
Reduce waterfall and scalar plot paint cost; fix uninitialised heatmap LUT entry

tmiw:
## Summary

Cuts the per-frame cost of the plot panels, and fixes an uninitialised read in the waterfall's colour lookup table found along the way.

The common thread: on macOS `wxGraphicsContext::DrawBitmap(wxBitmap)` is much more expensive than the `wxGraphicsBitmap` overload. `wxBitmapRefData::GetImage()` caches nothing, so every call allocates a fresh `NSImage` and draws through `-[NSImage drawInRect:]`, where the `wxGraphicsBitmap` path goes straight to `CGContextDrawImage`. The waterfall made one such call *per pixel block* — `m_imgHeight/dy`, about 200 on a 600px tall plot — on each of its 10 frames per second.

## Changes

- **Waterfall blocks are held as `{wxBitmap, wxGraphicsBitmap}` pairs.** The `wxBitmap` remains the `StretchBlit` target; the renderer-native copy is made once when the block is filled, instead of on every frame it spends scrolling down the screen.
- **`PlotScalar` composites its plot area the same way.** The conversion is per-frame here since `plotArea_` is redrawn each time, but it still beats the `NSImage` round trip. (`plotLinesBMP_` in `drawGraticuleFast()` has always taken this path — this just applies it to the data bitmap too.)
- **Waterfall graticule labels are laid out once per resize.** `wxWindowMac::DoGetTextExtent` builds and destroys a whole `wxGraphicsContext` per call, and `drawGraticule()` was making ~25 of them per frame for text that only moves on resize. It also measured every 1-second tick *before* checking whether that tick gets a label, discarding most of the work.
- **Dropped `BeginLayer(1.0)`/`EndLayer`** from the waterfall, scalar and spectrum plots — at opacity 1.0 the transparency layer allocates and composites an offscreen buffer to produce exactly what drawing directly produces.
- **Fixed `m_heatmap_lut`**, which was filled over `0..254` while `plotPixelData()` clamps intensity to 255 and reaches it whenever a bin sits at the top of the current range. The hottest pixels took their colour from indeterminate memory on essentially every frame carrying signal.

## Measurements

Timed in-process, driving `draw()` on each panel's real in-app repaint path:

| | master | this branch | |
|---|---|---|---|
| Waterfall 700×600 (full repaint) | 1.648 ms | **0.453 ms** | 3.6× |
| Frm Mic 700×200 (data-only repaint) | 0.381 ms | **0.205 ms** | 1.9× |

## Verification

Rendered both panels offscreen through `draw()` and diffed against master over **400 frames** — enough to fill the waterfall and exercise block recycling ~200 times, which is the risky part of the change since a recycled `wxBitmap` gets blitted into while older `wxGraphicsBitmap`s may still reference it:

- master vs. performance changes alone: **0 differing pixels** out of 420,000 (waterfall) and 140,000 (Frm Mic)
- performance changes vs. + LUT fix: **1,665 pixels**, all in the peak signal column, moving from `RGB(15,0,0)` / `RGB(62,0,0)` / `RGB(220,0,0)` to the intended full-scale red. That is the bug being fixed, and confirms the bad entry was genuinely being read.

`test_rade_loss.sh` passes, but it cannot really speak to these changes — it exercises no plotting, and it fails intermittently on master on the same machine (master 5 pass / 1 fail, this branch 4 pass / 3 fail across interleaved runs of the two builds).

## Notes

The `wxGraphicsBitmap` change is verified for macOS only; the label caching and transparency-layer removal should help on all platforms. `PlotWaterfall::refreshData()` and `PlotSpectrum::refreshData()` are still dead (the timers call `Refresh()` directly) — left alone deliberately, since `m_rxFreq` is `FDMDV_FCENTRE - g_RxFreqOffsetHz` and so never 0 in normal use, meaning wiring them up would change nothing.

https://claude.ai/code/session_01ADMCssMazpW1igtvMHXNWZ

### DIFF
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -955,6 +955,9 @@ LDPC | Low Density Parity Check Codes - a family of powerful FEC codes
     * Fix Hamlib-related build failure with Hamlib 5.0~git. (PR #1477)
     * macOS: Fix dyld error on startup when enabling sanitizers in build. (PR #1478)
     * Fix potential equalizer bug that could introduce corrupted audio. (PR #1480)
+    * Fix uninitialized value read during waterfall plot render. (PR #1481)
+2. Other:
+    * Waterfall and other plot performance improvements. (PR #1481)
 
 ## V2.4.0 August 2026
 
--- a/src/gui/controls/plot_scalar.cpp
+++ b/src/gui/controls/plot_scalar.cpp
@@ -289,8 +289,6 @@ void PlotScalar::draw(wxGraphicsContext* ctx, bool repaintDataOnly)
         lineMap_[index].y2 = INT_MIN;
     }
 
-    ctx->BeginLayer(1.0);
-
     int x, y;
 
     for(i = 0; i < m_samples; i++) {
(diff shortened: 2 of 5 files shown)


tmiw:
Need to confirm similar performance improvement on Linux before merge.

tmiw:
Ubuntu 24.04 performance measurement (per `perf record` / `perf report`):

* TX (`PlotScalar::draw`): 4.34% `master`, 4.12% this PR
* RX (`PlotWaterfall::draw`): 10.18% `master`, 5.72% this PR

So, slight improvement for the "Frm Mic" plot but about 50% ish improvement for the waterfall.

Note: testing performed with straight Wayland, not XWayland.

PR #1485, opened 2026-09-08
Test freedv-backend AGC changes.

tmiw:
DO NOT MERGE - TEST BRANCH ONLY

Tests freedv-backend AGC changes (see https://github.com/tmiw/freedv-backend/pull/49).


tmiw:
I would rebuild again if you already tried this earlier; there were bugs that I found in freedv-backend while updating the unit tests.

g4dya:
A quick test with e7a322b shows that the bargraph level meter is post-AGC.

tmiw:
> A quick test with [e7a322b](https://github.com/drowe67/freedv-gui/commit/e7a322b2b179f671fed0e60fe22a869be87082ba) shows that the bargraph level meter is post-AGC.

Yeah, this branch is off of master and not v3.0-dev. Should I rebase so we have the updated plot/level meter behavior?

g4dya:
@tmiw I think that would help.

barjac:
Also please don't remove the leveller 0dB gain reset on XMIT start.
At least not without reading:
https://github.com/drowe67/freedv-gui/pull/1472#issuecomment-5598182464

tmiw:
> @tmiw I think that would help.

Done.

g4dya:
@tmiw Thanks. Now it's easier to see what's going on. Would it also be possible, as a temporary diagnostic tool, to have buttons to enable the leveler and limiter stages separately? Each stage would have fixed unity gain if disabled.

barjac:
The leveller control is still from its input not looped back from the limiter, is this going to be implemented along with the limiter re-design?

First tests of this as it stands indicate that the leveller is making rapid step gain adjustments again rather than smooth ones since the change to fixed rate control. The original method with equal 2sec 'attack/decay' seemed better.

With improved test diagnostics built yesterday I am hoping to get useful graphical data to evaluate going forward using identical pre-recorded audio samples for each comparative test, to avoid any moving goal posts.

@Richard: Do you have any such recordings that would make good test samples? 30sec max. is enough. No audio playback would be included in any published test results.

barjac:
While diagnosing something else on this branch, I found that `WebRtcAgc_Process` (configured here as `compressionGaindB=0`/`limiterEnable=1`, intended as a pure limiter) still applies a hardcoded 3:1 compression ratio internally (`kCompRatio` in the vendored WebRTC AGC source, not exposed via `WebRtcAgcConfig`). Tapping the signal directly before/after that stage shows a consistent ~+2.85dB of makeup gain across most of the practical input range, independent of the leveller's own gain -- tapering only once the signal nears the stage's own internal target. This is present identically on stock `master`, so it's not specific to this PR.

As a partial workaround, shifting the leveller's own target down from the current -23 LUFS to around -26 LUFS compensates for it across the typical operating range -- not a full fix, since the WebRTC stage's contribution is itself level-dependent, but it brings the net output back close to the originally intended level for normal speech.

Flagging here since it's directly relevant to the leveller/limiter design being discussed.

---
🤖 Posted via [Claude Code](https://claude.com/claude-code) on Barry's behalf

tmiw:
In the interest of getting something out ASAP, I added two new command line arguments (`-agcleveler` and `-agclimiter`). They both default to 1 but you can pass 0 to disable one and/or the other. Example to disable just the leveler:

```
$ ./src/freedv -agcleveler 0
```

Note: I haven't tested this so I may need to push additional commits if you guys find issues.

g4dya:
@tmiw I tried d748073 but both the leveler and limiter appear to be off regardless of those command line arguments. This assumes that the Frm mic tabbed display is still post-AGC.

barjac:
OK I have some initial results of tests run on identical input audio.
The upper graph in each pair of graphs is showing:
image

The top graphs are using the original time based levelling with a 2 second 'attack/decay' it also has the workaround for the extra gain reported above from the limiter stage.

The bottom graphs are from this PR as is. The sharp steps in the leveller gain can be seen, these become larger as the input level diverges from 'normal'.

image

tmiw:
OK, try the latest in this PR again. The new command line flags were accidentally gated under the unit test stuff so they weren't even being processed. I'd also `rm -rf _deps/freedv-backend*` before building because some typos were fixed there too.

barjac:
> In the interest of getting something out ASAP, ....

Where is 'out'?
I thought this was an investigation aimed at getting this right at some point and why hasty changes were reverted before 2.4 release.

Is there a deadline?

tmiw:
> > In the interest of getting something out ASAP, ....
> 
> Where is 'out'? I thought this was an investigation aimed at getting this right at some point and why hasty changes were reverted before 2.4 release.
> 
> Is there a deadline?

No deadline. My computer time outside of my day job is pretty limited these days, so I didn't want you guys to be waiting potentially a week or two for actual GUI changes (especially if said changes aren't staying).

(Although maybe I could have just asked Claude to do them. Oh well.)

g4dya:
@tmiw Tested 5b50954. The leveler seems to be working as expected when the limiter is disabled. The limiter's behaviour seems odd and unlike a normal compressor or limiter. Difficult to describe what it's doing and difficult to know what to suggest.

tmiw:
FWIW Claude thinks that the limiter is supposed to be unity gain (mostly):

> **How the WebRTC "limiter" is wired into your pipeline**
>
> In AgcStep.cpp, WebRTC's AGC is deliberately run in kAgcModeUnchanged — i.e. you're using it purely as a peak limiter, not its analog/digital gain-adaptation modes. Your leveler (the ebur128 loudness step) computes and applies gain first; WebRtcAgc_Process is then called on the leveled samples as a second-stage brick-wall guard against clipping.
> 
> WebRtcAgc_Process internally does two things (agc.c:1932):
> - Always calls WebRtcAgc_ProcessDigital — this is where the actual limiting/compression happens.
> - Also calls WebRtcAgc_ProcessAnalog (since kAgcModeUnchanged < kAgcModeFixedDigital and mode != kAgcModeAdaptiveDigital), which updates an internal micVol and returns a suggested outMicLevel — but AgcStep.cpp discards that value (inMicLevel/outMicLevel are throwaway locals). So that whole analog branch runs but has zero effect on the output samples in your usage. Not a bug, just wasted cycles worth knowing about.
> 
> The digital limiter itself: two parts
> 
> 1. WebRtcAgc_CalculateGainTable (agc.c:376) — builds a static 32-entry gainTable once, whenever config changes. The table is indexed by zeros = number of leading zero bits in the squared peak-sample envelope for a 1ms subframe (computed via NormU32/__clz_uint32). So: loud signal → few leading zeros → low table index; quiet signal → many leading zeros → high table index.
> 
> For most indices (i >= limiterIdx), the table entry is a soft-knee compressor gain computed from a fixed-point log/exp approximation targeting compressionGaindB/targetLevelDbfs with a hardcoded 3:1 ratio (kCompRatio). With your config (compressionGaindB = 0), this curve is essentially flat/unity gain across most of the range — there's no real compression happening for typical signal levels.
> 
> For the lowest few indices (i < limiterIdx, agc.c:523–527), the gain is instead forced by a completely different, linear formula tied to limiterLvl:
> ```c
> if (limiterEnable && (i < limiterIdx)) {
>     tmp32 = (i - 1) * kLog10_2 - limiterLvl * (1 << 14);
>     y32 = DivW32W16(tmp32 + 10, 20);
> }
> ```
> This is the actual limiter: it overrides the compressor curve only at the very top of the table with a hard ceiling near targetLevelDbfs.
>
> With your specific config values, plugging through the math:
> - agcConfig_.compressionGaindB = 0, targetLevelDbfs = -LIMITER_LEVEL_DB = 1
> - analogTarget computed in WebRtcAgc_UpdateAgcThresholds works out to 4 (since compressionGaindB = 0)
> - limiterOffset is hardcoded to always be 0 in this code — the alternate branch that would compute a nonzero offset is commented out (agc.c:415-416, //zeroGainLvl += ...). This isn't something your branch changed; it's how upstream WebRTC's legacy AGC ships.
> - That gives limiterIdx ≈ 3
> 
> So only the top 3 of 32 table entries get the limiter override — i.e. the limiter only engages when the per-ms envelope is already within a hair of full-scale (0–2 leading zero bits on a squared int16 sample, which is essentially "already at or past clipping"). Everywhere else, gain just follows the near-unity compressor curve with no ceiling enforcement. That's a very narrow trigger band for something meant to be a general safety limiter.
>  
> 2. WebRtcAgc_ProcessDigital (agc.c:595) applies that table every 10ms frame: it tracks fast/slow envelope followers per 1ms subframe, looks up/interpolates the corresponding gain from gainTable via the zeros index, applies a speech-gate gain reduction, does an overload check that nudges gain down by ~0.1dB steps if applying it would still overflow, then linearly interpolates gain sample-by-sample and hard-clips to int16.
> 
> **Why it likely "doesn't seem to work" in practice**
> 
> Two things stand out from the code, independent of each other:
> 
> 1. The limiter only fires in a ~3-entry-wide band right at full scale, as derived above — for anything short of near-instantaneous full-scale samples, it's invisible, so moderate overshoot from the leveler won't be caught at all.
> 2. Upstream hard-clipping in AgcStep.cpp. The leveler applies its gain via a float multiply and converts back to int16 (ConvertSingleSampleToIntSampleType_, which clamps to [INT16_MIN, INT16_MAX]) before the samples ever reach WebRtcAgc_Process. So if the leveler's gain (up to AGC_MAX_GAIN_DB = 12dB) pushes a sample past full scale, it's already been hard-clipped (distorted) by the time WebRTC's limiter sees it — the limiter has nothing left to "limit" gracefully at that point, it can only act on a waveform that's already been chopped.
> 
> If the goal is a limiter that actually catches leveler overshoot before clipping, you'd want either the leveler's own gain scaling to soft-clip/leave headroom, or limiterIdx/limiterLvl tuned (e.g. via targetLevelDbfs/analogTarget) so the override band covers a wider range of "zeros" indices rather than just the top 3.

One thought I had is that we could remove WebRTC_AGC entirely and instead, do something like this prior to converting back to int16:

```c++
if (floatSampleVal >= 1.0) {
    floatSampleVal = std::tanh(floatSampleVal);
}
```

Although I remember trying tanh before for the Flex waveform and it didn't go over too well, so maybe there's another formula we can use instead?

tmiw:
@barjac / @g4dya, try building the latest again? Nothing changed in this repo but I did make a change in freedv-backend to use the cubic formula from https://hackaudio.com/tutorial-courses/learn-audio-programming-table-of-contents/digital-signal-processing/distortion-effects/soft-clipping/ for soft clipping.

barjac:
@tmiw Not tested yet as I am currently testing the full approach suggested by @g4dya using a soft knee compressor but with hard limiting detected and applied after a 3~5ms delay to avoid missing any samples. So far testing looks very good. With the leveller control signal fed forward from the limiter, narrow spikes are very noticeably not upsetting the leveller gain.
The cubic distortion as described in the video does look promising  and his comparison with the arctan explains the issue that would cause. I will test ASAP.

tmiw:
Any updates on whether the updated limiter helps at all?

PR #1492, opened 2026-09-14
Scale UI for Windows DPI awareness (FromDIP), with wx<3.1 fallback

tmiw:
## Summary
- The Windows resource template (`contrib/freedv.rc.in`) already declares Per-Monitor-V2 DPI awareness (`wxUSE_DPI_AWARE_MANIFEST 2`), but no code used wxWidgets' `FromDIP()` to scale hardcoded pixel sizes. Since the app tells Windows it's DPI-aware, Windows won't bitmap-stretch it as a fallback, so on a high-DPI display the UI rendered undersized/cramped.
- Wrap the ~170 hardcoded `wxSize`/`SetSize`/`SetMinSize`/`SetSizeHints` pixel values across the main window and dialogs with `FromDIP()` so layouts scale with the monitor's DPI.
- Add `src/gui/util/DipCompat.h`, a small compatibility shim so this still compiles against wxWidgets < 3.1.0 (`FromDIP` was added in 3.1.0), since Ubuntu 22.04 CI builds against the system `libwxgtk3.0` package rather than a bootstrapped newer copy. Under wx < 3.1.0 it provides no-op global `FromDIP()` overloads that unqualified calls fall through to via ordinary C++ lookup; under wx >= 3.1.0 the real member always takes priority, so behavior there is unchanged.

## Test plan
- [x] Built locally on macOS against the bootstrapped wxWidgets 3.3.3 — compiles and links cleanly.
- [x] Temporarily forced the wx<3.1 shim branch to compile alongside the real 3.3.3 headers to confirm no ambiguity/unused-function issues under `-Werror`.
- [ ] CI build/tests on this branch (Windows, macOS, Linux incl. Ubuntu 22.04 with system wx 3.0).
- [ ] Manual verification on an actual high-DPI Windows display.

https://claude.ai/code/session_01Jz1Nz4hmnEzwQ7hjQtCBhk


tmiw:
The issue is more that the text isn't resizing, not necessarily the windows themselves. Going to close this and create another PR once there's a solution.

PR #1500, opened 2026-09-18
Fix crash when testing Easy Setup with an invalid audio configuration

tmiw:
## Summary

A user reported that FreeDV crashes on Windows when clicking **Test** on the Easy Setup window while the configured audio devices are invalid/can't be opened.

Root cause: `WASAPIAudioEngine::start()` can fail internally (e.g. Windows Audio service issue, no active endpoints, a COM error) but `IAudioEngine::start()` returns `void`, so callers have no way to know it failed. When that happens, `inputDeviceList_`/`outputDeviceList_`/`devEnumerator_` are left null. `EasySetupDialog::OnTest()` calls `start()` and immediately calls `getAudioDevice()`, which calls into `getAudioDeviceList()`/`getDefaultAudioDevice()` — both of which unconditionally dereferenced those now-null COM pointers, causing a null pointer access violation instead of the intended "couldn't open device" error dialog.

## Changes

- `WASAPIAudioEngine::getAudioDeviceList()` and `getDefaultAudioDevice()`: null-check the device collection/enumerator before dereferencing, returning an empty list / invalid device instead of crashing when the engine failed to start.
- `EasySetupDialog::OnTest()`: show an error message when the analog playback test device fails to open, matching the existing behavior for the radio/TX test device (previously this failure was silent).
- `EasySetupDialog`: extracted the "stop test" logic into a `stopTest_()` helper and call it from `OnCancel`/`OnClose` as well, so closing the dialog while a test is running properly tears down the test audio devices/rig control instead of leaving the shared audio engine singleton started.

## Test plan

- [ ] Build on Windows and verify Easy Setup no longer crashes when clicking Test with an invalid/unavailable audio device configured
- [ ] Verify Test still works normally with a valid audio configuration
- [ ] Verify clicking Cancel or the window close button while a test is running stops the test cleanly
- [x] Syntax-checked `dlg_easy_setup.cpp`/`.h` changes against the project's actual compile flags (macOS); `WASAPIAudioEngine.cpp` is Windows-only and reviewed manually since it can't be compiled here

https://claude.ai/code/session_017BRyLXHuZHy1b9oLt4wJFi


tmiw:
Pushed a follow-up commit addressing a related gap: the first commit only handled `IAudioEngine::start()` failing (or `getAudioDevice()` not finding the device at all). It didn't cover the case where `getAudioDevice()` succeeds but the resulting `IAudioDevice`'s own `start()` subsequently fails (e.g. WASAPI `Initialize()`/`GetBufferSize()`/`Start()` failing due to an exclusive-access conflict or unsupported format).

None of the four backends (WASAPI, Mac, PulseAudio, PortAudio) propagate that failure back to the caller — `IAudioDevice::start()` returns `void`, and `EasySetupDialog::OnTest()` never registered an error callback or checked `isRunning()` — so the "Test" button would flip to "Stop Test" with no audio actually flowing and no error shown to the user. (This wasn't a crash on its own — each backend's `start()` is internally guarded against the specific HRESULT/API failures — but it's the same class of silent-failure bug as the one this PR fixes.)

Fix: register `IAudioDevice::setOnAudioError()` on both test devices before calling `start()`, and use it to surface the actual driver error message and flag the failure back to `OnTest()`. This is safe to check synchronously right after `start()` returns without extra locking, since `start()` blocks (via `std::future::wait()`) until any error callback from that same call has already run — verified for all four backends.

Also fixed a bug this introduced along the way: `stopTest_()` only called `audioEngine->stop()` when a test device object existed, but the new failure paths can reach it after `audioEngine->start()` was already called but before either test device object was successfully created — so `stopTest_()` now always stops the engine.

